Issue #17, opened 2017-06-12
Always on top from right click menu doesn't always get applied to the window


Nine-H:
setting one app to always on top is great (unless it's gtk2 where the menu doesn't appear)

setting two apps to always on top means that the first one is always on top, and the second one sinks when it loses focus to a non always on top window that overlaps it.

expected behaviour: on top windows should only sink if another on top window that overlaps it gets focus: there should be a "top layer" window buffer for the top flagged windows that keeps them on top instead of just one.

Philip-Scott:
Which windows are you trying to set to the top, and what method are you using to do so? I use set `Super + T` to set the window on top right from the Keyboard Settings and i can't confirm that it's not working as expected

Nine-H:
I updated and rebooted my computer (on stable) and it keeps happening, here are the steps to reproduce:

1. open chromium
2. open appcenter
3. open scratch
4. set scratch to always on top from the right click menu
5. set appcenter to always on top from right click menu
6. click chromium icon on plank or it's titlebar

it will pop up _UNDER_ scratch, and _OVER_ appcenter.

the correct behaviour for one window always on top is _OVER_ scratch, _UNDER_ appcenter
the correct behaviour if you allow multiple always on top windows is _UNDER_ scratch, _UNDER_ appcenter _OVER_ any other shit you have open.

I believe that multiple always on top windows is the correct behaviour because it allows users to "pin" multiple widget apps over for example a browser window.

Philip-Scott:
Interesting... It appears that the always on top doesn't always get applied from the right click menu. I think that's the bug here

Nine-H:
cool beans. how do I work around this bug like you did? because I want tomato and hourglass to keep me focused when I study.

Philip-Scott:
You could set a keyboard shortcut to toggle always on top.

`Settings > Keyboard > Shortcuts`

![screenshot from 2017-06-12 07 48 57](https://user-images.githubusercontent.com/10786508/27034505-da592870-4f32-11e7-9a9f-e3ff60e6a513.png)

PR #20, opened 2017-06-13
Fixed: Window menu applies previous settings to new window

Philip-Scott:
There is only one menu made that is shared across all windows. Setting `always_on_top.active` caused the `activate` (and `toggled` as well) signal to execute, causing the current window to get the results of the previous window. 

Fixes: #17

### DIFF
--- a/src/Widgets/WindowMenu.vala
+++ b/src/Widgets/WindowMenu.vala
@@ -25,13 +25,18 @@ namespace Gala
 	 */
 	public class WindowMenu : Gtk.Menu
 	{
+
+		private bool setting_current_window = false;
+
 		public Meta.Window current_window {
 			get {
 				return _current_window;
 			}
 			set {
+				setting_current_window = true;
 				_current_window = value;
 				update_window ();
+				setting_current_window = false;
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


jeremypw:
I am not an expert on Gala, but rather than create an adhoc boolean, why not connect to the toggled signal  instead and only change the current window state if it differs from the menu state? i.e.

>             
            always_on_top.toggled.connect (() => {
                if (!always_on_top.active && current_window.is_above ()) {
                    current_window.unmake_above ();
                } else if (always_on_top.active && !current_window.is_above ()) {
                    current_window.make_above ();
                }
            });

It seemed to work for me (only briefly tested)

Philip-Scott:
toggled was also getting triggered on my end when the values were changed. I may have fucked up somewhere so i'll give it another try

jeremypw:
On further testing it seems it is the extra clauses I put in the if ... else if clauses that actually fix the problem - connecting the same handler to activate works as well.:
>
        always_on_top.activate.connect (() => {
            if (!always_on_top.active && current_window.is_above ()) {
                current_window.unmake_above ();
            } else if (always_on_top.active && !current_window.is_above ()) {
                current_window.make_above ();
            }
        });

A similar handler should be able to be used for "on visible workspace".

PR #32, opened 2017-07-16
pip: Add Picture In Picture plugin

ricotz:


plugins/pip/Makefile.am
```diff
@@ -0,0 +1,62 @@
...
+
+libgala_pip_la_LIBADD = \
+	$(GALA_CORE_LIBS) \
+	$(NULL)
+
+libgala_pip_la_VALASOURCES = \
+	Main.vala \
+	MoveAction.vala \
+	PopupWindow.vala \
+	RoundedRectangleEffect.vala \
```
donadigo: The RoundedRectangleEffect is not used anywhere in the code and is not needed in the vala sources here. The file could be as well removed altogether.

PR #33, opened 2017-07-17
Make alt-tab switcher customizable

erandros:
This adds a couple of dconf keys to customize the switcher:
* **icon-size**: Size of icons in the switch bar
* **icon-switch-ease-duration**: Duration of easings for icons switch
* **icon-switch-out-opacity**: Opacity of switched out icons (as in, not selected in the switcher)
* **icon-bar-center**: Center the icons bar
* **icons-bar-ease-in-duration**: Duration of easings for icons when switching in
* **icons-bar-ease-out-duration**: Duration of easings for icons when switching out
* **window-switch-ease-duration**: Duration of  easing for windows when switching in
* **window-switch-out-opacity**: Opacity of switched out windows
* **window-switch-out-zoom**: Zoom out amount of switched out windows

Keep in mind that after changing the icon size, it needs to restart to look correctly. To make it work on the fly, the code needs to be refactored.  
I'm open to suggestions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,4 +44,5 @@ po/POTFILES
 po/Rules-quot
 po/stamp-po
 src/gala
+src/**/*.o
 build
--- a/data/org.pantheon.desktop.gala.gschema.xml.in
+++ b/data/org.pantheon.desktop.gala.gschema.xml.in
@@ -271,4 +271,51 @@
 			<summary>Only show corner masks on primary monitor</summary>
 		</key>
 	</schema>
+	
+	<schema path="/org/pantheon/desktop/gala/switcher/" id="org.pantheon.desktop.gala.switcher" gettext-domain="@GETTEXT_PACKAGE@">
+		<key type="i" name="window-switch-ease-duration">
+			<default>250</default>
+			<range min="0" max="10000"/>
+			<summary>Duration of  easing for windows when switching in</summary>
+		</key>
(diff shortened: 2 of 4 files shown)


danrabbit:
This seems incredibly overkill. In general we don't make design or engineering decisions into settings. Can you instead maybe file an issue report with the problem you're trying to solve? Maybe we can come up with a solution that doesn't involved adding hundreds of new possible configurations that need to be tested and maintained

peteruithoven:
I have say I really enjoyed playing with these settings. I've always found it slightly disorientating that all other windows disappear completely and that the "switcher" also appears on the bottom. I also find the eases to slow. 
Are you sure you would need to maintain all possible settings configurations? Aren't people who change settings in dconf aware that those changes can result in bugs and weird behaviour? It's not like it's shown in a GUI. 
This would probably be less of an issue if people would be able to switch to another switcher, but this is probably not possible with something like a gala plugin?

danrabbit:
I'm gonna close this PR. Let's instead try to identify the problems that are trying to be solved by changing these properties so we can come up with a new, better default implementation that works for everybody :)

beeblebrox3:
"better default implementation that works for everybody"
I don't think that this is possible 🤔 
@gtarsia thanks for this. I'll try on my machine

drozzy:
Can we have a better alt+tab behaviour? I don't really want to see windows going in-and-out-of darkness.

This plain switcher is really good!

PR #36, opened 2017-07-18
Start to use GResource features

ricotz:


lib/Utils.vala
```diff
@@ -332,5 +333,53 @@ namespace Gala
...
+					warning (e.message);
+					return null;
+				}
+			}
+
+			return resize_pixbuf;
+		}
+
+		/**
+		 * Creates a new reactive ClutterActor at 36px with the close pixbuf
```
donadigo: This should be ...at 36px with the **resize** pixbuf

ricotz: Isn't it?

lib/Utils.vala
```diff
@@ -332,5 +333,53 @@ namespace Gala
...
+		 *
+		 * @return The resize button actor
+		 */
+		public static Clutter.Actor create_resize_button ()
+		{
+			var texture = new Clutter.Texture ();
```
donadigo: I do not like the use of `Clutter.Texture` here. It is deprecated and we should use `Clutter.Image` instead.

ricotz: Replacing deprecated symbols is not part this. While there is already a wide usage of Clutter.Texture there is no harm in following the current style.

PR #57, opened 2017-08-13
Fix switching workspaces and maximizing a window distorts it

donadigo:
This PR fixes an issue where maximizing a window while switching to the next workspace would cancel it's maximizing animation (by reparenting it to another actor) hence making the window look distorted.

### DIFF
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -919,7 +919,7 @@ namespace Gala
 				actor.set_pivot_point (0.0f, 0.0f);
 				actor.set_translation (old_inner_rect.x - ex, old_inner_rect.y - ey, 0.0f);
 				actor.set_scale (1.0f / scale_x, 1.0f / scale_y);
-				
+
 				actor.save_easing_state ();
 				actor.set_easing_mode (Clutter.AnimationMode.EASE_IN_OUT_QUAD);
 				actor.set_easing_duration (duration);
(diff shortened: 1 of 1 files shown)


src/WindowManager.vala
```diff
@@ -873,6 +873,8 @@ namespace Gala
 
 				ui_group.add_child (old_actor);
 
+				maximizing.add (actor);
```
donadigo: Previously `maximizing` list object wasn't used at all, we now actually add the actor to the maximizing list.

src/WindowManager.vala
```diff
@@ -1228,6 +1236,8 @@ namespace Gala
 
 				ui_group.add_child (old_actor);
 
+				unmaximizing.add (actor);
```
donadigo: The same goes for the `unmaximizing` list.

src/WindowManager.vala
```diff
@@ -1290,10 +1306,10 @@ namespace Gala
 				unminimize_completed (actor);
 			if (end_animation (ref minimizing, actor))
 				minimize_completed (actor);
-			if (end_animation (ref maximizing, actor))
```
donadigo: The reason for this change is that `size_change_completed (actor);` is not needed here since this signal is always immediately emitted in the overriden `size_change` method.

src/WindowManager.vala
```diff
@@ -1367,6 +1383,7 @@ namespace Gala
 			// if we have a move action, pack that window to the static ones
 			if (moving != null) {
 				var moving_actor = (WindowActor) moving.get_compositor_private ();
+				kill_window_effects (moving_actor);
```
donadigo: Kill all window effects for the moving window that is the window that is moved to another workspace.

src/WindowManager.vala
```diff
@@ -1391,6 +1408,8 @@ namespace Gala
 					(moving != null && window == moving))
 					continue;
 
+				kill_window_effects (actor);
```
donadigo: Do the same for all other windows that are not moving (but possibly maximizing / unmaximizing).

PR #58, opened 2017-08-13
Fix switching workspaces and maximizing a window distorts it

Philip-Scott:
- Use the previously unused `maximized` and `unmaximized` lists to keep track of which windows are being animated. 
- Remove animations in `end_switch_workspace` to prevent window breakage:

Fixes: https://bugs.launchpad.net/gala/+bug/1602722


matzipan:
Can you explain how this pull request fixes the issue?

donadigo:
@matzipan:
1. Open a new window.
2. Perform a quick switch to next workspace and maximize of the window by rapidly doing Super+Arrow Up and Super+Arrow right.
3.The window will be distorted because it didn't complete it's maximizing animation.

matzipan:
@donadigo my question was more along the lines of "can you explain the bug and why this fixes it?" with stuff like this I think it's good to make sure we've covered all cases and it won't reappear again under a slightly different context.

Philip-Scott:
@matzipan We found about how to fix this bug a while back on slack. 

Currently, then you change workspaces the Gala grabs a "snapshot" of the window and uses that to create the sliding animation. If two animations occur at the same time though, one will overwrite the other causing the window breakage. 

This problem was solved for the other animations by storing the WindowActors currently in use in the hashsets [here](https://github.com/elementary/gala/pull/58/files#diff-0b302de4cca7c993881f45951a730685L87) The animations would be killed and the actors reset whenever [this was called](https://github.com/elementary/gala/pull/58/files#diff-0b302de4cca7c993881f45951a730685L1285) with an already running animation.

Problem was, that the maximizing and un-max animations were never stored to begin with, and the `kill_window_effects` function was also never called when changing workspaces.

matzipan:
I confirm that this seems to fix the issue. I'm gonna throw a few more text cases at it.

src/WindowManager.vala
```diff
@@ -1290,12 +1308,11 @@ namespace Gala
 				unminimize_completed (actor);
 			if (end_animation (ref minimizing, actor))
 				minimize_completed (actor);
-			if (end_animation (ref maximizing, actor))
-				size_change_completed (actor);
-			if (end_animation (ref unmaximizing, actor))
-				size_change_completed (actor);
```
matzipan: Why was invoking these signals bad?

Philip-Scott: Because that signal is already being called here: https://github.com/elementary/gala/blob/7e6444a8ae0ca26ab37a08fb51e79ce2471e1ced/src/WindowManager.vala#L780

It appears that at least with the version of mutter in Loki, `size_change_completed` has to be called right away after executing the method of `size_change`, if not it crashes

Rico and I both tried to run that method on the `transitions_completed` signal connections on both maximize and unmax, but it crashes on Loki (although not on Juno)

cassidyjames:
@Philip-Scott @matzipan any reason this hasn't been merged?

Philip-Scott:
@cassidyjames Looks like it's being mostly fixed in Juno, with the exception of the opening animation still breaking the window if it happens while changing desktops. I'll remove all the superseded changes to only have that one instead

danrabbit:
Does this fix #221 ?

PR #73, opened 2017-09-14
Fix minimized windows not fading out when closing window overview

### LINKED ISSUE #19: Minimized window has no animation in multitasking view
When minimized windows return to their original positions upon closing multitasking view, they abruptly disappear.

(From https://bugs.launchpad.net/elementaryos/+bug/1697532)

donadigo:
Fixes #19.

This fixes an issue with fading out minimized windows when closing the window overview as well as the multitasking view. In master when the window is minimized, it'll transition to it's original state but will not fade out meaning that it will reach it's original position and then disappear without any animation / fade out. The branch changes that so that minimized windows are faded out properly when exiting the window overview.

### DIFF
--- a/src/Widgets/WindowClone.vala
+++ b/src/Widgets/WindowClone.vala
@@ -246,8 +246,8 @@ namespace Gala
 		 */
 		bool should_fade ()
 		{
-			return overview_mode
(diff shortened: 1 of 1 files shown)


matzipan:
Just remebered this bug yesterday. I will have a look later today.

PR #78, opened 2017-09-16
pip: Compare the chosen point coordinates against the real window frame

### LINKED ISSUE #76: PIP: Grabbing an area too close to a Gtk3 window only grabs it's shadow
if you select an area too close to the gtk3 window, it will grab it's shadow instead of grabbing the window behind

donadigo:
Fixes #76.

Fixes comparasion of the choosen point on the screen with window actors: instead of comparing the point to the entire actor, the branch comapres it to the real window frame rectangle. I've added a `meta_rectangle_contains` static function since there is no replacement for that in the `Meta.Rectangle` itself.

### DIFF
--- a/plugins/pip/Main.vala
+++ b/plugins/pip/Main.vala
@@ -23,6 +23,12 @@ public class Gala.Plugins.PIP.Plugin : Gala.Plugin
 	private Gala.WindowManager? wm = null;
 	private SelectionArea? selection_area;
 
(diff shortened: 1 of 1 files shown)


Philip-Scott:
Fixes the problem with clicking on a window, but not with area selection as it was mentioned on the tagged bug report. But it looks like a good start :+1:

PR #80, opened 2017-09-19
fix for newer vala: you can't use return_if_fail in constructor

mpkh:
TODO: use contracts to check for null there


ricotz:
Could you explain your reference to "newer vala"?

Dropping the check is of course worse, then implicitly returning null which I think happens currently by using this vfunc in the end.

So doing "return_val_if_fail (schema != null, this);" would be correct.

ricotz:
See https://github.com/elementary/gala/pull/81

mpkh:
> Could you explain your reference to "newer vala"?

yes, it doesn't work for vala `0.38.1`

PR #81, opened 2017-09-19
Don't use return_*if_fail() if it leads to leaks or doesn't match sig…

ricotz:
…nature


mpkh:
@ricotz please explain the difference here:

``` diff
 -			return_if_fail (schema != null);
 -			
 +			if (schema == null)
 +				return;
```

PR #86, opened 2017-09-25
Allow the notification plugin to be HiDPI compatible

tintou:


tintou:
@ricotz fixed

Philip-Scott:
@tintou This only works in Juno and above, right? I got a friend with HiDPI and it didn't work on his machine

tintou:
@Philip-Scott Yep, only with Mutter 3.26

PR #95, opened 2017-09-30
New default folder to save screenshots

purejava:
Fix for https://github.com/elementary/screenshot-tool/issues/49

### DIFF
--- a/src/ScreenshotManager.vala
+++ b/src/ScreenshotManager.vala
@@ -143,14 +143,25 @@ namespace Gala
 			selection_area.get_selection_rectangle (out x, out y, out width, out height);
 		}
 
+		static unowned string find_target_path ()
+		{
+			unowned string? path;
+
+			path = Environment.get_variable ("PANTHEON_SCREENSHOTS_DIR");
+			if (path != null && FileUtils.test (path, FileTest.EXISTS))
+				return path;
+
+			path = Environment.get_user_special_dir (UserDirectory.PICTURES);
+			if (path != null && FileUtils.test (path, FileTest.EXISTS))
+				return path;
+
+			return Environment.get_home_dir ();
+		}
+
 		static bool save_image (Cairo.ImageSurface image, string filename, out string used_filename)
 		{
 			if (!Path.is_absolute (filename)) {
-				string path = Environment.get_user_special_dir (UserDirectory.PICTURES);
-				if (!FileUtils.test (path, FileTest.EXISTS)) {
-					path = Environment.get_home_dir ();
-				}
-
+				unowned string path = find_target_path ();
 				if (!filename.has_suffix (".png")) {
 					used_filename = Path.build_filename (path, filename.concat (".png"), null);
 				} else {


src/ScreenshotManager.vala
```diff
@@ -92,6 +93,18 @@ namespace Gala
 			throw new DBusError.FAILED ("SelectArea not implemented");
 		}
 
+		public static void create_dir_if_missing (string path)
+		{
+			if (!File.new_for_path (path).query_exists ()) {
```
donadigo: Don't create the same File instance twice. Instead do:
```
var file = File.new_for_path (path);
if (!file.query_exists ()) {
    file.make_directory ();
}

...
```

src/ScreenshotManager.vala
```diff
@@ -100,6 +113,9 @@ namespace Gala
 					path = Environment.get_home_dir ();
 				}
 
+				path += "%c".printf(GLib.Path.DIR_SEPARATOR) + SAVE_FOLDER;
```
donadigo: Use `Path.build_filename` here:
`path = Path.build_filename (path, SAVE_FOLDER);`

src/ScreenshotManager.vala
```diff
@@ -92,6 +93,18 @@ namespace Gala
 			throw new DBusError.FAILED ("SelectArea not implemented");
 		}
 
+		public static void create_dir_if_missing (string path)
```
donadigo: This does not need to be static.

purejava:
Thanks for the review and the code improvements!

src/ScreenshotManager.vala
purejava: `save_image` is static, without `create_dir_if_missing` being static to compilation fails with
>ScreenshotManager.vala:117.5-117.32: error: Access to instance member `Gala.ScreenshotManager.create_dir_if_missing' denied
>				create_dir_if_missing (path);

ricotz:
Is there some UX discussion reference? I am not really convinced this being a good idea.

Afaics the behavior differs some screenshot-tool too which doesn't create "Screenshots" in the users home-dir.

purejava:
@ricotz: There was no previous UX discussion, it‘s just fixing the bug to have it consistent with screenshot-tool.

Saying this, a subfolder of UserDirectory.PICTURES should be created.

ricotz:
This is the original issue: https://github.com/elementary/screenshot-tool/issues/41

 I would prefer elementary sets some custom ENVARS like in ~/.config/user-dirs.dirs which are localized and stable after creation. So that there are defines other (elementary-)apps can consistently rely on.

purejava:
> This is the original issue: elementary/screenshot-tool#41

This conversation belongs to elementary/screenshot-tool#49 which says, galas behaviour should be consistent to screenshot-tool's by saving screenshots to a subfolder of Pictures.

Screenshot-tool looks for a GSetting /io/elementary/screenshot-tool/folder-dir/. If it finds a path there, the path is offered as the save location for the screenshot. In case folder-dir is missing, the fall back is to offer a subfolder of Pictures as the save location.

The current behaviour of gala is to save screenshots to Pictures.

This PR implements saving to a subfolder of Pictures instead. To be consistent with screenshot-tool, looking for the GSetting could be added. 

> I would prefer elementary sets some custom ENVARS like in ~/.config/user-dirs.dirs which are localized and stable after creation. So that there are defines other (elementary-)apps can consistently rely on.

What is your point here? Your suggestion seems far out of scope of the issue mentioned above to me.

So if you want a bigger change on how all this should work from your point of view, you should start a discussion about that an slack. Just blocking the PR does not lead to any solution.

ricotz:
I am not arguing that the chosen folder should be the same for gala and screenshot-tool, neither I am arguing that a dedicated subfolder in the xdg-pictures folder being used. But saying screenshot-tool has implemented it that way is not an argument for it being the proper way.

The problem is that the subfolder is not localized (currently no po-file in screenshot-tool contains a translation) and adding it to gala will require it to be translated again. Also as you mentioned if screenshot-tool uses its folder-dir setting it will be inconsistent once more.

PS: I already mentioned it in slack before I commented here, see the #ux channel.

purejava:
OK, your point of view gets more clearly now.
I agree, that a solution like having an ENVAR for screenshot-tool and 
gala as a common base would be a good solution. Better than a solution 
for each single application that requires translations for each 
application etc.

 > The problem is that the subfolder is not localized (currently no 
po-file in screenshot-tool contains a translation) and adding it to gala 
will require it to be translated again.

I do not know why the translation for the string in screenshot-tool is 
missing. The string is marked as translatable. Referring to the 
elementary dev documentation, this is sufficient. Maybe no one made a 
translation on Weblate yet?

 From your comments in this discussion, it is still not clear to me, if 
you would reject any possible solution as you'd prefer gala not being 
changed at all. If you do not want any changes to gala for saving 
screenshots, you could say so and we would save some time for all of us.

 > PS: I already mentioned it in slack before I commented here, see the 
#ux channel.

You mentioned it on Slack together with your second comment here, 
according to the time stamps. 	

I'll poke #UX on Slack again. Let's see if someone will participate in 
the discussion on that.

purejava:
OK, your point of view gets more clearly now.
I agree, that a solution like having an ENVAR for screenshot-tool and gala as a common base would be a good solution. Better than a solution for each single application that requires translations for each application etc.

> The problem is that the subfolder is not localized (currently no po-file in screenshot-tool contains a translation) and adding it to gala will require it to be translated again.

I do not know why the translation for the string in screenshot-tool is missing. The string is marked as translatable. Referring to the elementary dev documentation, this is sufficient. Maybe no one made a translation on Weblate yet? 

From your comments in this discussion, it is still not clear to me, if you would reject any possible solution as you'd prefer gala not being changed at all. If you do not want any changes to gala for saving screenshots, you could say so and we would save some time for all of us.

> PS: I already mentioned it in slack before I commented here, see the #ux channel.

You mentioned it on Slack together with your second comment here, according to the time stamps. 	

I'll poke #UX on Slack again. Let's see if someone will participate in the discussion on that.

danrabbit:
I think the general consensus is that Screenshot Tool should eventually use Gala to take screenshots because Wayland, so that sorts out the different implementation issue.

Rico is probably right that this is the correct way to set this directory instead of relying on a gsetting

purejava:
Thank you for the discussion on Slack.
Attached you'll find the changed code that reads an envar to specify the folder location.
If the envar does not exist or is empty, the folder location defaults to the user's home dir.

purejava:
The next step is to create the envar for Screenshots and probably one for Webcam too.

This will be a separate PR for a different package. I'll do a draft and link it to this PR.

ricotz:
Just add it as first option. So keep it fallback to xdg-screenshots and xdg-home.
How about calling it "EOS_SCREENSHOTS_DIR"?

ricotz:
I think gala should not create the folder, that would be the task of the service which sets the envar.

purejava:
> Just add it as first option. So keep it fallback to xdg-pictures and xdg-home.
> How about calling it "EOS_SCREENSHOTS_DIR"?

Done.

> I think gala should not create the folder, that would be the task of the service which sets the envar.

Changed.
When I got you right on Slack, you don't want to change the program that creates `~/.config/user-dirs.dirs`. The same applies to changing the program that creates `~/.profile`, right?

So let's see what service is available that could set the envar. I've found greeter, gala, wingpanel and applications-menu.
Do you have any preference which one should be altered?

ricotz:
Push it myself to save us some back and forth.

ricotz:
I guess some tool/script which could live in and triggered by pantheon-xsession-settings.

purejava:
Ok, the script could be something like:
```bash
#!/bin/sh

# Translators: These are folders to save screen-shots and webcam videos
SCREENSHOT_FOLDER=Screenshots
WEBCAM_FOLDER=Webcam

check_for_user_dirs_file () {
  if [ ! -f ${XDG_CONFIG_HOME:-~/.config}/user-dirs.dirs ]
  then
    echo >&2 "~/.config/user-dirs.dirs does not exist"
    exit 1
  fi
}

read_user_dirs_file () {
  source ${XDG_CONFIG_HOME:-~/.config}/user-dirs.dirs

  PICTURES_FOLDER=${XDG_PICTURES_DIR:-$HOME}
  VIDEOS_FOLDER=${XDG_VIDEOS_DIR:-$HOME}  
}

create_dirs_if_missing () {
  mkdir -p $PICTURES_FOLDER/$SCREENSHOT_FOLDER
  mkdir -p $VIDEOS_FOLDER/$WEBCAM_FOLDER  
}

export_envars () {
  export EOS_SCREENSHOTS_DIR=$PICTURES_FOLDER/$SCREENSHOT_FOLDER
  export EOS_WEBCAM_DIR=$VIDEOS_FOLDER/$WEBCAM_FOLDER
}

check_for_user_dirs_file
read_user_dirs_file
create_dirs_if_missing
export_envars

# vim:set et sw=2:

```
Well, as far as I know exporting envars from a script is not supported. So there are still two problems:

1. the envars need to be exported to the shell or sourced in
2. I looked through all the files of pantheon-xsession-settings, but can't see a way to link up the script

Maybe @codygarver knows how to do it.

purejava:
Maybe in the end it's not

cassidyjames:
Nitpick, but we should never use "eos" to refer to elementary OS. Would Pantheon be more appropriate if it's a desktop environment concern?

danrabbit:
Yeah `PANTHEON` would probably be more appropriate and distro agnostic

Djaler:
Any movement in this direction?

peteruithoven:
We've chosen to go with another approach: https://github.com/elementary/gala/pull/459

PR #98, opened 2017-10-05
Let the HiDPI work out of the box!

tintou:


ricotz:
<danrabbit> ah I just upgraded and re-installed gala from master and I think you might be right that it's already fixed in libmutter 3.26.1

PR #102, opened 2017-10-07
Document libxml2-utils as dependency

peteruithoven:
Running `autogen.sh` would result in
```
configure: error: xmllint not found
```
I understood from [pull request #36](https://github.com/elementary/gala/pull/36) that it was missing `libxml2-utils` as dependency


peteruithoven:
@ricotz no problem :)

PR #126, opened 2017-11-17
Update HACKING file

Thalexander:
A lot of the links were outdated, as well the file was based around bzr and Launchpad.

### DIFF
--- a/HACKING
+++ b/HACKING
@@ -15,33 +15,34 @@ Join #elementary-dev on Freenode (irc.freenode.net).
 Go through problem reports and check unconfirmed bugs or those lacking
 information and mark any duplicates you spot.
 
-    http://bugs.launchpad.net/gala
+    https://github.com/elementary/gala/issues
 
 Help getting Gala translated in your language!
 
-    https://translations.launchpad.net/gala
+    https://l10n.elementary.io/projects/desktop/gala/
 
 Answer questions.
 
-    https://answers.launchpad.net/gala
+    https://elementaryos.stackexchange.com/questions/tagged/gala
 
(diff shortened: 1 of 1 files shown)


HACKING
```diff
@@ -15,30 +15,31 @@ Join #elementary-dev on Freenode (irc.freenode.net).
...
 release. Browse it online and look for other branches at:
 
-    http://code.launchpad.net/gala
+    https://github.com/elementary/gala
 
 ====== Build the code ======
 
 Prepare the source and compile:
-    ./autogen.sh --prefix=/usr
+    ./autogen.sh --prefix=/usr --libdir=/usr/lib/x86_64-linux-gnu/
```
tintou: --libdir=/usr/lib/x86_64-linux-gnu/ is an architecture and system dependent command, better omit it here

HACKING
```diff
@@ -262,40 +251,25 @@ project's source code.
...
-    bzr commit -m 'Merge changes from lp:gala'
-
-Save a little bandwidth, branch from an existing local copy that you keep
-around:
-
-    bzr branch lp:gala gala
-    bzr branch gala/ gala-fix-123456
-    cd gala-fix-123456
-    bzr pull lp:gala
+    git commit -m 'Merge master into fix-123456'
```
ricotz: I would encourage the use of "git fetch; git rebase origin/master" instead of merge.

Thalexander: Generally pull requests are squashed which does this anyway, and since it is likely that the pull request will take a while to be reviewed (at least if it is significant) master will end up being merged into it which will undermine the rebase.

PR #147, opened 2017-12-20
Implement flash area

donadigo:
Implements `flash_area` and adds support for the `flash` DBus parameter.

The area is flashed after reading the pixels into the image surface and before saving the screenshot to a disk so that there is no collision in the process of taking a screenshot.


Djaler:
@ricotz can you check this please?

PR #151, opened 2018-01-04
Add a blur API

donadigo:
This branch requires #85 to be merged.

Ports some blur utilities and classes from the deepin-mutter fork (https://cr.deepin.io/#/admin/projects/deepin-mutter, from branch `release/3.20`).

The C `BlurActor` class is slightly altered to account for the fact that it can be added beneath the window actors. I've also added a couple of helper methods (like `meta_blur_actor_get_supported` and constants to keep the code maintainable).

In the current state, the blur effect can have visual glitches, which is because Clutter actually caches the actors and it should be patched to enable the C code to add the blur actor to the internal always-redraw list of Clutter. The following patch: https://gist.github.com/sonald/2b7fd83c2e357fe92b815e34150e858f

This branch adds 2 new methods to the DBus API, for details, look at the diff for their documentation and how they work internally.

Known issues:
* ~~The blur actor doesn't set it's viewport to account for stage zoom (which is used by the zoom plugin)~~
* With high texture downscale, when quickly resizing the window, the texture is not staying "in place"
* Occasional visual glitches

To be tested:
* Lower end GPU's / CPU's
* The above mentioned Clutter patch
* Many blur effects at once for multiple windows (performance)


cassidyjames:
I want to pretty strongly push back on this unless or until we have a consensus on the design direction that this pushes us in. Currently we have only briefly mentioned and looked at a blur behind WingPanel for foreground protection, but I don't think there was a design consensus there nor much design-first thinking.

Further, I'd prefer to have @danrabbit, @harveycabaguio, and/or @snwh play with blur concepts before we just add it as a built-in public feature in our shell. It may be that we find very specific times to use it and want to define that in the HIG and our own apps, or it may be that we don't want to introduce blur at all.

But "add public features now, design later" is not an approach I'm comfortable with.

src/blur/blur-utils.c
```diff
@@ -0,0 +1,52 @@
+#include "blur-utils.h"
```
donadigo: This file didn't have any license included with it - I left it unmodified. We probably want to put the Deepin's license here.

src/blur/meta-blur-actor.c
```diff
@@ -0,0 +1,933 @@
...
+#include <clutter/clutter.h>
+
+#include "cogl-utils.h"
+#include "blur-utils.h"
+#include <meta/errors.h>
+#include <meta/screen.h>
+#include <meta/compositor-mutter.h>
+#include <meta-blur-actor.h>
+
+static void _stage_add_always_redraw_actor(ClutterStage *stage, ClutterActor *actor)
```
donadigo: The references to normally not available API, functionality in Clutter. Those will be called only if those methods are present in the clutter package, see patch in the PR description.

src/blur/meta-blur-actor.c
```diff
@@ -0,0 +1,933 @@
...
+void (*meta_copy_sub_tex)( GLenum      target, 
+            GLint level, 
+            GLint xoffset, 
+            GLint yoffset,
+            GLint x,
+            GLint y,
+            GLsizei width,
+            GLsizei height);
+
+gboolean meta_blur_actor_get_supported (void)
```
donadigo: This method checks out if the current Cogl context can perform certain actions that are needed to perform the blur effect - if this returns `false` the DBus `enable_blur_behind` method will throw an error.

src/blur/meta-blur-actor.c
```diff
@@ -0,0 +1,933 @@
...
+    g_clear_pointer (&priv->fbTex, cogl_object_unref);
+    g_clear_pointer (&priv->fb, cogl_object_unref);
+    g_clear_pointer (&priv->fb2, cogl_object_unref);
+}
+
+static void preblur_texture(MetaBlurActor* self)
+{
+    MetaBlurActorPrivate *priv = self->priv;
+
+    float resolution[2] = {priv->fb_width, priv->fb_height};
```
donadigo: We could experiment with the resoltion a bit here. Making it gradually lower, we could get even bigger blur radius while improving peformance.

src/blur/meta-blur-actor.c
```diff
@@ -0,0 +1,933 @@
...
+            return FALSE;
+        }
+
+        cogl_texture_get_gl_texture (priv->texture, &priv->gl_handle, &priv->gl_target);
+
+        cogl_pipeline_set_layer_texture (priv->pipeline, 0, priv->texture);
+        cogl_pipeline_set_layer_filters (priv->pipeline, 0,
+                COGL_PIPELINE_FILTER_LINEAR_MIPMAP_LINEAR,
+                COGL_PIPELINE_FILTER_LINEAR);
+    } else {
```
donadigo: If the texture is already created, clear it out. It looks like the `data` parameter is not copied in the `set_data` call so we don't need to free it, the texture or rather the bitmap takes care of that.

src/blur/meta-blur-actor.c
```diff
@@ -0,0 +1,933 @@
...
+        if (error) {
+            meta_warning ("clearing blur texture data failed: %s\n", error->message);
+        }
+    }
+
+    clutter_stage_ensure_current (clutter_actor_get_stage (self));
+    x = fminf(x, fw);
+    y = fminf(fh - y - height, fh);
+
+    cogl_flush ();
```
donadigo: We need to flush the Cogl drawing context so that we don't end up with garbage data being copied to the texture.

src/blur/meta-blur-actor.c
```diff
@@ -0,0 +1,933 @@
...
+    meta_bind_texture (priv->gl_target, priv->gl_handle);
```
donadigo: Bind the texture.

src/blur/meta-blur-actor.c
```diff
@@ -0,0 +1,933 @@
...
+    meta_copy_sub_tex (priv->gl_target, 0, 0, 0, x, y, width, height);
+    meta_bind_texture (priv->gl_target, 0);
```
donadigo: Then unbind it.

src/blur/meta-blur-actor.c
```diff
@@ -0,0 +1,933 @@
...
+static void
+meta_blur_actor_allocate (ClutterActor        *actor,
+                       const ClutterActorBox  *box,
+                       ClutterAllocationFlags  flags)
+{
+    MetaBlurActor *self = META_BLUR_ACTOR (actor);
+    MetaBlurActorPrivate *priv = self->priv;
+    ClutterActorClass *parent_class;
+
+    if (priv->window) {
```
donadigo: If we have any window attatched to the blur actor, get it's frame rectangle and allocate with this rectangle.

src/blur/meta-blur-actor.c
```diff
@@ -0,0 +1,933 @@
...
+{
+    MetaBlurActorPrivate *priv;
+
+    priv = self->priv = G_TYPE_INSTANCE_GET_PRIVATE (self,
+            META_TYPE_BLUR_ACTOR,
+            MetaBlurActorPrivate);
+
+    priv->radius = 0; // means no blur
+    priv->rounds = 1;
+    priv->enabled = TRUE;
```
donadigo: Todo: create a GSetting key for this.

src/blur/meta-blur-actor.c
```diff
@@ -0,0 +1,933 @@
...
+    }
+}
+
+
+void meta_blur_actor_set_rounds (MetaBlurActor *self, int rounds)
+{
+    MetaBlurActorPrivate *priv = self->priv;
+
+    g_return_if_fail (META_IS_BLUR_ACTOR (self));
+    g_return_if_fail (rounds >= 1 && rounds <= 100);
```
donadigo: Todo: replace this `100` with a constant.

src/blur/meta-blur-actor.c
```diff
@@ -0,0 +1,933 @@
...
+
+    g_return_if_fail (META_IS_BLUR_ACTOR (self));
+
+    priv->window_actor = META_WINDOW_ACTOR (g_object_ref (window_actor));
+    if (priv->window_actor)
+    {
+        priv->window = meta_window_actor_get_meta_window (priv->window_actor);
+    } else {
+        priv->window = NULL;
+    }
```
donadigo: Todo: queue a redraw here.

src/blur/meta-blur-actor.c
```diff
@@ -0,0 +1,933 @@
...
+        clutter_actor_queue_redraw (CLUTTER_ACTOR (self));
+    }
+}
+
+void meta_blur_actor_set_window_actor (MetaBlurActor *self, MetaWindowActor *window_actor)
+{
+    MetaBlurActorPrivate *priv = self->priv;
+
+    g_return_if_fail (META_IS_BLUR_ACTOR (self));
+
```
donadigo: Todo: check if the `window_actor` isn't the same as `priv->window_actor`.

superteshi:
Hello donadigo,

this sounds very promising and we are looking forward to see you work integrated into pantheon.
Unfortunately there seem to be slowdown caused by internal decision process. We hope this gets resolved soon!

Thank you!

src/blur/meta-blur-actor.c
```diff
@@ -0,0 +1,942 @@
...
+            return FALSE;
+        }
+
+        cogl_texture_get_gl_texture (priv->texture, &priv->gl_handle, &priv->gl_target);
+
+        cogl_pipeline_set_layer_texture (priv->pipeline, 0, priv->texture);
+        cogl_pipeline_set_layer_filters (priv->pipeline, 0,
+                COGL_PIPELINE_FILTER_LINEAR_MIPMAP_LINEAR,
+                COGL_PIPELINE_FILTER_LINEAR);
+    } else if (!_stage_add_always_redraw_actor) {
```
Jantrushar: I had to remove the condition to make it work correctly with pached mutter-clutter
`} else {`

donadigo:
Closing in favour of a newer, much more performant branch based on KDE's implementation: https://github.com/elementary/gala/tree/optimized-blur-vala

PR #154, opened 2018-01-07
Scale the dock for HiDPI during Alt+Tab switching

### LINKED ISSUE #31: Switchers plank doesn't adapt size to HiDPI?
The plank that's used when switching doesn't seem to adjust it's scale for HiDPI screens, while the default plank does. 

peek 2017-07-14 18-15

Also see:  https://bugs.launchpad.net/plank/+bug/1657616

davidmhewitt:
Fixes #31


davidmhewitt:
@danrabbit That should be resolved now. Still a couple of things I want to do before it's ready to merge though:

- [ ] Update documentation comments
- [ ] Refresh scale factor when opening window switcher (in case display scale factor changes)

danrabbit:
lookin' good!

PR #157, opened 2018-01-14
Show tile previews

### LINKED ISSUE #8: Give visual feedback when snapping Windows to the left/right
Right now, when trying to snap a window to the left/right, there is absolutely no visual feedback available. It would be nice to have snap animation like in other DEs that show how the windows will snap.

Similarly, trying to quarter tile Windows using Shift key + mouse drag, there is no animation feedback.

donadigo:
Fixes #8.

Implements `show_tile_preview` and `hide_tile_preview` for showing tile previews when window is about to be tiled into a certain area.

### DIFF
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -78,6 +78,8 @@ namespace Gala
 		Clutter.Actor? last_hotcorner;
 		ScreenSaver? screensaver;
 
+		Clutter.Actor? tile_preview;
+
 		Window? moving; //place for the window that is being moved over
 
 		LoginDRemote? logind_proxy = null;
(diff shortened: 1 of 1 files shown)


donadigo:
We may want to paint the area with some CSS instead of a single color, but that's up to @elementary/ux.

cassidyjames:
Yeah I think picking this up from the system stylesheet would be ideal, if that's something we can do.

cassidyjames:
@donadigo this looks like a good start! Something Compiz and GNOME Shell do that's nice is animate the rectangle from the window to the tile area; this helps better communicate that the shape being shown will be what the window transforms into. If it's not too crazy to implement, I'd love to see that.

I also feel like the animation is a little slow. I see you're using the snap duration, but I think a feed-forward animation should be faster than the actual transform that happens. Would it be crazy to like… just halve the snap animation duration for the preview?

Lastly, like I mentioned before, I think it'd be best if we could style this with CSS (setting the color, opacity, borders, etc.) in the system stylesheet so that design tweaks wouldn't require merging and recompiling. Let me know if that's possible or practical. I'd also be okay merging this and immediately opening an issue to add that. :)

PR #162, opened 2018-01-23
Expose keep below windows

### LINKED ISSUE #158: Provide affordance for keep-below windows
When a window with `keep-below` set is mapped, it's currently difficult to know anything even happened. For example, launching an "applet" app like Nimbus or Monilet from AppCenter maps the window onto the desktop, but the AppCenter client is above it, so you never see it spawn.

Instead, we could do something clever like briefly fade all open windows to (mostly?) transparent to show that a new keep-below app is spawned.

This _might_ be a little visually heavy, so we'll need to keep it brief and simple (probably no 3D transform going on, just simple opacity). We should also consider if this will tie into a future "widget layer" that lives onto the desktop or not, and if so, we should probably use the same sort of animation for it.

Another thing to keep in mind: the applets are keep-below, which means spatially they're below the windows. So they shouldn't come in on top of the other open windows or anything.

donadigo:
Fixes #158.

Creates a little animation when a bottom-stack window appears on the stage: fades out all the windows except the keep-below window, and animates it's scale to highlight it more.

Feel free to experiment with animation keyframe values.


danrabbit:
UX approve. Does the thing. I like it :+1:

cassidyjames:
Yep yep UX approve as well. I'd love to get this into Juno for all those applets in AppCenter, as the current experience there is really subpar.

PR #167, opened 2018-01-27
Move the fullscreened window to the next workspace

### LINKED ISSUE #116: Request: Open fullscreen workspaces next to the one where you activated fullscreen and not as the furthest right.
Hi.  

I'm a new Elementary OS user.

I'm really enjoying the experience, but i noticed one thing that i think it doesn't work that well.

Imagine that i have:  
Workspace 1: browser and music player
Workspace 2: Video player

If i put the browser in fullscreen mode, it goes to workspace 3, which in my opinion make things a little bit difficult to find.  

I think that fullscreen workspaces should be created next to the one where you started so that you remember it where it was. Like this:

Workspace 1: Music Player
Workspace 2: Browser (fullscreen)
Workspace 3: Video Player.

What do you guys think?

donadigo:
Fixes #116.

When a window is fullscreened, shift all the workspaces and their windows by 1 to make space for a new workspace next to the workspace that the window was originally fullscreened on.

That means if you fullscreen a window on the first workspace, it'll go to new a second one and all other workspaces will be shifted to the right to make space for that second workspace.

### DIFF
--- a/src/InternalUtils.vala
+++ b/src/InternalUtils.vala
@@ -142,24 +142,24 @@ namespace Gala
 		 */
 		public static void insert_workspace_with_window (int index, Window new_window)
 		{
-			unowned List<WindowActor> actors = Compositor.get_window_actors (new_window.get_screen ());
-
-			var workspace_manager = WorkspaceManager.get_default ();
+			unowned WorkspaceManager workspace_manager = WorkspaceManager.get_default ();
 			workspace_manager.freeze_remove ();
 
 			new_window.change_workspace_by_index (index, false);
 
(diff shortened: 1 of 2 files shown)


cassidyjames:
From a quick UX look, this looks good and I approve. Could use the feedback of someone else from @elementary/ux for sure. And then of course an actual code review.

danrabbit:
Yeah I think this is a more expected behavior :+1:

Astro-City:
Happy :+1:

ricotz:
@cassidyjames @donadigo I have boiled it down, so please take a another look.

Issue #174, opened 2018-02-14
Header context menu on not-native apps


Djaler:
In native apps we can open header context menu with some useful options like "Always on top". But with another apps we can't. Looks like this is only for GTK3 apps. But I remember that earlier, in previous version of elementary, we can do it with any app.

artemanufrij:
@Djaler It's also possible on third party apps...

![bildschirmfoto von 2018-02-14 21 32 58](https://user-images.githubusercontent.com/16517095/36226633-ab5459c8-11ce-11e8-94c8-03f28a0c0814.png)

peteruithoven:
@artemanufrij That app (GraphUI) is also made with GTK-3: 
https://github.com/artemanufrij/graphui/blob/master/CMakeLists.txt#L20

Examples would be Chromium and Electron based apps like Atom and Gitkraken.

apcamargo:
I can confirm that this happens to me in Firefox. This should be related with the fact that the title bar doesn't appear in screenshots.

![screenshot from 2018-02-26 10 20 44](https://user-images.githubusercontent.com/22940964/36672685-ff8c6ecc-1ade-11e8-96c9-a1fc37ed9845.png)

ChristopherRogers1991:
Just wanted to note that this also effects the menu button that can be added via dconf:

![image](https://user-images.githubusercontent.com/8608191/41824136-89b4e1cc-77d9-11e8-97a8-01bdff6b609c.png)

Repro:

1) With dconf, set org > pantheon > desktop > gala > appearance : button layout to: `close, menu:maximize`.
2) Open an non-gtk application (I've tested with Firefox and VLC)
3) Click the menu button

Expected behavior:

The window menu is displayed

Actual Behavior:

The menu is not displayed; pressing the button has no effect.

danrabbit:
Closing because this was fixed in #179

cassidyjames:
Reopening because apparently this wasn't ever fixed in #179, and #338 was opened as a dupe.

jlnr:
I've created a new virtual Juno installation with all the latest daily stuff, installed VirtualBox in the live session (it's a common non-GTK app), and right-clicked the VirtualBox title bar. No menu, but a new journal entry with every click:

![screenshot von 2018-09-20 19 53 51](https://user-images.githubusercontent.com/70772/45837230-25bb6f80-bd0f-11e8-9840-892587e28d52.png)

Not sure how to debug the daemon.

cassidyjames:
[...]

PR #179, opened 2018-02-17
Use separate daemon to display GTK window menus

### LINKED ISSUE #175: Window menus are not scaled correctly for HiDPI [$50]
It looks like the font size is doubled, but the actual scale of the surface is half sized

---
There is a bounty on this issue [on Bountysource](https://www.bountysource.com/issues/55060509-window-menus-are-not-scaled-correctly-for-hidpi)

<bountysource-plugin>

---
There is a **[$50 open bounty](https://www.bountysource.com/issues/55060509-window-menus-are-not-scaled-correctly-for-hidpi?utm_campaign=plugin&utm_content=tracker%2F62872462&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F62872462&utm_medium=issues&utm_source=github).
</bountysource-plugin>

davidmhewitt:
Fixes #175 

The new gala daemon will need to be running to test, you can either run it locally from the build folder with `daemon/io.elementary.gala.daemon` or add it to the `RequiredComponents` of the pantheon session in `/usr/share/gnome-session/sessions/pantheon.session`.

I.e.

`RequiredComponents=gala;io.elementary.gala.daemon;org.gnome.SettingsDaemon.A11yKeyboard; .....`

Screenshot of new menu on HiDPI:
image


Djaler:
Hey, so what about merging this?

cassidyjames:
Blocked by #85 as @davidmhewitt mentioned

Djaler:
So what about now?

daemon/MenuDaemon.vala
```diff
@@ -0,0 +1,189 @@
...
+				warning ("Failed to get Gala proxy: %s", e.message);
+			}
+		}
+
+		void lost_gala ()
+		{
+			gala_proxy = null;
+		}
+
+		void has_gala ()
```
donadigo: I would prefer this method to be named something like `gala_appeared`, judging by the name, I would think it returns if there is a gala connection currently established instead of actually connecting to it.

daemon/main.vala
```diff
@@ -0,0 +1,150 @@
...
+			}
+
+			return session_client;
+		}
+
+		private async bool register ()
+		{
+			sclient = yield register_with_session ("io.elementary.gala.daemon");
+
+			sclient.QueryEndSession.connect (()=> {
```
donadigo: Could use a one line lambda here:
sclient.QueryEndSession.connect (() => end_session (false));

Also, space after `(()`.

daemon/main.vala
```diff
@@ -0,0 +1,150 @@
...
+				end_session (false);
+			});
+
+			sclient.EndSession.connect (()=> {
```
donadigo: Same here.

daemon/main.vala
```diff
@@ -0,0 +1,150 @@
...
+				end_session (false);
+			});
+
+			sclient.Stop.connect (()=> {
```
donadigo: Same here.

daemon/main.vala
```diff
@@ -0,0 +1,150 @@
...
+			});
+
+			return true;
+		}
+
+		private void end_session (bool quit)
+		{
+			if (quit) {
+				Gtk.main_quit ();
+				return;
```
donadigo: Is there a reason why we return here? Does GNOME SessionManager automatically ends the session for this process when it quits?

src/WindowManager.vala
```diff
@@ -119,7 +129,30 @@ namespace Gala
...
+				}
+			}
+
+			Bus.watch_name (BusType.SESSION, MENU_DBUS_NAME, BusNameWatcherFlags.NONE, has_menu, lost_menu);
+		}
+
+		void on_menu_get(GLib.Object? o, GLib.AsyncResult? res)
+		{
+			try {
+				menu_proxy = Bus.get_proxy.end(res);
```
donadigo: Space after `(`.

src/WindowManager.vala
```diff
@@ -119,7 +129,30 @@ namespace Gala
...
 				} catch (Error e) {
 					warning ("Failed to get LoginD proxy: %s", e.message);
-				}				
+				}
+			}
+
+			Bus.watch_name (BusType.SESSION, MENU_DBUS_NAME, BusNameWatcherFlags.NONE, has_menu, lost_menu);
+		}
+
+		void on_menu_get(GLib.Object? o, GLib.AsyncResult? res)
```
donadigo: Space after `(`.

src/WindowManager.vala
```diff
@@ -119,7 +129,30 @@ namespace Gala
...
+		{
+			try {
+				menu_proxy = Bus.get_proxy.end(res);
+			} catch (Error e) {
+				warning("Failed to get Menu proxy: %s", e.message);
```
donadigo: Space after `(`.

src/WindowManager.vala
```diff
@@ -119,7 +129,30 @@ namespace Gala
...
+			}
+		}
+
+		void lost_menu()
```
donadigo: Same here.

src/WindowManager.vala
```diff
@@ -119,7 +129,30 @@ namespace Gala
...
+		{
+			menu_proxy = null;
+		}
+
+		void has_menu()
```
donadigo: Same here. I would like to have this be like `menu_appeared` or something.

donadigo:
Pulled and tested and there are copule of issues, at least for me:

* I right-click on any window titlebar, and as soon as I release the mouse button, the menu disappears which makes the menu appear for a split second. 

* Non GTK windows still don't have the menu, it doesn't appear at all.

daemon/main.vala
davidmhewitt: Yes, session manager sees the disconnection in DBus. Besides, even if we don't return here and try to respond to the session manager, it won't work since there isn't a Gtk main loop to communicate with DBus anymore.

davidmhewitt:
I wasn't seeing the menu closing itself until I tried with a mouse instead of a touchpad. The only difference seems to be that the touchpad sends the button down and button up events at the same time, whereas a physical mouse sends them when the button goes up and down.

I guess mutter is firing that event on the button down and the button up event is defocusing the menu window and closing it. I'll see what I can do.

As for it not working on non-GTK windows, I don't see as that being a problem with this branch since it didn't work before? Can you give me an example of a non-GTK window/application to test with because things like Slack are working here...

cassidyjames:
Does this also solve #174?

davidmhewitt:
@cassidyjames Based on @donadigo 's comments above, I'm not sure it does solve that. Nobody has given me an example of an app/window it doesn't work with though, so haven't tested.

@donadigo @ricotz I have now removed that dirty timeout code by moving the menu 1 pixel down so that the mouse pointer isn't in it when it opens as suggested by Rico. Should be ready to review again.

ricotz:
I would like to have the daemon dbus-activatable too.

davidmhewitt:
@ricotz Having it dbus-activatable slightly conflicts with the fact this is managed by the session, unless I've missed something.

I don't think a dbus-activated daemon would be able to register with the session. But I'm not super familiar with the gnome session management stuff, so I'm happy to be corrected.

I can appreciate that making it dbus-activatable is useful in the sense that it would be automatically restarted when it was needed should it have crashed or been killed, but surely we can make the session restart it automatically.

Unless you mean you want it to be only dbus-activatable and it is only spawned when requested (i.e. when someone right clicks a title bar). In which case, that's probably a rework of the code that I'm not sure I have time for at the moment. Happy for someone else to take over and do that though.

PR #183, opened 2018-03-02
Add support for mutter 3.27.92

ricotz:


peteruithoven:
Out of curiosity, are there changelogs available for Mutter? I'm mostly curious about quarter tiling, which was listed under [Gnome 3.26's feature plans](https://wiki.gnome.org/ReleasePlanning/FeaturePlans#Plans_for_GNOME_3.26), but I can't figure out how that relates to a version of Mutter. I've also opened: https://gitlab.gnome.org/GNOME/mutter/issues/52

peteruithoven:
I was pointed to the following changelog: https://gitlab.gnome.org/GNOME/mutter/blob/master/NEWS

PR #194, opened 2018-03-10
Better notification animations

Shnatsel:
I've found the following issues in Gala notification animations (circa Loki) very annoying:

1. Notification opening animations do not attract attention. This made me miss notifications I wanted to see.
2. Notification _closing_ animations, on the other hand, _do_ attract attention. So I got distracted every time one was closing, or worse, only noticed there was a notification when it was closing and there was nothing to see anymore.
3. When there are several notifications present and another one appears, the stack move animation attracts attention a lot more than the notification that has just appeared.

This branch is intended to fix all of that. Specifically, these are the changes made:

1. Notifications with priority set to anything other than "low" attract attention when they appear, without the animation being over-the-top or overly distracting.
1. Low-priority notifications use a deliberately short, non-attention-grabbing animation. We're talking about things like volume change feedback or music track change. They're used as a form of feedback or to complement other cues.
1. When a new notification appears while several are visible already, the stack slides down with a smoother animation, without attracting unnecessary attention to the stuff you've already seen.
1. All notifications use closing animation that is designed not to attract attention.
1. As a bonus, non-low-priority animations slide up as they disappear instead of to the right. Moving to the right made sense in the original mockup where missed notifications appeared to the right of the current workspace in some views; however, that never happened. The storage for missed notification is in an indicator right above the notifications, so that's the direction in which they go. Low-priority notifications are not stored in notification history and do not slide at all.

With this branch notifications are at last in a state where nothing about them has been bothering me in the course of my day-to-day use. I've been by dogfooding for a few days after every change, and e.g. point 2 only came about because of it.


Shnatsel:
I've tried to record a screencast to showcase the new animations, but it came out really janky and not representative of the actual product. I'm afraid you can only really experience this in person.

plugins/notify/Notification.vala
```diff
@@ -160,20 +160,26 @@ namespace Gala.Plugins.Notify
...
 			opacity_transition.set_from_value (0);
 			opacity_transition.set_to_value (255);
 
-			var flip_transition = new KeyframeTransition ("rotation-angle-x");
-			flip_transition.set_from_value (90.0);
-			flip_transition.set_to_value (0.0);
-			flip_transition.set_key_frames ({ 0.6 });
-			flip_transition.set_values ({ -10.0 });
-
+			var slide_in_transition = new PropertyTransition ("anchor-x");
```
donadigo: Create this object only when the `} else {` branch is getting executed, that way, you don't create this unnecessarily.

danrabbit:
Hm I feel like there's some inconsistency here. I'm kind of busy with a bunch of other things right now, but just making a note that I'd like to put some more though here and not merge too fast

donadigo:
@danrabbit you could follow up here, maybe we can pick this up again.

danrabbit:
I'm inclined to just close this since the plan is to break out notifications into its own process. We can re-evaluate animations once that's done

PR #200, opened 2018-03-14
meson: Use install_rpath as needed

ricotz:


jtojnar:
Do I understand it correctly that this was supposed to mean `dependency('mutter-cogl-path-2').get_pkgconfig_variable('libdir')`. And that none of this will be necessary with meson 0.47 (https://github.com/mesonbuild/meson/commit/e3757e3d3cf24327c89dd3fc40f6cc933510f676).

PR #203, opened 2018-03-14
Drop autotools buildsystem

ricotz:


tintou:
looks good to me
I don't know how POTFILES.in works but the generation of the Gettext template still works so I'm guessing that it's correct

ricotz:
@tintou Thanks I additionally dropped POTFILES.skip and renamed POTFILES.in.

PR #207, opened 2018-03-18
lib: Add missing datarootdir in pkg-config file

ricotz:


tintou:
Out of curiosity, why should anyone use this?

ricotz:
It is already used inside of gala.pc and therefore a bit broken.
The variable itself might be something ancient, but widely used.

Issue #210, opened 2018-03-23
Screenshots folder


Djaler:
We need a way to set screenshots folder. At least using gsettings.

cassidyjames:
Wait, no. I thought the solution here was to use an xdg folder? We don't want to set it using GSettings as that's not l10n friendly at all, and the OS already has a method for this with [xdg-user-dirs](https://www.freedesktop.org/wiki/Software/xdg-user-dirs/).

Djaler:
What if I don't want to save screenshots in pictures folder?

Djaler:
With my PR by default screenshots will save to pictures folder. But this behaviour can be overriden by gsettings option.

cassidyjames:
I don't think that's a realistic concern though; screenshots should be saved to the screenshots folder as determined by the OS. What you do with them afterwards is up to you, but I don't see how changing a hidden setting is an improvement.

Djaler:
With screenshot tool we can set the destionation folder, for example. Alsk, as I mentioned in Slack, we should merge screenshot-tool with gala screenshotter

PR #213, opened 2018-03-24
Add gsettings option to specify screenshots folder

Djaler:
Fix for #210


cassidyjames:
Please see https://github.com/elementary/gala/pull/95. I don't think we want to introduce a GSetting, but should us an xdg-user-dir for this. This way it would be l10n friendly and still technically be settable for users who want to change it under the hood.

Djaler:
I agree with xdg-user-dir for screenshots. But, in my opinion, it should not be hardcoded to be subfolder of Pictures.

peteruithoven:
Reading through #95 this isn't the preferred approach so I'm closing this PR.

PR #214, opened 2018-03-24
Fix selection area appearing in screenshots

### LINKED ISSUE #211: Screenshot area is shown with a grey layer on top when issued via custom shortcut
(Moved from [this issue](https://github.com/elementary/screenshot-tool/issues/37))

5 out of 10 times when using `screenshot-tool -r` via the `Shift+PrtScrn` shortcut to select only an area of the screen, the screenshot is stored with a transparent grey layer on top. Example:

These are my specs:

AMD A8-3870 APU with Radeon(tm) HD Graphics
NVIDIA Corporation G96 [GeForce 9500 GT] (rev a1)

donadigo:
Fixes #211.

Waits for the stage to repaint before returning from SelectArea. This makes the combination of methods like SelectArea and ScreenshotArea work correctly, as previously the selection area actor was removed from the stage but the caller (org.gnome.SettingsDaemon.MediaKeys) called ScreenshotArea immediately so the stage didn't have any time to redraw itself.


Djaler:
This doesn't work for me

src/ScreenshotManager.vala
```diff
@@ -97,8 +100,8 @@ namespace Gala
 			selection_area.start_selection ();
 
 			yield;
+			selection_area.destroy ();
```
Djaler: `yield wait_stage_repaint ();` after this line make it work for me

src/ScreenshotManager.vala
```diff
@@ -97,8 +97,10 @@ namespace Gala
...
+			
+			yield wait_stage_repaint ();
 			selection_area.get_selection_rectangle (out x, out y, out width, out height);
```
jeremypw: You appear to be referencing selection area after destroying it?

Djaler:
As I said before in Slack, the issue fixed for me only with `yield wait_stage_repaint ();` in `screenshot_area` too.

Djaler:
This is what I change in this branch to make it work:
```
diff --git a/src/ScreenshotManager.vala b/src/ScreenshotManager.vala
index 7a9162d..beaa431 100644
--- a/src/ScreenshotManager.vala
+++ b/src/ScreenshotManager.vala
@@ -54,9 +54,11 @@ namespace Gala
                        success = save_image (image, filename, out filename_used);
                }
 
-               public void screenshot_area (int x, int y, int width, int height, bool flash, string filename, out bool success, out string filename_used) throws DBusError
+               public async void screenshot_area (int x, int y, int width, int height, bool flash, string filename, out bool success, out string filename_used) throws DBusError
                {
                        debug ("Taking area screenshot");
+
+                       yield wait_stage_repaint ();
                        
                        var image = take_screenshot (x, y, width, height, false);
                        success = save_image (image, filename, out filename_used);
```

donadigo:
@jeremypw `select_area` is not a thing that's used in Gala, it's an implementation of the DBus interface exposed by org.gnome.Shell.Screenshot, `select_area` as well as other methods are to be used by e.g the screenshot tool etc.

src/ScreenshotManager.vala
donadigo: `destroy` doesn't unref the object itself, only frees all the resources allocated by Clutter for rendering and removes it from the stage, the object still exists. `selection_area` itself will be freed after the function ends.

jeremypw: Ah, right.  That would explain why it doesn't crash!

Djaler:
@donadigo so can you add my changes?

jeremypw:
@donadigo :  Thanks for the explanation.

djaler's amendment works for me too.

PR #224, opened 2018-04-13
Add blur API v2

donadigo:
This is a follow up to https://github.com/elementary/gala/pull/151 to add a blur behind API for other applications.

This branch is largely based off of the work done by Alex Nemeth for KDE here: https://phabricator.kde.org/D9848

Here's a code that provides a demo for this branch: https://gist.github.com/donadigo/6e93d4730b66f1a6f35100d43317fe18

The code itself is actually much more different than it was for the previous method and should be much more performant. We use gradual downsampling and upsampling of the texture and on every iteration we apply the corresponding shader. The algorithm itself was described by Marius Bjørge here: https://community.arm.com/cfs-file/__key/communityserver-blogs-components-weblogfiles/00-00-00-20-66/siggraph2015_2D00_mmg_2D00_marius_2D00_notes.pdf

Unlike the previous branch, it doesn't support multiple blur amounts for individual windows - this is because of the fact that all texture pipeline is initialized statically, not on every paint (which was really expensive). 

This is ready to review.

### DIFF
--- a/src/BlurActor.vala
+++ b/src/BlurActor.vala
@@ -78,7 +78,7 @@ namespace Gala
         uniform float half_height;
         uniform float offset;
         uniform float saturation;
-		uniform float brightness;
+        uniform float brightness;
 
         vec3 saturate (vec3 rgb, float adjustment) {
             const vec3 W = vec3(0.2125, 0.7154, 0.0721);
--- a/src/DBus.vala
+++ b/src/DBus.vala
@@ -19,9 +19,9 @@ namespace Gala {
     [DBus (name="org.pantheon.gala")]
     public class DBus {
         static DBus? instance;
-		static WindowManager wm;
-		
-		Gee.HashMap<uint32, BlurActor> blur_actors;
+        static WindowManager wm;
+
+        Gee.HashMap<uint32, BlurActor> blur_actors;
 
         [DBus (visible = false)]
         public static void init (WindowManager _wm) {
@@ -87,8 +87,8 @@ namespace Gala {
         }
 
         private DBus () {
-			blur_actors = new Gee.HashMap<uint32, BlurActor> ();
-			
+            blur_actors = new Gee.HashMap<uint32, BlurActor> ();
+            
             if (wm.background_group != null)
                 (wm.background_group as BackgroundContainer).changed.connect (() => background_changed ());
             else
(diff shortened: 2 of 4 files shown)


donadigo:
@cassidyjames okay so here are my thoughts so other people can follow this PR along:

1. I understand, it's something that we can pick on whenever we want to e.g next cycle. We will just need to evaluate if this can be useful for other apps / shell, and if yes, how (modal dialogs, panels, terminal etc.)

2. Yes, I experimented with the effect a little bit here and I came into conclusion that the contents behind should be a little bit more saturated and brightened - which I did locally (didn't push this yet), and I will say it actually works better with more opaque windows and in general.

3. I'm open for suggestions here. We should probably turn off the effect completely for systems that don't have any GPU at all, while keeping it enabled for all integrated (e.g Intel) and dedicated GPU's. Not sure about if the user should control this setting or not (accesibility settings?).

4. That's also something that I thought about. Yes, we would need strong do / do not when giving app developers this API so they don't over use it (making mac OS look alikes and what not), however I see this API being useful in apps such as Desktop Folder to give background protection. We should also probably set some contrast guidelines like that if you use this effect, the window's opacity should be around 0.8 or something along those lines.

danrabbit:
Just a technical comment, not a design comment. It looks like the blur isn’t applied in multitasking view

danrabbit:
Design comment: something I remember being mentioned in Slack is we could restrict the API to being accessible to Panel and Dock type windows only. That would cater to apps like the desktop folder and wingpanel without exposing ourselves to people doing macOS style sidebars and such

alex285:
@danrabbit 
> without exposing ourselves to people doing macOS style sidebars and such

Just some feedback. If elementary can have such functionality, why not let developers to use it if they want to, and see what they can create? They may come with good ideas. And if they dont like it, they simple wont use it

cassidyjames:
I forgot to copy a summary of our private discussion over here, but here's a summary to the best of my memory:

- Restricting to "system" UI might be a good compromise or first start, so we can get the background protection in things like the panel. Like Dan said.

- We also briefly looked at using it for modal system UI, which all I could think of at the time was the shutdown dialog. Blurring out the desktop was actually in an early mockup, possibly from Harvey, and that could work. But it would have to be a completely system modal thing, and I don't think we'd want arbitrary apps doing that. Right now that also has to be implemented with a large transparent window—if we were to explore this further, I'd want a window to be able to mark itself as system modal somehow and just have Gala take care of the blurring. Another place we briefly thought was Agent dialogs, but I don't know that we want them to be modal. Shell does this and it can be annoying.

- I still want a pretty thorough performance and power consumption comparison at some point so we can be better informed.

I want to reiterate that this is super impressive work as it stands, and my hesitation/resistance to use it should not diminish that! I just want to make sure we're making the best long term design and implementation decisions for elementary OS.

donadigo:
@zzag yes, you're right, here we can't animate the amount of blur. Now, I don't think there's really a need to support it, I think animating opacity is enough here to make the animation appealing.

I chose this path because:
1. it is one of the faster algorithms you can use for this type of effect and doesn't affect performance
2. we won't probably change the blur amount anywhere in the shell / apps

donadigo:
@danrabbit fixed.

zzag:
> we won't probably change the blur amount anywhere in the shell / apps

Well, that's not good engineering practice to hard code parameters. Newcomers [other developers] will be wondering what these numbers are and where they come from. I suggest to add a deterministic procedure to generate parameters, something like https://github.com/pixijs/pixi-filters/blob/master/filters/kawase-blur/src/KawaseBlurFilter.js#L87

name-route:
If this makes through (even just for shutdown dialog) please make it optional (I really like how it is now).

donadigo:
@cassidyjames here are some timing stats via https://github.com/elementary/gala/tree/optimized-blur-perf-tests:

Enabling effect causes a +30 MB increase of GPU memory for a 1920x1080 screen size (4 sample iterations - current setting). This would get bigger for bigger displays of course.

Hardware:
CPU: Intel i5 7500
GPU: NVidia GeForce GTX 1050 Ti

**1. Renderer: VM (VMware)**

1 paint cycle | Average (ms) | Minimum (ms) | Maximum (ms)
-- | -- | -- | --
CPU time | 0.21 | 0.11 | 0.36
GPU time | 0.35 | 0.30 | 0.45

**2. Renderer: NVidia Geforce GTX 1050 Ti, propertiary drivers**

1 paint cycle | Average (ms) | Minimum (ms) | Maximum (ms)
-- | -- | -- | --
CPU time | 0.61 | 0.53 | 0.67
GPU time | 0.20 | 0.20 | 0.21

**3.  NVidia Geforce GTX 1050 Ti, noveau drivers**

1 paint cycle | Average (ms) | Minimum (ms) | Maximum (ms)
-- | -- | -- | --
CPU time | 0.27 | 0.22 | 1.14
GPU time | 2.25 | 2.15 | 2.48

(I didn't test this with integrated GPU's like Intel yet)

In conclusion: the effect performs the worst on systems without any or noveau drivers. 2 milliseconds or above is far too much time spent on the effect, and it also causes noticeable lag. The effect should take up to say max 0.6 - 0.8 milliseconds.

VM's probably should also have the effect disabled.

vinicioslc:
I need this add plox  add plox *-*

donadigo:
@cassidyjames I've pushed some optimizations to the branch and it now doesn't take as much power to render for small windows (like the panel) since we now downsample and upsample only the region that we care about, not the entire texture, so the above measurements have to be re-done at some point because they are no longer accurate.

Regarding the first point, I know how that would be done and it's a simple fix, but I don't want to push that before we actually merge it because that would mean that the demo would stop working and people wouldn't be able to test.

Second: It's actually kind of how it is right now, if someone wants the blur, they request it from the WM via DBus. Enabling that for all modal windows would mean that we also have to figure out which window actually is modal, so I would prefer to just leave the API and let our services and thrid party apps request the effect themselves. It also gives more control to the caller, since it knows if the effect was added or not, and based on that it can decide whether to apply any special styling or remove it.

Third: Needs someone that would run the perf-tests branch and paste the output here I guess. I may at some point figure out how to switch to the Intel GPU from NVidia here on my system though.

Forth: done, needs testing from people who have different hardware.

mishamyrt:
I can help with tests on Intel, if you give instructions

donadigo:
@mishamyrt if you want you can clone Gala master and pull this branch here: https://github.com/elementary/gala/tree/optimized-blur-perf-tests

Then just follow the instructions in the README: https://github.com/elementary/gala/blob/optimized-blur-perf-tests/README.md (you can install all the needed deps by doing `sudo apt build-dep gala`)

After that, restart the current gala instance by doing `(gala -r &) &` (this ensures that when you close the terminal, the gala process won't be killed).

You can then compile the sample that uses the API: https://gist.github.com/donadigo/6e93d4730b66f1a6f35100d43317fe18, with `valac --pkg=gtk+-3.0 --pkg=gdk-x11-3.0 BlurWindow.vala` 

After compiling, launch the example `./BlurWindow`, maximize the window and the gala process should print information about the GPU / CPU times.
It would be great if you just pasted one paint cycle measurements here. Here's a sample:
```
Paint cycle:
CPU Avg. time: 0,164450
CPU Min. time: 0,150000
CPU Max. time: 0,222000

GPU Avg. time: 0,142848
GPU Min. time: 0,138240
GPU Max. time: 0,144384
```

mishamyrt:
Here it is:

`Paint cycle:
CPU Avg. time: 0,328373
CPU Min. time: 0,144000
CPU Max. time: 0,591000

GPU Avg. time: 0,363043
GPU Min. time: 0,341280
GPU Max. time: 0,413760`

Tested on Asus UX305FA, Intel Core M-5Y10c

donadigo:
@mishamyrt great, thanks for testing! Now, could you try open a lot of those windows and generally see if there's any noticeable slow-down while using the system?

mishamyrt:
No, everything feels great. I even have made terminal window blurred with d-feet and there are no lags. I don't see the difference at all. Also tried to do it with plank and wingpanel, but that apps is not ready for this.

mishamyrt:
Gnome system monitor screenshot

This is with 11 blurred windows running. I think it's great performance for mobile CPU.

donadigo:
@mishamyrt alright, thank you!

donadigo:
@cassidyjames so @mishamyrt performed some tests on the Intel and it looks like the effect works smoothly there. I guess now it's up to the design team if we want to use the effect and if yes, when and in what places.

MasonLeeBack:
Hey! Nice work on getting blur working in Gala.
However, I seem to be having an issue whenever I try to run the sample- blur doesn't seem to work. I get a transparent window with objects, but no blur.

I'm running on Ubuntu 18.04 with a GeForce GTX 1060 (notebook variant).
Any help would be appreciated (I can provide logs if necessary)

donadigo:
@MasonLeeBack hello! The effect should definitely work on such card, it works here on GTX 1050 Ti. Did you make sure to install gala and restart it? (You can also restart the whole system if you want)

It would be really helpful if you could paste the terminal output of launching the BlurWindow since it usually outputs the error that happened when requesting the effect.

MasonLeeBack:
Yep! I made sure to restart just in case, so gala should be replaced. There is no log coming from BlurWindow, however if you disable the effect you do get this in the log:

`
** (blur:27070): CRITICAL **: 23:42:08.895: file /home/masonleeback/Desktop/blur.vala.c: line 627: uncaught error: GDBus.Error:org.freedesktop.DBus.Error.Failed: Blur actor was not found for the specified window ID (g-dbus-error-quark, 0)
`

jantrushar:
Updated my system and retestet (I am on arch right now). It is working perfectly now. Sorry for false report.

donadigo:
@jantrushar thank you for investigating. This made me reconsider the dock calculations in the `copy_target_texture` method and they're much simpler now.

guidedlinux:
Hi, I really like the idea of having this blur algorithm implemented in Gala, but the demo isn't working for me.
I'm new to vala and elementary but I wanted to try this out, but it seems like the CSS rules are being ignored. I don't get a tinted or transparent window. I am using [this](https://gist.github.com/donadigo/6e93d4730b66f1a6f35100d43317fe18) example. I get no error messages whatsoever.
I don't know if this issue is even related to this PR, sorry if it isn't. I am using an Intel Core i7-6700 and a GTX 950 with proprietary drivers.

screenshot from 2018-06-23 18 04 53

donadigo:
@guidedlinux I have updated the gist for the demo, but I'm not sure if that'll work for you. It looks like the effect is applied, it's just not visible because the window is fully opaque. You can try the new demo version at the same link: https://gist.github.com/donadigo/6e93d4730b66f1a6f35100d43317fe18

jantrushar:
@donadigo it is running very good for me. Great job.

` float width = blur_clip_rect.width > 0 ? blur_clip_rect.width : rect.width;
float height = blur_clip_rect.height > 0 ? blur_clip_rect.height : rect.height;
`
Maybe change this to `> -1` so you can distinguish between "0 wanted for clip" and "don't use clip".

One missing piece for the future is the possibility to add blur for non-recangular windows and also to not blur behind shadows (e.g. menu popups with shadows and round corners).

guidedlinux:
Thanks for your answer @donadigo :)
Maybe I have some outdated versions of the required dependencies installed (as I said, I'm new to vala) so it wasn't working. I have changed "window" to "GtkWindow" in the CSS and now it's working fine (same for the button). I really love the effect and the performance is also good. Thanks again for your help.

screenshot from 2018-06-24 07 35 30

cassidyjames:
@donadigo hey following up from Slack.

So I finally just was able to vocalize some sort of **potential** human interface guidelines around blur…

If something is _already_ transparent or translucent (and would generally make sense to be without any blur), then blur can be used to enhance contrast.

If something is explicitly being de-prioritized, i.e. the background in a system-modal context, it could be blurred to further draw attention to the non-blurred foreground (like lowering brightness of the background). But do NOT use it in a foreground context where no transparency/blur would be just as effective.

Like, I think it works fine in this sort of context: https://cassidyjames.github.io/appage/com.github.cassidyjames.principles/

This is effectively a modal, where the background is deprioritized and blurred out. But it would be dumb to reverse it, making the background normal and the dialog a super blurred out "glass" dialog.

(I was just looking at macOS and Windows blurs recently, and they both do the latter, which makes it literally nauseating and completely unnecessary)

gabeps88:
Is not working for me, I can compile it but the blur is not working, I got this error in terminal

** (BlurWindow:5575): CRITICAL **: file /home/user/BlurWindow.vala.c: line 457: uncaught error: GDBus.Error:org.freedesktop.DBus.Error.UnknownMethod: No such method 'EnableBlurBehind' (g-dbus-error-quark, 19)

gabeps88:
screenshot from 2018-10-03 20 50 26

donadigo:
@gabeps88 you need to install and restart gala to be able to actually see the demo. 
Before you launch the demo, do `sudo ninja install` and run `(gala -r &) &`. Then run the demo.

gabeps88:
@donadigo  it didn't work
screenshot from 2018-10-05 21 00 42

donadigo:
@gabeps88 you have to be in the directory that you cloned this repo to. So `cd` into it before those commands. Did you even 
clone and compile this gala branch? The demo isn't the entire thing, it just uses functions from this branch.

ciampolo:
Any chance you will implement the blur effect for the context menus as well? Would be really awesome setting context menus to like 50% transparency and then the background.

naaando:
Looks good with lyrics 😎
P.S.: I know this is against elementary's HIG, was just a experiment.

blurred lyrics

alex47:
sorry for being offtopic again

@naaando are you sure about that?

untitled

donadigo:
@alex47 he has probably used a lower opacity for the effect in the DBus call, @naaando for going full opaque you should use 255 for opacity.

naaando:
Yep, I was using lower opacity.
Using full opacity on blur

## On content
**Lyrics white theme: #fffffff, 20% opacity**
lyrics-white-blurred-on-content
**Lyrics dark theme: #000000, 20% opacity**
lyrics-dark-blurred-on-content
## On background
**Lyrics dark theme: #000000, 20% opacity**
lyrics-dark-blurred-on-bg

donadigo:
@jeremypw thanks. I'm sorry, but I'm not really interested in updating this branch, until we get a clear yes from the UX team that this is something they want. I've updated this branch may times over the years now and it always takes a lot of time to do so. So it's better to wait for the final confirmation or decline.

jeremypw:
@donadigo Yes, I agree you should not spend more time on it if the UX team are not onboard.  @danrabbit Is this something that has any potential future use?

jeremypw:
I guess this can be closed now as it has attracted little interest recently.

PR #229, opened 2018-04-19
Add an action to switch to a new workspace

donadigo:
Exposes a new action through DBus to switch to a new workspace, useful for https://github.com/elementary/switchboard-plug-pantheon-shell/issues/98


donadigo:
@ricotz yeah, since when you have dynamic workspaces last is always empty.

I didn't decide to check if we have dynamic workspaces or not since I don't think it would make a difference anyway.

We could also change the enum value name, yes, it may be confusing at first.

donadigo:
@ricotz done.

PR #238, opened 2018-04-28
Add depencency installation details to README

jwinnie:


donadigo:
Perhaps `apt build-dep gala` would be better here.

danrabbit:
Hey thanks for your branch. Unfortunately I'm going to close this since we don't include package-manager specific directions in our READMEs. Package names may also be different across releases so this isn't guaranteed to always work even in Debian-based systems

PR #239, opened 2018-04-28
Fix maximize / minimize animations

### LINKED ISSUE #75: [Juno] Smooth maximize/tile [$20]
It looks like something changed with our Juno builds, as we no longer get smooth maximizing and tiling. 

---
There is a bounty on this issue on [Bountysource](https://www.bountysource.com/issues/49390506-juno-smooth-maximize-tile)

donadigo:
Fixes #75.

It looks like that some internal changes in mutter caused `get_frame_rect` to return not updated new size of the window resizing. (This is because for some reason the frame size is updated after the actual `size_change` method in the plugin is called: https://github.com/GNOME/mutter/blob/20176d0395294b9eeb7f0e28f1a1a35705fa0888/src/core/window.c#L2866)

That's why the animation happened, but there was no scale animation visible, only opacity. Waiting for the `size_changed` signal works here.

### DIFF
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -821,7 +821,7 @@ namespace Gala
 
 		public override void size_change (Meta.WindowActor actor, Meta.SizeChange which_change, Meta.Rectangle old_frame_rect, Meta.Rectangle old_buffer_rect)
 		{
-			var window = actor.get_meta_window ();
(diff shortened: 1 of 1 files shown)


danrabbit:
Can confirm. Works as expected :tada:

PR #241, opened 2018-04-29
Add timeout to dock animation

### LINKED ISSUE #236: Only show alt+tab animation when holding keys
The Alt+Tab animation is great. However, it gets annoying when quickly cycling through windows.

alt tabretry

Maybe there should be a timeout before the animation is triggered.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/57693036-only-show-alt-tab-animation-when-holding-keys?utm_campaign=plugin&utm_content=tracker%2F62872462&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F62872462&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jwinnie:
Partially fixes #236.
Adds a timeout to the alt+tab dock animation, and cancels the timeout on `close()`. That way the dock animation doesn't show until you hold down Alt+Tab for some time.

### DIFF
--- a/src/Widgets/WindowSwitcher.vala
+++ b/src/Widgets/WindowSwitcher.vala
@@ -26,6 +26,9 @@ namespace Gala
 		const float BACKGROUND_OPACITY = 155.0f;
 		const float DIM_WINDOW_BRIGHTNESS = -BACKGROUND_OPACITY / 255.0f;
 
+		uint animate_dock_width_timeout;
+		uint place_dock_timeout;
+
 		public WindowManager wm { get; construct; }
 
 		WindowIcon? current_window = null;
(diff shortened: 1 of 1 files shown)


jwinnie:
**Update**: `show_background` is now also on timeout.

I couldn't get `dim_windows` inside of the timeout because the windows kept flickering for some reason.

jwinnie:
Setting the `dim_windows` opacity value to `0` also causes the flicker issue. I can’t figure it out; does anyone have an idea? It would be really cool if quick Alt+Tab switching worked seamlessly.

cassidyjames:
@jwinnie is this ready for review, or still in progress?

jwinnie:
This patch is abandoned since I don't use elementaryOS anymore and I don't know of any way to test this from my current environment. I'm very sorry. Maybe someone else can take this patch up.

jwinnie:
What I recall is that the quick window switching worked but a random window flickers for a few milliseconds before you get to the right window. I was never able to figure out the reason behind the flickering.

PR #242, opened 2018-05-05
HiDPI Workspace Switcher

### LINKED ISSUE #101: Workspace Switcher doesn't scale for HiDPI [$50]
Like #31, the Workspace Switcher also doesn't scale by the UI scaling factor, meaning it's tiny on HiDPI.

davidmhewitt:
Fixes #101 

Most of this work was done in a branch by @donadigo , but I'm opening a PR for visibility. I've fixed a couple of bits that looked wrong after testing on a HiDPI display. Still needs some in depth looking at compared to a non-HiDPI display though. I can't tell if some of the animations look a bit funky when you're dropping them on a new workspace or if they were always like that.

### DIFF
--- a/src/Widgets/IconGroup.vala
+++ b/src/Widgets/IconGroup.vala
@@ -96,12 +96,15 @@ namespace Gala
 
 		construct
 		{
-			width = SIZE;
-			height = SIZE;
+			var scale = InternalUtils.get_ui_scaling_factor ();
+			var size = SIZE * scale;
+
+			width = size;
+			height = size;
 			reactive = true;
 
 			var canvas = new Canvas ();
-			canvas.set_size (SIZE, SIZE);
+			canvas.set_size (size, size);
 			canvas.draw.connect (draw);
 			content = canvas;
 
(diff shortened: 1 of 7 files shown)


danrabbit:
Right on! This is definitely way closer. The only real issue I'm noticing is in the switcher itself at the bottom of the display. The icons are the right size, but the wrong scale. So where it should be a 64px icon @2x, it's a 128px icon. And the line widths for the workspace "folder" border things are 1px (looks like half a pixel in HiDPI) instead of 2px (looks like 1px in HiDPI).

davidmhewitt:
@danrabbit Thanks! Think I've got both of those things. I'll let you cast your more finely tuned wrong icon spotting eye over it again.

danrabbit:
Nice! There's just one small issue and that's that the "shadow" of the workspace tile folder thingy isn't offset by the correct number of pixels so it's covered up partially by the border now. You can see that the corners look brighter because of this. Otherwise great work!

davidmhewitt:
There's now one other thing I've noticed, if I have 5 windows in a workspace, it seems it doesn't leave enough space in the box at the bottom for the 5 icons.

I'll have a look what's causing that when I can, but I'm just about to grab some dinner.

davidmhewitt:
On second thoughts, this may already be a "feature". If I switch my scaling factor to 1, I notice I can reproduce this behavior. Though, I don't have a non HiDPI monitor handy to test master to check if it occurs there too. So this may be ready to review again.

Issue #104, opened 2017-10-08
Margin around some applications when using quicktile


peteruithoven:
I've been using [quicktile](https://github.com/ssokolow/quicktile) because it enables quarter tiling (https://github.com/elementary/gala/issues/105) and it enables resizing snapped windows (https://github.com/elementary/gala/issues/46).
One issue is that a lot of Elementary Apps like the Terminal, Videos, Files, Quilter, Webpin have a big margin around them when being tiled by quicktile. Applications like Firefox and Atom don't have this margin. 
![33fe922a-e856-11e4-8d6f-71f2feff2efd](https://user-images.githubusercontent.com/523210/31317653-227b8a2a-ac45-11e7-8600-b605588ae600.png)

There was an issue opened at the quicktile repo already: https://github.com/ssokolow/quicktile/issues/49

From this issue I understand that quicktile asks the window manager (Gala in this case?) to resize. 

I'm assuming this has something to do with the wide shadow, that also appears when taking screenshot of applications. 

Is there something Gala can do?

Philip-Scott:
This is something that must be fixed In quicktile. GTK windows have a shadow that gets concidered part of the window, while other non-native apps don't. So the Developer of quicktile must take that into account.

peteruithoven:
Thanks for the quick answer, I'll forward the tip to the Quicktile repo.

peteruithoven:
@Philip-Scott do you happen to know if `libwnck` takes this into account? The developer of quicktile [responded](https://github.com/ssokolow/quicktile/issues/49#issuecomment-335065834) with:

> I'll double-check when I have time to address this, but QuickTile should be relying on libwnck (a GNOME-family library, just like GTK+) for getting and setting window geometry specifically to avoid this kind of "What definition of 'window geometry'?" mess.

cassidyjames:
There was a change in window geometry between GTK 3.18 and GTK 3.22 that I believe should have fixed this. It fixed window size requests for a lot of things in GTK itself, no longer allocating the shadows as part of the window.

ssokolow:
Makes sense.

If it turns out that *doesn't* fix it, I'm hoping to find time to port to GTK+ 3.x to account for libwnck 2.x being pulled from Ubuntu 18.x as unmaintained. (Something about QuickTile triggers a segfault in the PyGTK API porting shim on *buntu 14.04 LTS, so it's a bit of an involved process to do the port.)

PR #250, opened 2018-05-15
Render shadows using Gtk

tintou:
It's not ready to merge until @elementary/ux fixes the CSS ;)

### DIFF
--- a/src/Widgets/WindowClone.vala
+++ b/src/Widgets/WindowClone.vala
@@ -187,6 +187,9 @@ namespace Gala
 		{
 			window.unmanaged.disconnect (unmanaged);
 			window.notify["on-all-workspaces"].disconnect (on_all_workspaces_changed);
+			window.notify["fullscreen"].disconnect (check_shadow_requirements);
+			window.notify["maximized-horizontally"].disconnect (check_shadow_requirements);
+			window.notify["maximized-vertically"].disconnect (check_shadow_requirements);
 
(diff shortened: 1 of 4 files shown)


danrabbit:
Can we get an additional style class for the workspace itself so we can style that differently than the windows. Maybe `.decoration.workspace`?

danrabbit:
Hm something I noticed is that it seems like only maximized windows actually need additional box shadow. Other windows already have it, including SSD windows.

screenshot from 2018-05-15 09 37 28 2x

peteruithoven:
Any change this will influence https://github.com/elementary/gala/issues/104 ? Because libwnck might get a different window geometry?

cassidyjames:
@peteruithoven no, this shouldn't affect that since it's only in the window overview.

src/Widgets/WindowClone.vala
```diff
@@ -136,6 +136,9 @@ namespace Gala
 
 			window.unmanaged.connect (unmanaged);
 			window.notify["on-all-workspaces"].connect (on_all_workspaces_changed);
+			window.notify["fullscreen"].connect (check_shadow_requirements);
+			window.notify["maximized-horizontally"].connect (check_shadow_requirements);
+			window.notify["maximized-vertically"].connect (check_shadow_requirements);
```
donadigo: You should disconnect all those signals in the destructor method.

PR #251, opened 2018-05-16
pip: Do not hide the window when it is on another monitor

### LINKED ISSUE #249: PIP: Only hide when both the window and the PIP view are in the same monitor
### Expected Behavior
PIP should only hide if the window is also on the same monitor as the PIP view. We're hiding the view to avoid covering the same window anyways, and if it's on a separate display then we shouldn't have to worry about covering the window 

### Story Time!
I had a presentation in class and wanted to use PIP to show parts of different windows on the projector while i was controlling them on my main screen. Problem is that when you focus into the window the PIP view disappears even if it's on the same monitor.

donadigo:
This should fix #249.

I'm not sure if this does the thing it's supposed to do. Someone with multi monitors has to test and review this.


Philip-Scott:
Hmmm i tried it out and it wasn't really working, the PIP view still disappears on the external display.... My guess is that `screen.get_monitor_geometry` is returning the whole virtual display area with both monitors and not just the one, or maybe `container.get_content_box` is returning relative cords to the PipWindow instead of on the display. (I can do some deeper debugging probably tomorrow morning)

plugins/pip/PopupWindow.vala
```diff
@@ -299,15 +299,26 @@ public class Gala.Plugins.PIP.PopupWindow : Clutter.Actor
...
+		var monitor_rect = screen.get_monitor_geometry (window.get_monitor ());
+		var actor_box = container.get_content_box ();
+		Meta.Rectangle container_rect = {
+			(int)actor_box.get_x (),
+			(int)actor_box.get_y (),
+			(int)actor_box.get_width (),
+			(int)actor_box.get_height ()
+		};
+
+		if (window.appears_focused && monitor_rect.intersect (container_rect, out container_rect)) {
```
Philip-Scott: Hmmm, maybe it's also failing since it's using the same variable for in and out :man_shrugging:

plugins/pip/PopupWindow.vala
```diff
@@ -299,15 +299,26 @@ public class Gala.Plugins.PIP.PopupWindow : Clutter.Actor
...
 			return;
 		}
 
 		var window = window_actor.get_meta_window ();
-		if (window.appears_focused) {
+
+		var monitor_rect = screen.get_monitor_geometry (window.get_monitor ());
+		var actor_box = container.get_content_box ();
+		Meta.Rectangle container_rect = {
+			(int)actor_box.get_x (),
```
Philip-Scott: space after `(int)`?

danirabbit:

PR #255, opened 2018-05-20
Disallow translating icon name

### LINKED ISSUE #254: gala-multitaskingview have translated icon name
You can see this in /usr/share/applications/gala-multitaskingview.desktop
```
Icon[ca]=Vista de multitasca
Icon[da]=multitasking-view
Icon[en_GB]=multitasking-view
Icon[es]=vista-multitarea
Icon[fi]=multitasking-view
Icon[fr]=multitasking-view
Icon[id]=pancatugas-tampilan
Icon[it]=vista-multitasking
Icon[ja]=multitasking-view
Icon[ko]=멀티태스킹 뷰
Icon[lt]=multitasking-view
Icon[nl]=multitasking-weergave
Icon[pl]=podglad-wielozadaniowy
Icon[ru]=Многозадачный вид
Icon[vi]=multitasking-view
Icon[zh_CN]=多任务视图
Icon[zh_TW]=多工作業-檢視
Icon=multitasking-view
```
And because of this we have entry without an icon if we use one of this languages.
image

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/58603909-gala-multitaskingview-have-translated-icon-name?utm_campaign=plugin&utm_content=tracker%2F62872462&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F62872462&utm_medium=issues&utm_source=github).
</bountysource-plugin>

davidmhewitt:
Fixes #254 

I think this is the best way to fix this, due to the fact that I don't think we can pass args to `msgfmt` in meson (yet?). But it would be good for @tintou to confirm.

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -8,22 +8,42 @@ gala_gschema = configure_file(
 
 gala_gschema_compile = gnome.compile_schemas()
 
-i18n.merge_file(
-	input: 'gala-multitaskingview.desktop.in',
-	output: 'gala-multitaskingview.desktop',
-	po_dir: join_paths(meson.source_root (), 'po'),
-	type: 'desktop',
-	install: true,
-	install_dir: join_paths(data_dir, 'applications')
+msgfmt = find_program('msgfmt')
+
+custom_target('gala-multitaskingview.desktop_merge',
+    input: 'gala-multitaskingview.desktop.in',
+    output: 'gala-multitaskingview.desktop',
+    command : [msgfmt,
+               '--desktop',
+               '--keyword=Comment',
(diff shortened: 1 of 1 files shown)


davidmhewitt:
Going to close this branch as I think it's unrealistic that we'd use this method in every repo to prevent translation of the icon name. A comment for translators is probably a better short term method, and I'll look into how possible it would be to submit a PR to meson to make this same functionality easier.

PR #276, opened 2018-06-19
Fix background position for scaled actors (e.g. when animating scale)

jantrushar:
Small change to fix blur background position when window is animating (scaling).


jantrushar:
Welcome! I appreciate your work very much!

PR #278, opened 2018-06-21
Don't remove workspaces on switching in multitasking view

### LINKED ISSUE #277: Don't remove empty workspaces when in workspace view
Automatically removing empty workspaces results in a few bad things:
* you can't freely rearrage apps how you want because as soon as you move only app from a workspace and switch to another it gets deleted
* you can't "swap" apps for their workspaces: say I want to move app A to workspace B, and app B to workspace A, this isn't possible since as soon as you switch to workspace B, workspace A is deleted
* it causes unnecessary animation shift in the view itself

Empty workspaces should be cleaned up when the user actually quits the workspace view.

donadigo:
Fixes #277.

Prevents from aggressive removing of workspaces when switching between them in the multitasking view.
All empty workspaces get cleaned up when the multitasking view is closed except active one, so if the user enters a non-last empty workspace, it doesn't get removed.


donadigo:
Now ready to review.

donadigo:
Ready to review again.

jeremypw:
@donadigo Converting to draft as has conflicts, unaddressed issues and no recent activity.  Please feel free to close or reopen for review after fixing conflicts as desired.

danirabbit:

PR #314, opened 2018-08-08
Implement EndSessionManager

### LINKED ISSUE #324: Logout shortcut doesn't work
When I press Ctrl+Alt+Delete nothing happen

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/62773257-logout-shortcut-doesn-t-work?utm_campaign=plugin&utm_content=tracker%2F62872462&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F62872462&utm_medium=issues&utm_source=github).
</bountysource-plugin>

donadigo:
Fixes #324 

- [x] Requires https://github.com/elementary/wingpanel-indicator-session/pull/58 to be merged.

Implements `org.gnome.SessionManager.EndSessionDialog` and forwards DBus calls to the session indicator in wingpanel. This is done so that we have consistency when it comes to owning GNOME specific DBus interfaces.

Worth mentioning is also the original reference implementation of the same interface in Budgie Desktop: https://github.com/solus-project/budgie-desktop/blob/master/src/wm/shim.vala

### DIFF
--- a/src/DBus.vala
+++ b/src/DBus.vala
@@ -54,6 +54,15 @@ namespace Gala
 				() => {},
 				() => {},
 				() => critical ("Could not acquire name") );
+
+			Bus.own_name (BusType.SESSION, "org.gnome.SessionManager.EndSessionDialog", BusNameOwnerFlags.NONE,
+				(connection) => {
+					try {
+						connection.register_object ("/org/gnome/SessionManager/EndSessionDialog", SessionManager.init ());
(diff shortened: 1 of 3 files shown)


danrabbit:
I can confirm this works as expected with the sibling indicator session branch :+1:

PR #327, opened 2018-08-30
Fix #326 by repositioning wallpapers during transition

### LINKED ISSUE #326: Transition glitch in multi-monitor setup
I am using a dual monitor setup that looks like this:

photo

When I switch workspaces using Meta+Left/Right, the wallpaper is vertically offset during the transition:

screenshot von 2018-08-30 10 46 16

This seems to happen because my primary screen does not start at (x=0, y=0). The top of the left screen is lower than the top of the right screen (which starts at y=0), and this offset seems to be added *again* when drawing the wallpaper during the transition.

If I arrange both screens so that they start at y=0, the bug goes away.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/62795440-transition-glitch-in-multi-monitor-setup?utm_campaign=plugin&utm_content=tracker%2F62872462&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F62872462&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jlnr:
It's still broken when `workspaces-only-on-primary` is disabled, but at least it seems to fix #326 with the default settings.

### DIFF
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -1526,6 +1526,7 @@ namespace Gala
 			if (move_primary_only) {
 				wallpaper = background_group.get_child_at_index (primary);
 				wallpaper.set_data<int> ("prev-x", (int) wallpaper.x);
+				wallpaper.set_data<int> ("prev-y", (int) wallpaper.y);
 			} else
 				wallpaper = background_group;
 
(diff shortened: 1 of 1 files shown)


jlnr:
Thanks for the review! 👍 Should I wait for a second review or can I go ahead and merge this?

tintou:
@jlnr : @ricotz is the maintainer of Gala and merges things manually :slightly_smiling_face:

Issue #21, opened 2017-06-13
Graphical glitches when resuming from suspension


SamCarlberg:
This has been an off-and-on problem for me. Generally it seems to be fixed (or broken) every few months by OS updates from AppCenter

### Desktop view

Look at the corners; there are small ~6px squares in each corner filled with noise

![screenshot from 2017-06-13 13-16-11](https://user-images.githubusercontent.com/6320992/27095379-6ffc632a-503b-11e7-9f6d-f6c901d0ca36.png)

And this occasionally happens to the desktop:

![screenshot from 2017-06-16 14 03 21](https://user-images.githubusercontent.com/6320992/27239125-a52bba1e-529d-11e7-8a85-15ba61741ef1.png)

### Multitasking view

Don't know what's happening here. If Gala is tiling the same image for the background, it might be reading from junk memory after resuming.

![screenshot from 2017-06-13 13 17 28](https://user-images.githubusercontent.com/6320992/27095378-6ff8edb2-503b-11e7-9b12-91daa02d36b6.png)

kmf:
I have similar issues, after I lock my screen and unlock again,

![2017-09-08-17 10 19](https://user-images.githubusercontent.com/78700/30242666-dca94d18-959a-11e7-9337-bc9cc15128d2.png)

![2017-09-09-19 53 35](https://user-images.githubusercontent.com/78700/30242667-f0ca5b0c-959a-11e7-9504-c1f0be536dfb.png)

https://www.reddit.com/r/elementaryos/comments/6dxx67/disappearing_login_sessions_when_locked/
```
Run sudo -i scratch-text-editor /etc/default/acpi-support
Find and change POST_VIDEO=true to POST_VIDEO=false
Find and change SAVE_VBE_STATE=true to SAVE_VBE_STATE=false
Restart your laptop, issue should be solved.
```

kmf:
could resolve it with a 
```
gala --replace
```

Nine-H:
I've had this issue on my PC's nvidia card, but haven't seen it on my intel laptop.

peteruithoven:
@kmf How would you recommend running that?
I noticed that when I try that from the Terminal it's running "in the foreground", meaning you can't use that terminal anymore and can't close it. 
I would recommend running it from the Applications menu.

Relevant Stack overflow question: https://elementaryos.stackexchange.com/questions/8952/artifacts-in-background (which has many duplicates)

peteruithoven:
[...]

PR #330, opened 2018-09-08
Nvidia drivers: Refresh background and screen corners (#21)

jlnr:
The proprietary Nvidia drivers discard FBOs when the system suspends from resume (maybe on all sorts of modesets?). The current workaround for this was added in Feb 2016:

https://github.com/elementary/gala/commit/4b001cad523e67e496ed04c2d6fd7222cd9be9f7 

This correctly causes the wallpaper images to be reloaded when resuming from sleep mode, but the multitasking background (SystemBackground) and the round corners are still garbled here (see screenshots in #21).

The Bugzilla issue for the same issue in GNOME Shell has more details:

https://bugzilla.gnome.org/show_bug.cgi?id=739178

There's now a `Display.gl_video_memory_purged` signal that allows us to handle the loss of OpenGL resources without parsing Nvidia vendor strings. I've used this signal to fix wallpaper images and the rounded corners plugin.

This did not fix the SystemBackground. Looking at [meta-background.c](https://gitlab.gnome.org/GNOME/mutter/blob/master/src/compositor/meta-background.c), I think the issue is that `G_DESKTOP_BACKGROUND_STYLE_WALLPAPER` backgrounds are not fully discarded by `Meta.Background.refresh_all`. Either way, simply calling the color setter seems to fix the issue.

All tested on Juno with a GTX 1080 (not mobile) and the Nvidia 390 drivers from AppCenter. Without this patch, I can reproduce garbled textures after every sleep-resume cycle. Now everything seems to work fine.


jlnr:
Testing this required a few reboots because `gala --replace` stopped working for me. The wingpanel never came back afterwards. Anyone else seeing that?

src/WindowManager.vala
```diff
@@ -121,18 +98,15 @@ namespace Gala
...
-					warning ("Failed to get LoginD proxy: %s", e.message);
-				}
-			}
+#if HAS_MUTTER322
+			get_screen ().get_display ().gl_video_memory_purged.connect (refresh_backgrounds);
+#endif
+		}
 
-			Bus.watch_name (BusType.SESSION, DAEMON_DBUS_NAME, BusNameWatcherFlags.NONE, daemon_appeared, lost_daemon);
+		void refresh_backgrounds ()
```
donadigo: This method should be static as well as compiled when HAS_MUTTER322.

jlnr:
Thanks for the review @donadigo.

Instead of having two `#ifdef` blocks right next to each other, I've turned the whole `refresh_backgrounds` method into a closure, which I find easier to parse. What do you think?

(I can't re-test this branch right now because I'm travelling without my Nvidia box.)

PR #332, opened 2018-09-10
Install autostart .desktop to the right place

cassidyjames:
For #179


davidmhewitt:
I think this needs to be rolled back and a separate PR made to https://github.com/elementary/session-settings to modify the session file to add it as a required component, like the message in #179 says.

davidmhewitt:
While this may work, it's probably better to have the daemon launched by the session manager, as it should handle restarting the daemon better should anything happen to it.

PR #339, opened 2018-09-20
Supposedly fix the issue with server side context menus

### LINKED ISSUE #174: Header context menu on not-native apps

donadigo:
Fixes #174.
It happens so that if we enter modal mode and add an Idle callback, the thing starts to work on server side decorations.

### DIFF
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -877,11 +877,21 @@ namespace Gala
 					if (window.can_close ())
 						flags |= WindowFlags.CAN_CLOSE;
 
-					try {
-						daemon_proxy.show_window_menu.begin (flags, x, y);
-					} catch (Error e) {
-						message ("Error invoking MenuManager: %s", e.message);
-					}
+					Idle.add (() => {
+						unowned Screen screen = get_screen ();
+						var time = screen.get_display ().get_current_time ();
+			
+						begin_modal (0, time);
+						try {
+							daemon_proxy.show_window_menu.begin (flags, x, y);
+						} catch (Error e) {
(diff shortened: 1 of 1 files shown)


jlnr:
I've been running this branch for an hour now and I'm seeing a weird glitch where I invoke the window menu on the same window several times, and the menus has different contents - it alternates between the shorter menu where "[x] Show on all workspaces" is checked, and then it goes back to the one where this is unchecked, and where the menu items for "Move to next/previous workspace" are available.

It might be an unrelated glitch, but I've never seen this behavior before.

donadigo:
@jlnr yes, I've also experienced this and tbh I think it does exist in master but I have still to confirm it. This branch didn't touch any of the daemon that handles the menu, just invoking it to show, so the issue is probably in the daemon.

Djaler:
Doesn't work for me :C

danrabbit:
Superseded by #429

PR #346, opened 2018-09-26
Fix "Other" Notifications

cassidyjames:
Should actually fix https://github.com/elementary/default-settings/issues/14 since https://github.com/elementary/default-settings/pull/15 did not.


cassidyjames:
Still something weird here. After installing, the default GSetting is `false`, but the plug still shows it as enabled. Once you manually toggle it to false, though, it works as expected…

danrabbit:
Closing since notifications is its own thing now

PR #357, opened 2018-10-08
Fix multi-monitor windows preventing workspaces from closing automatically 

### LINKED ISSUE #347: Bug: workspace does not close on multimonitor setup
I've found a small issue, which I think is a bug. If on the main monitor the workspace is empty, it will not close if there is a window active on the second monitor. In a multi monitor setup, only the main monitor is part of the workspace switching. Therefore a window on the second monitor shouldn't be counted as an active window on the workspace.

I've confirmed this by closing the window on the second monitor: the workspace closes fine.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/63991968-bug-workspace-does-not-close-on-multimonitor-setup?utm_campaign=plugin&utm_content=tracker%2F62872462&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F62872462&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Philip-Scott:
In juno, it appears that windows on the non-primary monitors now return `false` on `window.is_always_on_all_workspaces ()`, this made it so gala thought that there was always a window on in this function. I've modified the code to use the property `on_all_workspaces` instead of the function. This works even if making false the property on `gsettings org/gnome/mutter/workspaces-only-on-primary`

fixes #347

### DIFF
--- a/lib/Utils.vala
+++ b/lib/Utils.vala
@@ -209,7 +209,7 @@ namespace Gala
 		{
 			var n = 0;
 			foreach (weak Meta.Window window in workspace.list_windows ()) {
-				if (window.is_always_on_all_workspaces ())
+				if (window.on_all_workspaces)
 					continue;
 				if (window.window_type == Meta.WindowType.NORMAL ||
 					window.window_type == Meta.WindowType.DIALOG ||


jlnr:
I haven't looked at the code, but this fixes the problem on my machine. :tada:

Philip-Scott:
Modified the code to use the property `on_all_workspaces` instead of the function. This works even if making false the property on gsettings `org/gnome/mutter/workspaces-only-on-primary`

PR #363, opened 2018-10-18
Add missing languages to LINGUAS files

ryonakano:
If we don't add these languages, they won't be available to translate in Weblate, and people who speak these languages and want to translate into them can not submit their translation.


ryonakano:
@leonardo-lemos Could you merge this? I have no right to merge.

leonardo-lemos:
> 
> 
> @leonardo-lemos Could you merge this? I have no right to merge.

Only @ricotz can merge this.

ryonakano:
@ricotz These po files have no translations because they're newly generated and added in this PR. Since I added lacking language codes to the LINGUAS file in this PR, I thought I need to generate and add po files for these languages.

ricotz:
@ryonakano And what is the reason to add the "missing" languages?

ricotz:
@ryonakano I assume this is to enable weblate picking this up and providing these missing languages to be translated. If so, this kind of information should be part of the merge proposal description.

ryonakano:
@ricotz Yes, you're right. if we don't add these languages, they won't be available to translate in Weblate, and people who speak these languages and want to translate into them can not submit their translation.
Yes, I should have written these information first. I'm sorry.

ricotz:
Pushed https://github.com/elementary/gala/commit/dbbfb50f9b95439d0ba79e89325dd4b36c92049b
Let's see if this is enough.

ryonakano:
Thank you.

PR #350, opened 2018-09-28
 New default folder to save screenshots #2

peteruithoven:
Attempt to get https://github.com/elementary/gala/pull/95 rolling again.

PR #366, opened 2018-10-18
Tweaked the screenshot path

WattanaGaming:
This still need a confirmation if it will work. I was to scared to install this over the original Gala on my eOS 5.0


peteruithoven:
Closing this one in favor of the pending #350

Issue #358, opened 2018-10-10
[i18n]Broken icon for gala-multitaskingview.desktop


bagjunggyu:
We still have an issue with broken icon for _gala-multitaskingview.desktop_ till to RC-1.
It can be resolved by deleting translations as **Icon[i18n]=multitasking-view** and leaving only **Icon=multitasking-view** itself in gala-multitaskingview.desktop file.

opnay:
Also Korean (after build added Icon[lang]={translated multitasking-view})
I think to add blacklist  "data/gala-multitaskingview.desktop.in:7" in gala/po

Suzamax:
[I proposed a pull request in order to fix this.](https://github.com/elementary/gala/pull/371)

rscm:
this is currently on my gala-multitaskingview.desktop file @ Juno

> Icon[ca]=Vista de multitasca
> Icon[da]=multitasking-view
> Icon[en_GB]=multitasking-view
> Icon[es]=vista-multitarea
> Icon[fi]=multitasking-view
> Icon[fr]=multitasking-view
> Icon[id]=pancatugas-tampilan
> Icon[it]=vista-multitasking
> Icon[ja]=multitasking-view
> Icon[ko]=멀티태스킹 뷰
> Icon[lt]=multitasking-view
> Icon[nl]=multitasking-weergave
> Icon[pl]=podglad-wielozadaniowy
> Icon[ru]=Многозадачный вид
> Icon[vi]=multitasking-view
> Icon[zh_CN]=多任务视图
> Icon[zh_TW]=多工作業-檢視
> Icon=multitasking-view

we can see in what languages the problem persists... mine is Spanish [es]
I'll try editing the name of the icon now

TheRock892:
same problem in italian languages.. 
..the string Icon in gala-multitaskingview.desktop in many language are have been translated while it should not be ..Icon string should be 'multitasking-view' in all languages!

bagjunggyu:
resolved while upgrade

PR #371, opened 2018-10-19
Fixing Multitasking View icon for some languages

Suzamax:
I noticed the wrong icon showed (refer to https://github.com/elementary/gala/issues/358 issue) so I fixed it.

### DIFF
--- a/po/ca.po
+++ b/po/ca.po
@@ -81,7 +81,7 @@ msgstr "Visualitza totes les finestres obertes i totes les àrees de treball"
 
 #: data/gala-multitaskingview.desktop.in:7
 msgid "multitasking-view"
-msgstr "Vista de multitasca"
+msgstr "multitasking-view"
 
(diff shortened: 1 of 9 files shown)


davidmhewitt:
Thank you for the PR.

However, translations are handled in Weblate (https://l10n.elementary.io/) which pushes them into Git. it is not possible to modify .po files like this as it will be overwritten or cause conflicts with Weblate.

You can join the [translators Slack](https://ele-l10n.slack.com/join/shared_invite/enQtMjkwMjI2Mzk5ODQxLWM3NWZlMjMxMTUyNzg0MjdiNTdkYTM5ZDA3NzE5YTIwMzZmZjhmZjg0MzQwMGE5MjVhMGU2Yjk2MDU1MGZiYTU) and ask for write access to be granted to your Weblate account, if you don't have this already.

I'm going to close this PR as there isn't anything we can do about these translations here.

Issue #341, opened 2018-09-21
Can't change tiling position directly?


peteruithoven:
It looks like applications in Juno that are already tiled to the left or right can't directly be tiled to the opposite side. It seems to go into a weird state where only some corners turn round. 
![screencast appcenter webm](https://user-images.githubusercontent.com/523210/45907246-7d350a80-bdf7-11e8-8735-c9d68e5a422c.gif)

peteruithoven:
I tried reproducing this on Loki in a VM, but there it works fine.

peteruithoven:
I've opened an issue in the Mutter issue queue: https://gitlab.gnome.org/GNOME/mutter/issues/343

peteruithoven:
Maybe someone could try this with the latest Mutter version? Ruling out it's an already fixed issue.

konstantinbo:
Happens to me too. I'd like to add another demonstration:

![window-tiling-glitch](https://user-images.githubusercontent.com/9221119/47151560-5e347600-d2da-11e8-880f-fc2de522ab5d.gif)

peteruithoven:
@konstantinbo which screencast tool did you use? That pressed key visualization is nice.

konstantinbo:
> @konstantinbo which screencast tool did you use? That pressed key visualization is nice.

It's a combination of Kazam (Screen record) and [Screenkey](https://github.com/wavexx/screenkey) to show the pressed keys. I used it in debug (-d) mode, because it wouldn't start normally.

peteruithoven:
Thanks to a tip on Slack it looks like the issue is actually caused by Gala. The tip was that it might be caused by it getting stuck in an animation.

I've figured out that by removing the following check I can change tile position directly. 
https://github.com/elementary/gala/blob/master/src/WindowManager.vala#L963

peteruithoven:
Now that I know a bit more, here is a easy way to reproduce it:
1. Open a fresh workspace 
2. Open an app like Files (not minimized or tiles)
3. Tile left by pressing super+ctrl+left
4. Then try tilling right by pressing super+ctrl+right
5. Weird things happen.

This seems to happen because in both tile locations the window has the same size and we depend on a size changed event to finish a animation. 
When the tile sides have a different size this is no issue.

Issue #305, opened 2018-07-20
Tiling an identically-sized window causes window to glitch


cassidyjames:
Originally reported by @ntilotta at https://github.com/elementary/os/issues/102

>elementary OS Juno 5.0 (Beta 1)
>
>If you open an electron-based app, this usually freezes after minimize/maximize, then for example Atom is transparent if you launch it maximized, and every application loses user' settings after you close and re-open it.

I've also noticed similar issues when tiling Google Chrome, both with and without the "system title bar and borders" option checked. I imagine it's the same bug since both are using Chromium and that cross-platform window code.

Djaler:
I noticed this with Insomnia. But this is not occure with Slack, for example

GammaGames:
I've had the tiling issues with both Chrome and VSCode

GammaGames:
I just found a slightly stranger issue, but I thought I'd report it here in case it is related. If it's not related it can be its own issue.

VSCode did the broken tiling thing again, so I opened the multitasking view so I could close the window. After I closed it it the window stayed on my desktop and won't go away. I can open and close more windows of VSCode but this phantom window remains.

![image](https://user-images.githubusercontent.com/7832163/47191417-702b0e80-d304-11e8-81c4-9cf0877578a6.png)

And, because I thought it was funny, here's Principles overlaying the phantom window:

![image](https://user-images.githubusercontent.com/7832163/47191520-31e21f00-d305-11e8-9574-3c192227f2f4.png)

GammaGames:
If you are able to find the edge of the window and resize it, the window will go to its correct position and correctly resize.

suprko:
I hope that this is helpful. I can get this to reproduce consistently:

1. Open Electron-based app that saves previous window width (e.g. Signal)
2. Tile the app left or right
3. Close the app
4. Reopen the app
5. Try to tile the app left or right => app wigs out

This can avoided by changing the app window size before tiling.

Because this specifically happens when the app window starts at the same size as it would be after tiling, I wonder if it's related to this other bug: [this comment](https://github.com/elementary/gala/issues/341#issuecomment-433566218) suggests that weird things are happening when there's an animation on a window whose size is exactly the same before and after an animation.

carn1x:
Getting this inconsistently but regularly with Thunderbird and Chrome.

btkostner:
@carn1x Terminal and AppCenter are not electron apps... That might be a different issue you are seeing, or the scope of this issue is different then what we thought.

[...]

PR #387, opened 2018-10-24
Tile switching fix

peteruithoven:
Looks like the size_change isn't called when switching directly between tile positions using the keyboard. Possibly because the size isn't changed. But position_changed is called. 
Fixes:  #341 
Possibly also fixes: #305

### DIFF
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -500,7 +500,7 @@ namespace Gala
 					|| window.minimized) {
 					return;
 				}
-	
+
 				var actor = window.get_compositor_private () as Clutter.Actor;
 				if (enable_animations) {
 					var op_trans = new Clutter.KeyframeTransition ("opacity");
@@ -511,7 +511,7 @@ namespace Gala
 					op_trans.set_to_value (255.0f);
 					op_trans.set_key_frames (op_keyframes);
 					op_trans.set_values (opacity);
-	
+
 					actor.add_transition ("opacity-hide", op_trans);
 				} else {
 					Timeout.add ((uint)(fade_out_duration * op_keyframes[0]), () => {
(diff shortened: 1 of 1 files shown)


peteruithoven:
Like requested I tried reusing the same logic in both signal handlers. I really don't like how this looks, but I couldn't figure out a better solution.

mtomov:
Thanks!

src/WindowManager.vala
```diff
@@ -963,28 +963,33 @@ namespace Gala
...
-						break;
-					case Meta.SizeChange.FULLSCREEN:
-					case Meta.SizeChange.UNFULLSCREEN:
-						handle_fullscreen_window (actor.get_meta_window (), which_change);
-						break;
-				}
-
-				size_change_completed (actor);
-			});
+			}
```
tintou: empty newline after closed bracket

src/WindowManager.vala
```diff
@@ -963,28 +963,33 @@ namespace Gala
...
+					maximize (actor, new_rect.x, new_rect.y, new_rect.width, new_rect.height);
+					break;
+				case Meta.SizeChange.UNMAXIMIZE:
+					unmaximize (actor, new_rect.x, new_rect.y, new_rect.width, new_rect.height);
+					break;
+				case Meta.SizeChange.FULLSCREEN:
+				case Meta.SizeChange.UNFULLSCREEN:
+					handle_fullscreen_window (actor.get_meta_window (), which_change);
+					break;
+			}
```
tintou: empty newline after closed bracket

src/WindowManager.vala
```diff
@@ -963,28 +963,33 @@ namespace Gala
...
-				size_change_completed (actor);
-			});
+			}
+			ulong size_signal_id = 0U;
+			ulong position_signal_id = 0U;
+			size_signal_id = window.size_changed.connect (() => window_change_complete (actor, which_change, size_signal_id, position_signal_id));
+			position_signal_id = window.position_changed.connect (() => window_change_complete (actor, which_change, size_signal_id, position_signal_id));
+		}
+		
+		private void window_change_complete ( Meta.WindowActor actor, Meta.SizeChange which_change, ulong size_signal_id, ulong position_signal_id) {
```
tintou:
```suggestion
		private void window_change_complete (Meta.WindowActor actor, Meta.SizeChange which_change, ulong size_signal_id, ulong position_signal_id) {
```

peteruithoven:
Thanks for the review @tintou. I've fixed those code style issues.

peteruithoven:
@tintou could you have a look at this again?

peteruithoven:
@ricotz If I understand things correctly your approach isn't yet possible with valac, would it be an idea to revert back to my workaround? I could open a seperate PR with your approach, but mark it as blocked. Maybe, if available link to an relevant valac issue?

PR #388, opened 2018-10-25
Add command to quickly install all dependencies

solarkraft:
This really reduces the time required for beginners to get up and running with Gala development.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,11 @@ You'll need the following dependencies:
 * libxml2-utils
 * valac (>= 0.28.0)
 
+Apt command to make sure all are installed on Elementary OS:
+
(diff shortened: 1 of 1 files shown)


arshubham:
Doesn't `sudo apt build-dep gala` achieve the same?

peteruithoven:
@arshubham as I've included in the follow tips, that only works when the dependencies match the last published version.
https://github.com/elementary/website/blob/ee96472e5f8483a59590ff7628a23fee94c64d0f/docs/code/os-dev.md

danrabbit:
I don't think we want to add distro specific commands like this to the README

ricotz:
I am with @danrabbit here, sorry.

Issue #105, opened 2017-10-08
Window Tiling / Snapping to quarters of screen


peteruithoven:

peteruithoven:
I'm pretty sure this logic is handled by Mutter, which Gala uses, so this would probably be out of scope for Gala itself? 
Recent update on this: https://gitlab.gnome.org/GNOME/mutter/issues/52#note_77121
The actual Mutter issue to track: https://bugzilla.gnome.org/show_bug.cgi?id=751857

peteruithoven:
I've implemented a quarter tiling Gala plugin that works with keys only at the moment:
https://github.com/elementary/gala/pull/408

1280px:
Is this issue any active now? Would be nice to see this feature in 7.1 or 8

nevillepark:
I also really miss this feature from other WMs and would love to see it in elementary.

RustoMCSpit:
Tiling top and bottom would be nice too!

PR #408, opened 2018-11-10
Basic quarter tiling

peteruithoven:
Basic quarter tiling support. 
<kbd>Ctrl</kbd> + <kbd>Super</kbd> + <kbd>1</kbd>: Tile top left
<kbd>Ctrl</kbd> + <kbd>Super</kbd> + <kbd>1</kbd>: Tile top left

Partially Implements: #105 (only the keyboard support)

### DIFF
--- a/plugins/quartertiler/Main.vala
+++ b/plugins/quartertiler/Main.vala
@@ -1,0 +1,86 @@
+//
+//  Copyright (C) 2018 Peter Uithoven
+//
+//  This program is free software: you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License as published by
+//  the Free Software Foundation, either version 3 of the License, or
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//
+
+using Clutter;
+using Meta;
+
+namespace Gala.Plugins.QuarterTiler
+{
+	public class Main : Gala.Plugin
+	{
+		Gala.WindowManager? wm = null;
+		Screen screen;
+
+		public override void initialize (Gala.WindowManager wm)
+		{
+			this.wm = wm;
+			screen = wm.get_screen ();
+			Display display = screen.get_display ();
+			var settings = new GLib.Settings (Config.SCHEMA + ".keybindings");
+			display.add_keybinding ("tile-topleft", settings, Meta.KeyBindingFlags.NONE, (Meta.KeyHandlerFunc) on_initiate);
+			display.add_keybinding ("tile-topright", settings, Meta.KeyBindingFlags.NONE, (Meta.KeyHandlerFunc) on_initiate);
+			display.add_keybinding ("tile-bottomleft", settings, Meta.KeyBindingFlags.NONE, (Meta.KeyHandlerFunc) on_initiate);
+			display.add_keybinding ("tile-bottomright", settings, Meta.KeyBindingFlags.NONE, (Meta.KeyHandlerFunc) on_initiate);
+		}
+
+		[CCode (instance_pos = -1)]
+		void on_initiate (Meta.Display display, Meta.Screen screen,
+			Meta.Window? window, Clutter.KeyEvent event, Meta.KeyBinding binding)
+		{
+			Window focused_window = display.get_focus_window ();
+			if (!focused_window.allows_move () || !focused_window.allows_resize ()){
+				return;
+			}
+			Meta.Rectangle wa = focused_window.get_work_area_current_monitor ();
+
+			int x = wa.x;
+			int y = wa.y;
+			var width = wa.width / 2;
+			var height = wa.height / 2;
+			switch (binding.get_name ()) {
+				case "tile-topright":
+					x += width;
+					break;
+				case "tile-bottomleft":
+					y += height;
+					break;
+				case "tile-bottomright":
+					x += width;
+					y += height;
+					break;
+			}
+			focused_window.move_resize_frame (true, x, y, width, height);
+		}
+
+		public override void destroy ()
+		{
+
+		}
+	}
+}
+
+public Gala.PluginInfo register_plugin ()
+{
+	return
+	{
+		"Quarter tiler",
+		"Peter Uithoven <peter@peteruithoven.com>",
+		typeof (Gala.Plugins.QuarterTiler.Main),
+		Gala.PluginFunction.ADDITION,
+		Gala.LoadPriority.IMMEDIATE
+	};
+}
--- a/data/org.pantheon.desktop.gala.gschema.xml.in
+++ b/data/org.pantheon.desktop.gala.gschema.xml.in
@@ -15,7 +15,7 @@
 		<value nick='grid' value='0'/>
 		<value nick='natural' value='1'/>
 	</enum>
-	
+
 	<schema path="/org/pantheon/desktop/gala/behavior/" id="org.pantheon.desktop.gala.behavior" gettext-domain="@GETTEXT_PACKAGE@">
 		<key enum="GalaActionType" name="hotcorner-topleft">
 			<default>'none'</default>
@@ -83,7 +83,7 @@
 			<description>When launching the window-switcher, Gala iterates over this list and attempts to find a window matching the names. If it does, it will hide this window and fade it back in, once the the switcher is closed.</description>
 		</key>
 	</schema>
-	
+
 	<schema path="/org/pantheon/desktop/gala/keybindings/" id="org.pantheon.desktop.gala.keybindings" gettext-domain="@GETTEXT_PACKAGE@">
 		<key type="as" name="switch-to-workspace-first">
 			<default><![CDATA[['<Super>Home']]]></default>
@@ -147,8 +147,28 @@
 			<summary>The shortcut to enable picture-in-picture window</summary>
 			<description>The shortcut to show the selection area to choose a window.</description>
 		</key>
+		<key type="as" name="tile-topleft">
+			<default><![CDATA[['<Super><Ctrl>1']]]></default>
+			<summary>Tile window to top left</summary>
+			<description>The shortcut to quarter tile window to top left.</description>
+		</key>
+		<key type="as" name="tile-topright">
+			<default><![CDATA[['<Super><Ctrl>2']]]></default>
+			<summary>Tile window to top right</summary>
+			<description>The shortcut to quarter tile window to top right.</description>
+		</key>
+		<key type="as" name="tile-bottomleft">
+			<default><![CDATA[['<Super><Ctrl>3']]]></default>
+			<summary>Tile window to bottom left</summary>
+			<description>The shortcut to quarter tile window to bottom left.</description>
+		</key>
+		<key type="as" name="tile-bottomright">
+			<default><![CDATA[['<Super><Ctrl>4']]]></default>
+			<summary>Tile window to bottom right</summary>
+			<description>The shortcut to quarter tile window to bottom right.</description>
+		</key>
 	</schema>
-	
+
 	<schema path="/org/pantheon/desktop/gala/appearance/" id="org.pantheon.desktop.gala.appearance" gettext-domain="@GETTEXT_PACKAGE@">
 		<key type="b" name="attach-modal-dialogs">
 			<default>false</default>
@@ -175,7 +195,7 @@
 			<summary>Background-image used in the workspace-switcher</summary>
 		</key>
 	</schema>
-	
+
 	<schema path="/org/pantheon/desktop/gala/animations/" id="org.pantheon.desktop.gala.animations" gettext-domain="@GETTEXT_PACKAGE@">
 		<key type="b" name="enable-animations">
 			<default>true</default>
@@ -206,7 +226,7 @@
 			<summary>Duration of the menu mapping animation</summary>
 		</key>
 	</schema>
-	
+
 	<schema path="/org/pantheon/desktop/gala/shadows/" id="org.pantheon.desktop.gala.shadows" gettext-domain="@GETTEXT_PACKAGE@">
 		<key type="as" name="normal-focused">
 			<default>['20', '-1', '0', '15', '220']</default>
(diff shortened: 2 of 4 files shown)


cassidyjames:
I think we'd want these to be bound to arrow keys by default if possible, so:

Ctrl+Super+Up+Left, Ctrl+Super+Up+Right, etc.

jchannon:
I’d prefer these but not sure it fixes your issue. 

Move to the upper left — ⌃⌘←
Move to the lower left — ⌃⇧⌘←
Move to the upper right — ⌃⌘→
Move to the lower right — ⌃⇧⌘→

peteruithoven:
@jchannon I might be confused by your usage of ⌃ (which means ctrl right?), but isn't ⌃⌘←( <kbd>Ctrl</kbd>+<kbd>Super</kbd>+<kbd>←</kbd>) the same as the current tile left keyboard shortcut?

jchannon:
Quite possibly, those are the shortcuts I use on OSX for quarter tiling :)

ianmjones:
This is awesome, but I worry that changing the key combinations to use two cursor keys at the same time would be problematic given that using one of those keys is already ingrained and as @peteruithoven has found requires some special handling of the keybindings.

I would propose something a little more visual and hopefully simpler to implement.

<kbd>Ctrl</kbd>+<kbd>Super</kbd>+<kbd>u</kbd>: Top Left
<kbd>Ctrl</kbd>+<kbd>Super</kbd>+<kbd>i</kbd>: Top Right
<kbd>Ctrl</kbd>+<kbd>Super</kbd>+<kbd>j</kbd>: Bottom Left
<kbd>Ctrl</kbd>+<kbd>Super</kbd>+<kbd>k</kbd>: Bottom Right

This could later be suplemented with 3rd tiling...

<kbd>Ctrl</kbd>+<kbd>Super</kbd>+<kbd>d</kbd>: Left 3rd
<kbd>Ctrl</kbd>+<kbd>Super</kbd>+<kbd>f</kbd>: Middle 3rd
<kbd>Ctrl</kbd>+<kbd>Super</kbd>+<kbd>g</kbd>: Right 3rd

<kbd>Ctrl</kbd>+<kbd>Super</kbd>+<kbd>e</kbd>: Left 2 3rds
<kbd>Ctrl</kbd>+<kbd>Super</kbd>+<kbd>t</kbd>: Right 3rds

I might be influenced by muscle memory, but I do honestly think it's less problematic.

magnet for macos keybindings

peteruithoven:
One issue with using letters is that it's logical orientation breaks down on non Querty keyboards. I couldn't find statistics on keyboard layout popularity.

ianmjones:
Hmm, yeah, that did occur to me.

I wonder if there is some sort of hardware code mapping that could be used more reliably than letters for saving to settings?

jchannon:
Any update to this, most distros have quarter tiling so it would be good to get it into elementary

plugins/quartertiler/Main.vala
```diff
@@ -15,22 +15,15 @@
...
 		{
-			Screen screen = wm.get_screen ();
-			Display display = screen.get_display ();
+			unowned Meta.Display display = wm.get_screen ().get_display ();
+
 			var settings = new GLib.Settings (Config.SCHEMA + ".keybindings");
-			if (!("tile-topleft" in settings.list_keys ())) {
-					warning ("Quarter tiling key bindings not found");
-					return;
-			}
```
peteruithoven: @ricotz Thanks for the cleanup! But why remove this check? Without this, if a user where to try gala with this version and then downgrade he wouldn't be able to login anymore. 
See: https://elementaryos.stackexchange.com/a/17582/3892

ricotz: gsettings schemas and keys are part of the API. We can't deal with not properly installed systems of all kinds.

peteruithoven: After I've experiencing suddenly not being able to login that seems scary / fragile, but yeah it probably doesn't make sense to make an exception for this property.

thomasfaller:
Would be a terrific addition!

Julioevm:
I hope this get added soon, its really useful productivity wise!

peteruithoven:
@donadigo nop it definitely wasn't. I just rarely use maximized / fullscreen windows. I added those allow resize / move checks because that seemed safe. But I changed it, it now also works for maximized and fullscreen windows.

cassidyjames:
I want to get feedback from @elementary/ux about the shortcuts before we merge

btkostner:
Is it possible to tell if a window is already half tiled? If so, maybe have it so you half tile first, then do up or down to go into quarter tiling?

peteruithoven:
@btkostner yes I can tell if it's tiled. 
What would that sequence of actions accomplish?

btkostner:
@peteruithoven So <kbd>super</kbd>+<kbd>ctrl</kbd>+<kbd>left</kbd> would left half tile it, then <kbd>super</kbd>+<kbd>ctrl</kbd>+<kbd>top</kbd> would quarter tile it to the upper left. Just a thought :thinking:

peteruithoven:
@btkostner Ah! Interesting idea! I've implemented the idea. Please let me know what you think.

jchannon:
So you’d have to do two sets of shortcuts to get it to tile to top left for
example? Seems unproductive

btkostner:
@jchannon 	The way I've seen it implemented is where you _could_ (but don't have to) let go of <kbd>left</kbd> or <kbd>right</kbd>. But this way, you get instant feedback.

btkostner:
@peteruithoven Just tried it and it feels pretty cool :confetti_ball:. Two things might need to be changed, but I think @cassidyjames might have some insight to this. First being <kbd>up</kbd> + <kbd>left</kbd> should be the same thing as <kbd>left</kbd> + <kbd>up</kbd> if possible (not sure the technical constraints)... Second, I found myself wanting to move the whole window when it was already tiled. For instance, if I had a window in the upper left part of the screen and I did <kbd>right</kbd>, it would then be tiled in the upper right of the screen. This might be my tiling window manager brain, but might be something to try out...

peteruithoven:
> First being <kbd>up</kbd> + <kbd>left</kbd> should be the same thing as <kbd>left</kbd> + <kbd>up</kbd> if possible (not sure the technical constraints)... 

I think I've explained the technical limitation earlier. Properly implementing that keyboard shortcut simply is not supported. With the left/light > top/down I can use the tiling position as a workaround. 

I thought I could maybe also register to <kbd>Super</kbd>+<kbd>Ctrl</kbd>+<kbd>left</kbd>/<kbd>right</kbd> and maybe do some kind of timeout where if <kbd>Super</kbd>+<kbd>Ctrl</kbd>+<kbd>up</kbd>/<kbd>down</kbd></kbd> and <kbd>Super</kbd>+<kbd>Ctrl</kbd>+<kbd>left</kbd>/<kbd>right</kbd> are pressed within a certain time I could trigger a quarter tile, but listening for <kbd>left</kbd>/<kbd>right</kbd> breaks the existing tile to side behaivior. I assume only one listener is allowed.

peteruithoven:
> Second, I found myself wanting to move the whole window when it was already tiled. For instance, if I had a window in the upper left part of the screen and I did right, it would then be tiled in the upper right of the screen. This might be my tiling window manager brain, but might be something to try out...

Let's first get quarter tiling to work...

plugins/quartertiler/Main.vala
```diff
@@ -0,0 +1,112 @@
...
+					break;
+				case "tile-bottomleft":
+					y += height;
+					break;
+				case "tile-bottomright":
+					x += width;
+					y += height;
+					break;
+				case "tile-top":
+					if (!was_maximized_vertically) {
```
jlnr: This `if` logic means a window in the bottom-left corner cannot be moved to the top-left corner by pressing ctrl+meta+up. I think this is a bit of a missed opportunity.

jlnr: Conversely, maximized windows do react to ctrl+meta+down, and move to the bottom left corner. Maybe it would be more consistent if they occupied the bottom half of the screen? (And vice versa for up)

peteruithoven: But doing that opens up a lot of edge cases, for example for determining on which side the window is. Since I'm limiting myself to already tiled windows it's left when x is 0, it's on the right when it's not.

plugins/quartertiler/Main.vala
```diff
@@ -0,0 +1,112 @@
...
+						return;
+					}
+
+					if (prev_rect.x != 0) { // right side
```
jlnr: I think checking for `prev_rect.x == wa.x` would be safer, because the primary workarea does not always start at 0/0.

plugins/quartertiler/Main.vala
jlnr: `maximized_vertically` is true for tiled windows *and* maximized windows, which is why maximized windows can be shrunk into the bottom left corner using ctrl+shift+down right now.

I feel either the up/down shortcuts should be ignored in maximized windows, or they should do something that does not randomly favor the left over the right side (e.g. by using the full bottom half of the screen in this special case).

jlnr: The code would still be simpler that way: If the cases for "tile-top" and "tile-bottom" would leave the `x` and `width` of the window alone, and only adjust `y` and `height`, then ctrl+shift+up could be used to move from "left half" to "top left corner", etc.

plugins/quartertiler/Main.vala
peteruithoven: Good point! For example when putting something like the plank on the left side, like Ubuntu.

plugins/quartertiler/Main.vala
peteruithoven: Good points, except for the fact that when I "unmaximize" a window it's position and size will be restored to a pre-tiled state. If I don't unmaximize I can't change the position and size. 

> I feel either the up/down shortcuts should be ignored in maximized windows, or they should do something that does not randomly favor the left over the right side (e.g. by using the full bottom half of the screen in this special case).

Good point! I'm now tiling to the full bottom or top of the screen when the window was maximized.

peteruithoven:
So a little update, beside tiling using <kbd>Super</kbd>+<kbd>Ctrl</kbd>+<kbd>1</kbd>/<kbd>2</kbd>/<kbd>3</kbd>/<kbd>4</kbd> I've implemented @btkostner's suggestion. You can now quarter tile using a two step process:
1. Tile window to left or right using <kbd>Super</kbd>+<kbd>Ctrl</kbd>+<kbd>Left</kbd>/<kbd>Right</kbd>
2. Quarter tile window to top or bottom of that side using  <kbd>Super</kbd>+<kbd>Ctrl</kbd>+<kbd>Up</kbd>/<kbd>Down</kbd>

Tiling to top or bottom half is also supported:
1. Maximize window using <kbd>Super</kbd>+<kbd>Up</kbd>
2. Tile to top or bottom half using: <kbd>Super</kbd>+<kbd>Ctrl</kbd>+<kbd>Up</kbd>/<kbd>Down</kbd>

And thanks @jlnr for the code review.

jlnr:
Can't wait to try the updated branch 💪

Another thought: Does a majority of people use keyboards with numpads? (Definitely not me, I'm a complete TKL fanatic 😬) Would 1, 3, 6, 9 be more intuitive shortcuts for corner tiling in that case? You could also use 4 and 6 for half-screen tiling, etc...

btkostner:
@jlnr 	@cassidyjames 	That's a pretty cool idea. I would wonder how many people use a 10 key with elementary. I personally do, and this solution sounds like the best for me, but I'm not sure that would work for everyone...

peteruithoven:
I've personally never had those, on any of the laptops I've owned. But sounds like someone could just remap the <kbd>Super</kbd>+<kbd>Ctrl</kbd>+<kbd>1</kbd>/<kbd>2</kbd>/<kbd>3</kbd>/<kbd>4</kbd> and <kbd>Super</kbd>+<kbd>Ctrl</kbd>+<kbd>Left</kbd> shortcuts? 
You might miss something for the <kbd>2</kbd> and <kbd>8</kbd> though, something to directly tile something to the top or bottom half of the screen.

plugins/quartertiler/Main.vala
```diff
@@ -0,0 +1,108 @@
...
+			if (focused_window.fullscreen) {
+				focused_window.unmake_fullscreen ();
+			}
+
+			Meta.Rectangle wa = focused_window.get_work_area_current_monitor ();
+			int x = wa.x;
+			int y = wa.y;
+			int width = wa.width / 2;
+			int height = wa.height / 2;
+			bool leftSide = prev_rect.x == wa.x;
```
jlnr: This variable is unused and causes a warning.

peteruithoven: Good catch, thanks!

renevall:
I hope this makes it. I'm actually using a UHD 3440x1440 and three panels side by side would be awesome too. But I yes, hope all the work on this ticket makes it

jlnr:
@btkostner About the numpad idea: Turns out Ubuntu MATE 19.10 just changed it's tiling shortcuts *away* from a numpad-based system ([see "New Key-bindings"](https://ubuntu-mate.org/blog/ubuntu-mate-19-10-eoan-ermine-release/)). I don't know what data they've based it on, but it seems numpads are not an overwhelming majority then.

jchannon:
How is this still not merged 😄 Looking at the issue list there is 10 issues about tiling. Do elementary not think tiling is important?

jlnr:
Because it's important to get it right - that's why there are several issues and even alternative PRs (#595). The underlying APIs are also a bit limited (keyboard shortcuts, mutter tiling). The best way to accelerate this is to provide a PR that has absolutely no glitches 😜

gazsp:
Please get this added! 

It's one of the main things I miss from macOS / SizeUp - not being able to fling windows around with super + numpad (quartered / halved / centred)

Just make the keys re-definable in Keyboard Settings to keep everyone happy.

jchannon:
Anyone tried this? https://www.linuxuprising.com/2020/02/wintile-for-gnome-shell-windows-10-like.html

jeremypw:
@peteruithoven Converting to draft as there are conflicts and unresolved design issues, however there seems to be a lot of interest in it.
@danrabbit Is it just choice of shortcuts blocking this?  Can the design issues be resolved or is there a fundamental conflict?

CodingSpiderFox:
For me, the initially proposed key combinations would make the most sense:

Ctrl + Super + 1: Tile top left
Ctrl + Super + 2: Tile top right
Ctrl + Super + 3: Tile bottom left
Ctrl + Super + 4: Tile bottom right

Are there any plans to implement such a feature?
I've seen multiple PRs in this repo which also provide very cool methods for organizing windows withing a single display.

Would be nice of one of these would be discussed with the design team and eventually make it into a future release of gala.

So far, these approaches seem the most promising:

https://github.com/elementary/gala/pull/595
https://github.com/elementary/gala/pull/890

PR #438, opened 2019-01-03
Change PIP Default Location to Bottom Right

liamkinne:
Most apps will render content top to bottom, left to right. Putting the popup window on the bottom right by default (instead of left) makes sense to have it less likely to cover up content.

screenshot from 2019-01-04 02-17-18

### DIFF
--- a/plugins/pip/PopupWindow.vala
+++ b/plugins/pip/PopupWindow.vala
@@ -128,7 +128,7 @@ public class Gala.Plugins.PIP.PopupWindow : Clutter.Actor
 		var x_position = monitor_rect.width + monitor_rect.x - SCREEN_MARGIN - width;
 		var y_position = monitor_rect.height + monitor_rect.y - SCREEN_MARGIN - height;
 
-		if (Clutter.get_default_text_direction() == Clutter.TextDirection.RTL) {
+		if (Clutter.get_default_text_direction () == Clutter.TextDirection.RTL) {
 			x_position = SCREEN_MARGIN + monitor_rect.x;
 		}
 
--- a/plugins/pip/PopupWindow.vala
+++ b/plugins/pip/PopupWindow.vala
@@ -132,7 +132,7 @@ public class Gala.Plugins.PIP.PopupWindow : Clutter.Actor
 			x_position = SCREEN_MARGIN + monitor_rect.x;
 		}
 
-		set_position(x_position, y_position);
+		set_position (x_position, y_position);
 
 		close_action = new Clutter.ClickAction ();
 		close_action.clicked.connect (on_close_click_clicked);
--- a/plugins/pip/PopupWindow.vala
+++ b/plugins/pip/PopupWindow.vala
@@ -125,7 +125,7 @@ public class Gala.Plugins.PIP.PopupWindow : Clutter.Actor
 		Meta.Rectangle monitor_rect;
 		get_current_monitor_rect (out monitor_rect);
 
-		set_position (SCREEN_MARGIN + monitor_rect.x, monitor_rect.height + monitor_rect.y - SCREEN_MARGIN - height);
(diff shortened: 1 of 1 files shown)


btkostner:
@cassidyjames would this make sense to be based on language direction. Right side for English, and left side for RTL languages?

cassidyjames:
@btkostner ah yeah that's smart. I like it all; putting the window at the "end" of the reading/skimming flow as well as being aware of RTL. I know bottom-right is the default in LTR for Android at least and it always made sense to me.

liamkinne:
Added RTL awareness and tested with Arabic.

plugins/pip/PopupWindow.vala
```diff
@@ -125,7 +125,14 @@ public class Gala.Plugins.PIP.PopupWindow : Clutter.Actor
 		Meta.Rectangle monitor_rect;
 		get_current_monitor_rect (out monitor_rect);
 
-		set_position (SCREEN_MARGIN + monitor_rect.x, monitor_rect.height + monitor_rect.y - SCREEN_MARGIN - height);
+		var x_position = monitor_rect.width + monitor_rect.x - SCREEN_MARGIN - width;
+		var y_position = monitor_rect.height + monitor_rect.y - SCREEN_MARGIN - height;
+
+		if (Clutter.get_default_text_direction() == Clutter.TextDirection.RTL) {
```
cassidyjames:
```suggestion
		if (Clutter.get_default_text_direction () == Clutter.TextDirection.RTL) {
```

plugins/pip/PopupWindow.vala
```diff
@@ -125,7 +125,14 @@ public class Gala.Plugins.PIP.PopupWindow : Clutter.Actor
...
+			x_position = SCREEN_MARGIN + monitor_rect.x;
+		}
+
+		set_position(x_position, y_position);
```
cassidyjames:
```suggestion
		set_position (x_position, y_position);
```

plugins/pip/PopupWindow.vala
```diff
@@ -125,7 +125,14 @@ public class Gala.Plugins.PIP.PopupWindow : Clutter.Actor
 		Meta.Rectangle monitor_rect;
 		get_current_monitor_rect (out monitor_rect);
 
-		set_position (SCREEN_MARGIN + monitor_rect.x, monitor_rect.height + monitor_rect.y - SCREEN_MARGIN - height);
+		var x_position = monitor_rect.width + monitor_rect.x - SCREEN_MARGIN - width;
+		var y_position = monitor_rect.height + monitor_rect.y - SCREEN_MARGIN - height;
+
+		if (Clutter.get_default_text_direction () == Clutter.TextDirection.RTL) {
+			x_position = SCREEN_MARGIN + monitor_rect.x;
```
Philip-Scott: Its really a nitpick, so I'm not going to "block" the PR for this, but instead of posibbly setting `x_position` twice, can you do it all inside an if-else statement? :)  You can always declare the variable outside of the if-else block

liamkinne: Yeah I agree. I originally wrote what you're suggesting, but went with this cause you can easily remove the RTL logic without effecting it. But, putting the default in an else works. Will do.

plugins/pip/PopupWindow.vala
```diff
@@ -125,7 +125,17 @@ public class Gala.Plugins.PIP.PopupWindow : Clutter.Actor
...
 		get_current_monitor_rect (out monitor_rect);
 
-		set_position (SCREEN_MARGIN + monitor_rect.x, monitor_rect.height + monitor_rect.y - SCREEN_MARGIN - height);
+		var x_position = 0.0f;
+		var y_position = monitor_rect.height + monitor_rect.y - SCREEN_MARGIN - height;
+
+		if (Clutter.get_default_text_direction () == Clutter.TextDirection.RTL) {
+			x_position = SCREEN_MARGIN + monitor_rect.x;
+		}
+		else {
```
ricotz: No line break here: `} else {`

PR #460, opened 2019-02-06
Do not block drawing the stage when saving the image

donadigo:
This fixes a large lag when the image is being saved into a file when taking a screenshot. Mostly noticeable on HDD's. This changes all exposed method signatures to async but doesn't affect callers in any way.


peteruithoven:
Works great! Also makes a difference on SSD's with HiDPI screens when taking a whole screen screenshot.

Issue #311, opened 2018-08-01
Shift+PrtSc takes screenshot even if we press Esc key while we're selecting area


ryonakano:
## Expected behavior

If we press `Esc` while we're selecting area, Shift+PrtSc should cancel taking a screenshot.

## Actual behavior

`Shift+PrtSc` takes screenshot even if we press `Esc` while we're selecting area.

## Steps to reproduce the behavior

1. Press `Shift+PrtSc` and press `Esc`
2. The action cancels and no screenshot is taken
3. Press `Shift+PrtSc`, drag to select area to grub, and press `Esc` without releasing your finger from touchpad/mouse
4. A screenshot is taken although we press `Esc` to cancel taking the screenshot

peteruithoven:
When creating a custom keybinding to `io.elementary.screenshot-tool -r` cancel using <kbd>esc</kbd> does work as expected.

danrabbit:
Closing because this seems to be fixed

PR #463, opened 2019-02-06
Cancel area screenshot when pressing esc

peteruithoven:
Fixing: #311 
Using 2 signals to differentiate between captured and cancelled just like the [SelectionArea in Screenshot](https://github.com/elementary/screenshot/blob/master/src/Widgets/SelectionArea.vala).

### DIFF
--- a/src/ScreenshotManager.vala
+++ b/src/ScreenshotManager.vala
@@ -86,6 +86,11 @@ namespace Gala
 		public async void screenshot_area (int x, int y, int width, int height, bool flash, string filename, out bool success, out string filename_used) throws DBusError, IOError
 		{
 			debug ("Taking area screenshot");
+			if (width == -1 && height == -1) {
+				success = false;
+				filename_used = "";
+				return;
+			}
 
 			yield wait_stage_repaint ();
 
--- a/src/ScreenshotManager.vala
+++ b/src/ScreenshotManager.vala
@@ -132,7 +132,8 @@ namespace Gala
 		public async void select_area (out int x, out int y, out int width, out int height) throws DBusError, IOError
 		{
 			var selection_area = new SelectionArea (wm);
-			selection_area.closed.connect (() => Idle.add (select_area.callback));
+			selection_area.captured.connect (() => Idle.add (select_area.callback));
+			selection_area.cancelled.connect (() => selection_area.destroy ());
 			wm.ui_group.add (selection_area);
 			selection_area.start_selection ();
 
--- a/src/ScreenshotManager.vala
+++ b/src/ScreenshotManager.vala
@@ -136,25 +136,21 @@ namespace Gala
 
 		public async void select_area (out int x, out int y, out int width, out int height) throws DBusError, IOError
 		{
-			bool cancelled = false;
 			var selection_area = new SelectionArea (wm);
-			selection_area.captured.connect (() => Idle.add (select_area.callback));
-			selection_area.cancelled.connect (() => {
-				cancelled = true;
-				Idle.add (select_area.callback);
-			});
+			selection_area.closed.connect (() => Idle.add (select_area.callback));
 			wm.ui_group.add (selection_area);
 			selection_area.start_selection ();
 
 			yield;
 			selection_area.destroy ();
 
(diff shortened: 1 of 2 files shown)


ricotz:
Have you try using an error to propagate the cancellation rather than setting and checking "invalid" values?

peteruithoven:
I haven't, but I'd be basically guessing, I've been trying to find some api documentation of this `/org/gnome/Shell/Screenshot` / `org.gnome.Shell.Screenshot` interface.

This seems relevant: https://gitlab.gnome.org/GNOME/gnome-shell/blob/master/data/dbus-interfaces/org.gnome.Shell.Screenshot.xml

This looks like the gnome shell implementation, where they indeed seem to use an IOError:
https://gitlab.gnome.org/GNOME/gnome-shell/blob/master/js/ui/screenshot.js#L169-183

This looks like the actual implementation of the screenshot part of the media keys plugin of the GNOME settings daemon:
https://gitlab.gnome.org/GNOME/gnome-settings-daemon/blob/master/plugins/media-keys/gsd-screenshot-utils.c

It has this `if (!geometry)` check, but I'm not sure how it gets from the individual x, y, width, height to this one geometry value. 
https://gitlab.gnome.org/GNOME/gnome-settings-daemon/blob/master/plugins/media-keys/gsd-screenshot-utils.c#L223-228

tintou:
@peteruithoven If you look at https://gitlab.gnome.org/GNOME/gnome-shell/blob/master/data/dbus-interfaces/org.gnome.Shell.Screenshot.xml#L137 the SelectArea method has 4 out arguments, which is just represented by a GLib.Variant tuple: `(iiii)` → means `tuple-of(integer, integer, integer, integer)` so you can then get the result with `variant.get ("(iiii)", out x, out y, out width, out height)`

tintou:
Vala allow you to do it even easier:
```vala
public struct ScreenshotGeometry {
    public int x;
    public int y;
    public int width;
    public int height;
}
…
// DBus method:
public async ScreenshotGeometry select_area () throws GLib.Error;
```

peteruithoven:
@tintou Interesting, so that should be automatically possible because of how vala connects to these dbus interfaces? That has the potential to be more readable, I've tried the following (Ideally SelectionArea would have a `ScreenshotGeometry` property):
``` vala
public struct ScreenshotGeometry {
    public int x;
    public int y;
    public int width;
    public int height;
}

public async ScreenshotGeometry select_area () throws DBusError, IOError
{
  var selection_area = new SelectionArea (wm);
  selection_area.closed.connect (() => Idle.add (select_area.callback));
  wm.ui_group.add (selection_area);
  selection_area.start_selection ();

  yield;
  selection_area.destroy ();

  yield wait_stage_repaint ();
  int x;
  int y;
  int width;
  int height;
  debug ("selection_rectangle: %ix%i %ix%i", x, y, width, height);
  return ScreenshotGeometry() {
    x = x,
    y = y,
    width = width,
    height = height
  }
}
```
But, while the logged `selection_rectangle` seemed valid it's not picked up properly:
```
* (gala:19137): DEBUG: 13:12:56.032: ScreenshotManager.vala:155: selection_rectangle: 2313x620 1393x906
** (gala:19137): DEBUG: 13:12:56.034: ScreenshotManager.vala:88: Taking area screenshot
*** BUG ***
In pixman_region32_init_rect: Invalid rectangle passed
Set a breakpoint on '_pixman_log_error' to debug

** (gala:19137): DEBUG: 13:12:56.053: ScreenshotManager.vala:43: Flashing area

(gala:19137): Gdk-CRITICAL **: 13:12:56.054: gdk_pixbuf_get_from_surface: assertion 'width > 0 && height > 0' failed

(gala:19137): GdkPixbuf-CRITICAL **: 13:12:56.054: gdk_pixbuf_save_to_stream_async: assertion 'GDK_IS_PIXBUF (pixbuf)' failed
```

ricotz:
If an error is set then nothing is assigned and NULL is returned. So the code is fine as it is now.

Whether a restructuring by using a struct for optimizations makes sense is out of scope of that merge.

PR #464, opened 2019-02-08
Add ability to reorder workspaces

### LINKED ISSUE #259: Can't figure out how to reorder workspaces
I'm not sure how to reorder my existing workspaces.

It's easy enough to work around by creating a new workspace and manually moving each application to it one at a time, but this can be tedious when both workspaces have several applications open on them.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/59130197-can-t-figure-out-how-to-reorder-workspaces?utm_campaign=plugin&utm_content=tracker%2F62872462&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F62872462&utm_medium=issues&utm_source=github).
</bountysource-plugin>

donadigo:
Fixes #259.
~~Depends on #566 and https://gitlab.gnome.org/GNOME/mutter/merge_requests/670.~~
Depends on a distro-patch for mutter (https://gist.github.com/donadigo/fb99b5a9e8a092c76c4d6f03cc7ef930) that is already in the repos.

This adds the ability to drag a workspace and reorder it to a different place. This required some changes in core components to workaround the absence of a native method in libmutter to do this.

Some changes in the codebase:
* DragDropAction now can be initialized with both source and destination flags and supports those. This was done to have `IconGroup` as a source (DND to different place) and as a destination (DND a window to an icon group).

* `remove_group_in_place` is used to remove an `IconGroup` without having the `IconGroupContainer` reallocate and change it's position.

* The delay of `WorkspaceInsertThumb` is now dynamic, this is because dragging an `IconGroup` has a delay of 0 (the space is immediately made for the widget) and the standard (when dragging a window icon to make a new workspace)

* ~~Methods like the new `move_workspace_to_index` require freezing all animations in the multitasking view to stop having unnecessary animations when reordering workspaces. `MultitaskingView` now has static methods to change animation state.
We need to wait until https://gitlab.gnome.org/GNOME/mutter/merge_requests/670 hits our repos, then, none of the previous solutions will be required.~~ This was merged as well.


donadigo:
This is now ready to review.

cassidyjames:
@donadigo from a brief UX review, this looks really good and works as I'd expect! I'll continue using it and would like to hear from some people more familiar with the codebase, but I'm :+1:

Blast-City:
This is such a great feature.

Will it have a keyboard shortcut? Ctrl+Super+Alt+Left/Right?

peteruithoven:
One first tiny note. The workspace box, with the icons, what you drag seem completely transparent, which looks a bit weird when dragging them over other workspaces:
Screenshot from 2019-07-02 02 08 04@2x
Maybe the background of those boxes should be semi-transparent or opaque?

donadigo:
@peteruithoven fixed.

danrabbit:
It seems like there a couple of unnecessary animations, but otherwise this is a great functional improvement.

Open a few workspaces with one app on each of them. Open Multitasking view. :
- [ ] Grab the current workspace and drag it ever so slightly to the left or right (like 6px). The workspace to your left will jump to the position of your current workspace and then quickly animate back to its original position. It probably shouldn't move until you drag past it.
- [ ] Switch the order of any two workspaces. You'll notice that when you drop an app, it appears to be in a folder with the other app momentarily before it animates back to full size with a big "pop"

donadigo:
@danrabbit I've tried to fix 2. but it's really hard to do this because *yet again* mutter doesn't let us reorder workspace indexes so what we do instead is move all apps one by one to the next workspace so that we can make room for the new workspace. Moving those apps causes signals to fire and gala reanimates the apps that changed. It would be quite hacky in gala to prevent that.

peteruithoven:
I noticed I can easily lose the icons and I can't get them back unless I restart Gala. 
Screen record from 2019-07-02 19 40 07

I ran gala with `G_MESSAGES_DEBUG=all` but seen no warnings or errors.

danrabbit:
@donadigo ah damn. Would it be awful to wait to animate for like 100ms or something really tiny? Long enough for the move to occur but short enough that people won't really notice?

donadigo:
@danrabbit I will try to figure this out, you can see that there are still major problems found.

src/Widgets/IconGroupContainer.vala
```diff
@@ -65,6 +65,27 @@ namespace Gala
...
 		}
 
+		public void remove_group_in_place (IconGroup group)
+		{
+			var thumb = (WorkspaceInsertThumb) group.get_previous_sibling ();
+			float thumb_width = thumb.get_width ();
+			float group_width = group.get_width ();
+			remove_group (group);
+			
+			x += thumb_width + group_width;
```
peteruithoven: Why does the x position of the container need to be moved when dragging begins?

src/Widgets/MultitaskingView.vala
```diff
@@ -46,6 +46,24 @@ namespace Gala
...
+		{
+			animation_freeze_count++;
+		}
+
+		public static void thaw_animations ()
+		{
+			animation_freeze_count--;
+
+			assert (animation_freeze_count >= 0);
+		}
```
peteruithoven: Why is this counting system necessary? Shouldn't there only be one drag end handler active at the same time?

src/DragDropAction.vala
```diff
@@ -208,7 +230,13 @@ namespace Gala
 
 							var source_list = sources.@get (drag_id);
 							if (source_list != null) {
+								var dest_list = destinations[drag_id];
 								foreach (var actor in source_list) {
+									// Do not unset reactivity on destinations
```
peteruithoven: "reactivity"?

src/DragDropAction.vala
```diff
@@ -139,17 +146,22 @@ namespace Gala
...
-			if (drag_type == DragDropActionType.SOURCE) {
+			if (DragDropActionType.SOURCE in drag_type) {
 				actor.button_press_event.disconnect (source_clicked);
 
 				var source_list = sources.@get (drag_id);
 				source_list.remove (actor);
+			} 
+			
+			if (DragDropActionType.DESTINATION in drag_type) {
+				var dest_list = destinations[drag_id];
```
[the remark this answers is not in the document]
donadigo: Gee supports native Vala syntax and that's why I used it here. I didn't touch any of the code that wasn't needed to be. It is no different than calling `@get`.

src/Widgets/WorkspaceInsertThumb.vala
```diff
@@ -43,8 +44,8 @@ namespace Gala
 			x_align = Clutter.ActorAlign.CENTER;
 
 			var drop = new DragDropAction (DragDropActionType.DESTINATION, "multitaskingview-window");
-			drop.crossed.connect ((hovered) => {
-				if (!Prefs.get_dynamic_workspaces ())
+			drop.crossed.connect ((target, hovered) => {
+				if (!Prefs.get_dynamic_workspaces () && (target != null && target is WindowClone))
```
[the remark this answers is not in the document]
donadigo: This behaviour is handled directly in `WindowClone`: https://github.com/elementary/gala/blob/ca32785fe4ffc64eaa01d9c87cbcb6b4336bca6a/src/Widgets/WindowClone.vala#L605

So instead of handling it here, `WindowClone` watches for crossed destinations and calls `set_window_thumb` which resizes the `WorkspaceInsertThumb` and adds a pulse animation.

src/Widgets/IconGroup.vala
```diff
@@ -365,7 +367,13 @@ namespace Gala
 				5 * scale
 			);
 
-			cr.set_source_rgba (0, 0, 0, 0.1);
+			if (drag_action.dragging) {
+				cr.set_source_rgba (0, 0, 0, 0.1);
+			} else {
+				const double BG_COLOR = 53.0 / 255.0;
+				cr.set_source_rgba (BG_COLOR, BG_COLOR, BG_COLOR, 0.7);
+			}
```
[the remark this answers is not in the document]
donadigo: Great catch, this actually wasn't working at all (the condition was constant at runtime) and there were timing issues.

src/Widgets/MultitaskingView.vala
donadigo: This is just some future-proofing so that if more objects use this, it won't break. This kind of thing is also used in other classes such as `WorkspaceManager`.

src/DragDropAction.vala
donadigo: Reactivity => means that the actor catches input events such as key presses and mouse enters / leaves. I can change that comment if you want.

src/Widgets/IconGroup.vala
```diff
@@ -262,6 +262,8 @@ namespace Gala
 		 */
 		public void add_window (Window window, bool no_redraw = false, bool temporary = false)
 		{
+			var prev_n_windows = icon_container.get_n_children ();
```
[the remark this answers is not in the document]
donadigo: Good catch.

src/Widgets/IconGroupContainer.vala
donadigo: This is because if you start dragging the first icon group, the container will reallocate it's position and move to the left since there is no icon group there anymore. We need to add what was deducted back to `x` so that when you start dragging, the empty space still exists.

src/Widgets/WorkspaceInsertThumb.vala
```diff
@@ -55,7 +56,7 @@ namespace Gala
 
 					transform (false);
 				} else
-					expand_timeout = Timeout.add (EXPAND_DELAY, expand);
+					expand_timeout = Timeout.add (delay, expand);
```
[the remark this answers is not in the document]
donadigo: Not quite understanding what you want here. The constant is used in both `IconGroup` and `WindowClone` yes, do you want to have that constant only in `IconGroupContainer` and then the argument that defaults to this constant?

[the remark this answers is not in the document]
peteruithoven: 1. I'd keep that default `EXPAND_DELAY` internal (private) to the `WorkspaceInsertThumb`. 
2. Then when you want this default you simply don't customize the `WorkspaceInsertThumb.delay`, which is relevant to `IconGroupContainer.reset_thumbs` and by extension the `IconGroup` where you currently call that function with `EXPAND_DELAY`. 
3. When you need to retrieve the delay you could query the `WorkspaceInsertThumb.delay`, which is relevant for `WindowClone`

IconGroupContainer doesn't need to know the default, as long as it doesn't adapt `delay` the default is used?

src/DragDropAction.vala
peteruithoven: Hm, I didn't actually see that the actual property is called `reactive`, so I guess you're comment is actually really appropriate. Seems weird to explain this reactive term in this location.

donadigo:
@danrabbit @peteruithoven @cassidyjames all known issues fixed.

peteruithoven:
The issue where the issues would disappear is fixed indeed. 

The animation from right to left when a workspace is dropped, is a bit weird:
Screen record from 2019-07-14 00 28 35
When you just drag and drop a workspace to the same place it's somewhat less weird, because it's just the items on the right of it. 
Screen record from 2019-07-14 00 30 51

Is that something we can improve with the current Mutter API?

donadigo:
@peteruithoven I feel like this should be resolved in another PR, I believe that to make this animation seamless there will be a lot of reworking needed on how `IconGroupContainer` works internally. I would like to avoid making the diff bigger.

src/Widgets/IconGroupContainer.vala
```diff
@@ -65,14 +65,34 @@ namespace Gala
 			update_inserter_indices ();
 		}
 
+		/**
+		 * Removes an icon group "in place".
+		 * When initially dragging an icon group we remove
+		 * it and it's previous WokrspaceInsertThumb. This would make
```
peteruithoven: `WokrspaceInsertThumb` > `WorkspaceInsertThumb`

peteruithoven:
Nice! That fix helps:
Screen record from 2019-07-14 01 50 06

Issue #445, opened 2019-01-15
Wrong notification location after DPI switch


peteruithoven:
When I connect my LoDPI (1920 x 1080) projecter to my HiDPI (3840 x 2160) laptop (in mirror mode) elementary OS has to switch from HiDPI to LoDPI. This causes a couple of issues in the UI, for example the notifications location:
![screenshot from 2019-01-15 23 copy 04 17](https://user-images.githubusercontent.com/523210/51213075-64a6ae00-191a-11e9-82d5-650c6060ac71.png)

peteruithoven:
I think this is handled here: https://github.com/elementary/gala/blob/master/plugins/notify/Main.vala#L65-L73

hyphone:
This is not solved with the newest 5.1 release. When switching to LoDPI it's again to far left.

![Screenshot from 2019-12-07 10-44-52](https://user-images.githubusercontent.com/6410852/70374925-832de300-18f8-11ea-9350-2b7a8d9b634c.png)

PR #476, opened 2019-02-28

peteruithoven:
Fixes: #445


peteruithoven:
Yeah, true. I didn't want to further complicate the reviews by introducing dependencies. It order doesn't really matter but yeah, let's add #477 as a dependency.

Issue #483, opened 2019-03-01
Multitasking view bottom DPI switch issues


peteruithoven:
The bottom part of the Multitasking view kind of falls apart after switching from HiDPI to LoDPI.
![screenshot from 2019-03-01 23 43 28](https://user-images.githubusercontent.com/523210/53670933-c3cb4280-3c7c-11e9-9d16-bf0b8055ab7f.png)

peteruithoven:
https://github.com/elementary/gala/issues/243 might also solve this.

davidmhewitt:
Fixed in #1628 and #1629

PR #488, opened 2019-03-03
Multitasking view icon group DPI switch fix

peteruithoven:
Fixes: #483

Before:
screenshot from 2019-03-01 23 43 28

These first changes resizes IconGroup's on redraw, fixing part of the issue:
screenshot from 2019-03-03 16 37 28

I haven't figured out yet how these groups are positioned...


peteruithoven:
Looks like it's layed out by a Clutter.BoxLayout: https://github.com/elementary/gala/blob/master/src/Widgets/IconGroupContainer.vala#L41
Why it's not updated when the size of it's children is updated I don't know. I also can't find a way to manually trigger an update.

donadigo:
@peteruithoven I'm not sure if this will help you but MultitaskingView also takes care of repositioning the icon container: https://github.com/elementary/gala/blob/master/src/Widgets/MultitaskingView.vala#L255.

peteruithoven:
But that's just the x axis. Aligning it in the middle of the screen.

donadigo:
Yeah, you probably will have to reposition every group in the container itself manually.

davidmhewitt:
Fixed in #1628 and #1629

Issue #487, opened 2019-03-02
Format screenshot file name with scaling factor


danrabbit:
[In Screenshot](https://github.com/elementary/screenshot/blob/4b60e550c61d7aecabb0cdab3daa78fa7980d1c6/src/Widgets/SaveDialog.vala#L91), we append the display scaling factor to the file name if it's greater than 1. This is especially useful for uploading to websites which treat HiDPI images differently such as Medium.

```
            if (this.scale_factor > 1) {
                file_name += "@%ix".printf (this.scale_factor);
            }
```

PR #489, opened 2019-03-04
Add scale factor to screenshot file

peteruithoven:
Fixes: #487

### DIFF
--- a/src/ScreenshotManager.vala
+++ b/src/ScreenshotManager.vala
@@ -167,15 +167,22 @@ namespace Gala
 
 		static async bool save_image (Cairo.ImageSurface image, string filename, out string used_filename)
 		{
-			if (!Path.is_absolute (filename)) {
-				var path = find_target_path ();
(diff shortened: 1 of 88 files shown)


src/ScreenshotManager.vala
```diff
@@ -167,15 +167,22 @@ namespace Gala
...
+			if (!Path.is_absolute (used_filename)) {
+				if (!used_filename.has_suffix (".png")) {
+					used_filename = used_filename.concat (".png");
 				}
-			} else {
-				used_filename = filename;
+
+				var scale_factor = InternalUtils.get_ui_scaling_factor ();
+				if (scale_factor > 1) {
+					used_filename = used_filename.splice (-4, -4, "@%ix".printf (scale_factor));
```
donadigo: At this point the .png string should be a constant and here instead of using hardcodes, use the length of the constant.

peteruithoven: Good point. Done.

PR #527, opened 2019-04-21
Do not try to move a window to left if it is on the first workspace

### LINKED ISSUE #503: Weird workspace switcher animation after trying to move window to left of the first workspace using the move window keyboard shortcut.
elementary OS Juno fully updated.

1- Open an app in workspace 1
2- Press Super+Alt+Left to try to move the window furthest to the left of the first workspace.
3- Press Super+Right to switch to workspace 2.

**Result:** The workspace switch animation is different than a simple Super+Right. It seems that the window on workspace 1 will move to workspace 2, but it doesn't. 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/72023867-weird-workspace-switcher-animation-after-trying-to-move-window-to-left-of-the-first-workspace-using-the-move-window-keyboard-shortcut?utm_campaign=plugin&utm_content=tracker%2F62872462&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F62872462&utm_medium=issues&utm_source=github).
</bountysource-plugin>

donadigo:
Fixes #503.

Fixes a glitch where a window would be "moved" to the same workspace and then switching to the next workspace would have an animation glitch. This was caused by trying to move the window from first workspace to first workspace. It is worth noticing that this change also results in a bell when moving window from first workspace to previous. This is now consistent with the second behaviour when moving to the right.

### DIFF
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -575,8 +575,8 @@ namespace Gala
 			var next = active.get_neighbor (direction);
 
 			//dont allow empty workspaces to be created by moving, if we have dynamic workspaces
-			if ((Prefs.get_dynamic_workspaces () && 
-				Utils.get_n_windows (active) == 1 && next.index () == screen.n_workspaces - 1) || next.index () == 0) {
+			if ((Prefs.get_dynamic_workspaces () && Utils.get_n_windows (active) == 1 && next.index () == screen.n_workspaces - 1)
+				|| (active.index () == 0 && next.index () == 0)) {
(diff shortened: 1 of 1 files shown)


ricotz:
@donadigo Would not a simple `|| (active == next)` work too?

donadigo:
@ricotz yes, that works too. Updated.

PR #534, opened 2019-05-11
Do not animate check items in the right-click menu

donadigo:
Disables transition animations on check items in the right-click menu to prevent them from animating when right-clicking on the titlebar.


danrabbit:
This seems like more of a style sheet bug. I’m not sure it’s a good idea to add workarounds to Gala like this

donadigo:
@danrabbit submitted here: https://github.com/elementary/stylesheet/pull/505. Closing.

PR #535, opened 2019-05-14
Notify plugin control server

donadigo:
This adds a plugin control DBus server which exposes one method to set the y offset of the notification stack. If the wingpanel expands, it can call this method to make notifications update their position so that they don't overlap. This is needed by https://github.com/elementary/wingpanel/pull/212 to work correctly.


jeremypw:
@donadigo Converting to draft as has conflicts and/or no recent activity.  Please feel free to close or reopen for review after fixing conflicts as desired.

danirabbit:
Closing since we haven't had movement here in 2 years and notifications are their own process now

Issue #68, opened 2017-09-03
Wayland: design window communication API


donadigo:
We should design a way to inspect and interact with windows that are currently running, possibly through DBus.

Couple of features that I think should be implemented within the server:
1. Get a list of currently opened windows
2. Get information about a particular window:
 * Title
 * State (minimized, maximized)
 * Type (normal, desktop, dock etc.)
 * Icon
 * Size and position
 * The workspace it's currently on
 * The screen it's currently on
 * Process ID
 * Where in the window stack the window is (on top, on bottom)
 * Possibly a unique ID (not needed with DBus)

3. Performing different actions on a particular window:
 * Minimizing
 * Unminizing
 * Unmaximizing
 * Maximizing
 * Activating
 * Closing
 * Moving to a particular workspace
 * Changing it's size and position (for dock windows)

4. Receiving information about events:
 * Window workspace changed
 * Window closed
 * Window state changed (minimized, maximized etc.)
 * Window name changed

Feel free to discuss the API in the comments.

matzipan:
Who can access this dbus? Any process about any window?

donadigo:
I think any application can now access this information through libbamf and libwnck, so there is no difference although, since we will be implementing this API, we can restrict the access to this information.

matzipan:
> I think any application can now access this information through libbamf and libwnck

Sure, but this is a problem and we shouldn't propagate this to wayland-land.

trickreich:
Any process on this?

properlypurple:
This should be a good starting point
https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/master/stable/xdg-shell/xdg-shell.xml

Congee:
It's been ... one year since the last reply 😕

orowith2os:
What are the topics that need to be discussed here now? The Wayland protocol (wayland and wayland-protocols) should handle everything for clients, and for the shell, you can look into how KDE/kwin and Sway/wlroots handle it. 

Being based on Mutter, it might be preferable to investigate how GNOME Shell manages everything.

lenemter:
@orowith2os Currently discussions are taking place in https://github.com/elementary/gala/issues/1693.

> and for the shell

I think Mutter doesn't allow that by default. See https://gitlab.gnome.org/GNOME/mutter/-/issues/1922 and https://gitlab.gnome.org/GNOME/mutter/-/issues/1053.

> you can look into how KDE/kwin and Sway/wlroots handle it

[...]

PR #539, opened 2019-05-18
Add style transition API

donadigo:
Adds a new method to the DBus API called `transition_from_snapshot` for apps (and primarily the switchboard plug) to be able to create a smooth stylesheet transition. The transition will not happen when animations are disabled but the snapshot will still be created and added (that is it will stay for 200ms and hide without any animations). 

The code should be quite optimized for the task, I previously used a simple `Cogl.Texture.get_data` for creating a static texture but it turned out to be quite long to execute: transferring pixel data from GPU to CPU took ~15ms on my machine which is little bit above 60FPS but there were longer runs that caused visible lag when creating textures for maximized windows. We now copy the texture directly on the GPU which takes about ~2ms.

**Update**: the code has been updated to Cogl2 and to automatically transition when the style changes, however there still seems to be the problem of some elements flickering while transitioning. I think this is a matter of timing things correctly.


danrabbit:
I guess an obvious question is, what about Wayland? Do we get window IDs like this once we move away from X?

donadigo:
1. No one knows. See https://github.com/elementary/gala/issues/68. I'm not sure if mutter will do anything about this but it seems like there's no standard for those things either. I also don't know how mutter on wayland handles window ID's.

2. We can optionally separate the other part of the method to a new one but it will be still public though.

donadigo:
@danrabbit updated.

cassidyjames:
Does this need to be an external API at all? Can't Gala just listen to the setting and do this transition automatically?

donadigo:
@cassidyjames yeah that and if gala listened to that setting you have to account for the fact that the signal could fire later than the theme change, causing a jarring flick before the actual transition. That's why it's safer to have the app call before actually changing the setting. If we had a Gtk fork / patch then perhaps that would be possible to do automatically.

donadigo:
Branch updated to use the newest Cogl2 API and updated to master.

donadigo:
Ready for review again.

danrabbit:
Hm, I can't really see that this smooths anything out for me. Is there something I need to do other than install this branch and restart my session?

donadigo:
@danrabbit hm, that's weird. Installing the branch & restarting should be enough. Is the transition like instant, or do you see at least some trace of the animation?

lib/Constants.vala
```diff
@@ -30,5 +30,7 @@ namespace Gala {
         SNAP = 250,
         // Duration of the workspace switch animation
         WORKSPACE_SWITCH = 300,
+        // Style switch animation
+        STYLE_SWITCH = 200
```
cassidyjames, line 34: 200 would feel nice and snappy, but it seems to not be slow enough to mask all the GTK transitions. Something like 500 seems to mask them better.

```suggestion
        STYLE_SWITCH = 500
```

src/DBus.vala
```diff
@@ -256,5 +296,125 @@ namespace Gala {
...
+            }
+        }
+
+        static async void transition_window (Meta.Window window, bool animate) {
+            /**
+             * Because Gtk takes time to apply the styling to all elements,
+             * we will simply define an arbitrary wait time for this to happen
+             * to smoothen out the transition.
+             */
+            const uint STYLE_CHANGE_COMPLETION_DURATION = 150;
```
cassidyjames, line 359: I've been playing with this, and 300 seems to be long enough to mask WebKit transitioning as well which is a really nice effect.

```suggestion
            const uint STYLE_CHANGE_COMPLETION_DURATION = 300;
```

lib/Constants.vala
cassidyjames: Oh, nevermind, ignore me. 200 works fine if we increase the delay below slightly.

cassidyjames:
@donadigo would it be possible to resolve conflicts here? I'm not sure why we weren't seeing it work in the past, but now that we have all the other dark style bits merged in, this would be an incredibly nice effect.

JoseExposito (changes requested):
I didn't notice there was a PR open with this functionality, definitely something really cool to have! 
It's most likely something related to my set up, but I'm seeing some noise in the panel and shadows:

https://user-images.githubusercontent.com/1335948/126907909-8545c1cd-9f28-4c14-886c-61a7b432f57d.mp4

danirabbit:
Gonna close this since there was a lot more progress on #1242

Issue #404, opened 2018-11-05
Not able to display an application/video on second monitor in fullscreen.


andreasfelix:
Since Juno it is not longer possible to display some applications (like chrome/firefox) in fullscreen on a second monitor.

However it is possible to start the application or video in fullscreen on the first monitor and than switch it to the second monitor via the shortcut super+shift+right-arrow.

jamesdwilson:
I think I am experiencing the same issue. 
I have my builtin laptop display marked as a non-primary display (not-starred in Displays), and if I play videos on it, it works, but if I fullscreen on the laptop display, it is very flakey and will most of the time not go fullscreen correctly. It will work like 1/5-1/10 times.

However, if I fullscreen the video on my primary monitor and then escape fullscreen, and then move the window to my laptop display, it can fullscreen repeatedly with no issues.
uname -a:


Linux xps 4.15.0-38-generic #41-Ubuntu SMP Wed Oct 10 10:59:38 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

elementary OS 5.0 Juno
GTK+ 3.22.30
NVIDIA 390.77 in performance mode

Dell XPS 9570

Cariffo:
I am also experiencing this issue. I have a Laptop with built-in display as primary monitor and a secondary monitor connected to it.

I can start chrome/ firefox/ epiphany (issue is same on all of these) and fullscreen on the primary monitor without any issues. 
But if i have chrome on my secondary monitor and fullscreen a video, it freezes (sound keeps playing, only the video is bad). I cant recover from this, i have to exit chrome on that display completely because only exiting the fullscreen leaves it lagging.

A workaround to this is to start the fullscreen on primary and then move it to the secondary monitor as suggested above... but this is not a fix! this needs attention asap

Using: 
Elementary os 5.0 Juno
GTK+ 3.22.30
Linux 4.15.0-39-generic
Nvidia 415.13

MSI GP60 2QF

Cariffo:
Hey there! I have an update... the issue is still very much present but today i figured out it only occurs when you are trying to fullscreen a video from a maximizes window on the second monitor... if i fullscreen a video from a non-maximized video it works fine

...still... needs fixin

Thalhammer:
Experienced this as well:
I have a Notebook with an Intel/Nvidia combo (Optimus).
My Primary display is 1920x1080 internal connected via eDP.
My secondary display is 1920x1200 connected via an HDMI to VGA adapter on HDMI0.
[...]

PR #540, opened 2019-05-19
Fix fullscreen a maximized chromium window on secondary screen

andreasfelix:
~~Temporary~~ Fix for #404.

### DIFF
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -972,6 +972,8 @@ namespace Gala
 			ulong position_signal_id = 0U;
 			size_signal_id = window.size_changed.connect (() => window_change_complete (actor, which_change, size_signal_id, position_signal_id));
 			position_signal_id = window.position_changed.connect (() => window_change_complete (actor, which_change, size_signal_id, position_signal_id));
+			//  TEMPORARY FIX: Issue 404
+			if (which_change == Meta.SizeChange.UNFULLSCREEN || which_change == Meta.SizeChange.FULLSCREEN) window_change_complete (actor, which_change, size_signal_id, position_signal_id);
 		}
 
 		void window_change_complete (Meta.WindowActor actor, Meta.SizeChange which_change, ulong size_signal_id, ulong position_signal_id) {
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -965,7 +965,7 @@ namespace Gala
 			unowned Meta.Window window = actor.get_meta_window ();
 
 			if (which_change == Meta.SizeChange.UNFULLSCREEN || which_change == Meta.SizeChange.FULLSCREEN) {
-					handle_fullscreen_window (window, which_change);
+				handle_fullscreen_window (window, which_change);
 			} else if (window.get_tile_match () == null) { // don't animate windows with tiled match
 				ulong size_signal_id = 0U;
 				ulong position_signal_id = 0U;
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -963,21 +963,18 @@ namespace Gala
 		public override void size_change (Meta.WindowActor actor, Meta.SizeChange which_change, Meta.Rectangle old_frame_rect, Meta.Rectangle old_buffer_rect)
 		{
 			unowned Meta.Window window = actor.get_meta_window ();
-			if (window.get_tile_match () != null) {
-				size_change_completed (actor);
-				return;
-			}
 
-			ulong size_signal_id = 0U;
-			ulong position_signal_id = 0U;
-
-			//  TEMPORARY FIX: Issue 404
 			if (which_change == Meta.SizeChange.UNFULLSCREEN || which_change == Meta.SizeChange.FULLSCREEN) {
-				window_change_complete (actor, which_change, size_signal_id, position_signal_id);
-			} else {
+					handle_fullscreen_window (window, which_change);
+			} else if (window.get_tile_match () == null) { // don't animate windows with tiled match
+				ulong size_signal_id = 0U;
+				ulong position_signal_id = 0U;
 				size_signal_id = window.size_changed.connect (() => window_change_complete (actor, which_change, size_signal_id, position_signal_id));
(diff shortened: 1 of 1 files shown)


andreasfelix:
I used this branch from some time now and didn't notice any side effects or other glitches. So I think it could be used as temporary fix till someone with more experience provides a better/cleaner solution...

andreasfelix:
I have added a commit according to your suggestion, where signals are not connected when operation is fullscreen/unfullscreen.

But I think the lines
```vala
ulong size_signal_id = 0U;
ulong position_signal_id = 0U;
```
are needed in both cases. Otherwise I get a compilation error:
```
../src/WindowManager.vala:973.66-973.83: error: The name `position_signal_id' does not exist in the context of `Gala.WindowManagerGala.size_change'
				window_change_complete (actor, which_change, size_signal_id, position_signal_id);
```
So i put these lines in front of the if-else statement.

andreasfelix:
After spending some time reading the source code and older PRs, I think I figured out a proper fix:

**Overview of the problem:**
The issue was introduced with #239, which was needed to fix the window animations. These weren't working as expected, because of an internal change of mutter, which caused the `Meta.Window.get_frame_rect` to update after the `Meta.Plugin.size_change` call. This led to the problem that the window transitions were using the old window size instead of the new window size. So it looked as no animation would happen.

The idea of #239 to fix this was to use the `size_changed` signal of the `Meta.Window`, so that in a later callback `get_frame_rect` would return the new window size. While this worked as expected for *(un)-maximizing* or *tiling* a window, it has a "small logical error": 

`Meta.Plugin.size_changed` also handles events where the size of the window must not change in general. For these events `Meta.Plugin.size_changed_completed` is not called, which practically disables these actions. This cases are:

#341: Moving a tilled window, which does not change its size. This was already solved (#387) by also registering `window_change_complete` to the `Meta.Window.position_changed` changed signal.

#404: Fullscreening an already maximized window, which also does not change its size. The reason this issue does **not** occur on a primary screen is because of the *wingpanel*. After fullscreening a maximized window on a primary screen the window actually increases its size by the area of the wingpanel, so that the `Meta.Window.size_changed` signal gets emitted anyway. This is not the case for a secondary display, which does not have a wingpanel! 

To reproduce this issue on a primary screen just `killall cerbere` and the `killall wingpanel`.

**One Solution (this PR)**
(Un-)Fullscreen does not need to know the s the rectangle that bounds window, so there is no need to connect these cases to the `size/position-changed` events (which are only used to await this property).

**New proposed logic:**

```
// pseudo code
size_change
    If (FULLSCREEN or UNFULLSCREEN)
       handle_fullscreen_window()
    else if (no matching tilled window)
      connect to positon/size_changed --> size_change_complete gets called by callback
      return;
    size_change_complete()
```

**Questions**
I am not completely sure, if there could be other edge cases, as I can't find the `Meta.Plugin.size_change_completed` in the [MutterAPI reference 3.16.1.1](https://developer.gnome.org/meta/stable). Isn't there a newer one? 

Where can I find a list of all possible values for `Meta.SizeChange`? If there are only `FULLSCREEN`, `UNFULLSCREEN`, `MAXIMIZE` and `UNMAXIMIZE`, I think this branch covers all cases.

It would be nice to get some feedback from @donadigo, @danrabbit, @peteruithoven and @ricotz as you already fixed the related issues #239 and #387.

**Other solution:**
Another possibility would be to implement a similar method to `Meta.Window.get_frame_rect` for the new window size in Mutter, so the callback work-around would not be needed. This would also prevent future edge cases.

peteruithoven:
@andreasfelix thanks for looking into this so thoroughly.  
I wanted to reproduce #404 so I killed the Wingpanel (I recommend [this method](https://elementary.io/docs/code/os-dev#wingpanel)) and maximized and then fullscreen Terminal. It seems to work fine though? Also with other apps. 
Screen record from 2019-07-11 16 18 26

Am I doing something different?

andreasfelix:
@peteruithoven try to log the `size_change_completed` call. You will see that the callback it is not called for maximized windows when fullscreening, instead the callbacks stack up until the next `position/size-changed` signal is emitted.

Some applications can cope with that (like terminal). For some applications the glitch only occurs sometimes (for me code, files, firefox) In some applications there are only minor glitches (epiphany). 

For me it always happens with a Chromium window with disabled system titlebar and borders.

andreasfelix:
@peteruithoven, Also the applications will always correctly maximizes, but will not be usable. Try to type something in the terminal or scroll through some folders in files.

Sometimes this also happens for the terminal for me.

donadigo:
@andreasfelix thank you for such detailed investigation. For all values of `SizeChange` please refer to the vapi: https://github.com/elementary/gala/blob/master/vapi/libmutter.vapi#L1393, these 4 are indeed all the values there are. Unfortunately I don't have access to a multi monitor setup, so I can't test this, but I will revise the logic again when reviewing.

andreasfelix:
@donadigo to test this on a single monitor kill the wingpanel.

andreasfelix:
To show that I am not the only one who experiences this issue, I listed the others reports. 
- https://www.reddit.com/r/elementaryos/comments/a5jme6/fullscreen_video_freezes_on_external_monitor/
- https://www.reddit.com/r/elementaryos/comments/bugb2v/having_a_problem_with_gala_and_fullscreening/
- https://www.reddit.com/r/elementaryos/comments/aoxof2/the_most_annoying_bug_on_elementary_os_judo_50/

- https://elementaryos.stackexchange.com/questions/16526/unable-to-fullscreen-video-on-second-monitor
- https://elementaryos.stackexchange.com/questions/16595/going-fullscreen-with-a-browser-locks-up-that-window-only-on-secondary-display
- https://www.reddit.com/r/elementaryos/comments/9y62ca/cant_fullscreen_a_video_on_secondary_monitor/
- https://elementaryos.stackexchange.com/questions/17082/instability-in-full-screen-videos-elementaryos-5
- https://elementaryos.stackexchange.com/questions/15561/google-chrome-stuck-when-watching-video-at-full-screen-size

The corresponding issue #404 is also the most commented gala-issue within the last year. I don't want to push anyone, but I would appreciate it, if somebody could review this branch, so this bug gets finally fixed. Thank you!

PR #552, opened 2019-06-13
Notify rework

### LINKED ISSUE #703: Notification windows should always appear on top
In the notifications plug we keep notifications on screen all the time, but notification windows seem to be confined to the workspace they spawn on

### LINKED ISSUE #704: Notification windows have window context menus
They should not

cassidyjames:
Fixes #703
Fixes #704 

@donadigo's work. Creating a draft PR for visibility.

- [ ] In the Gala plugin bubbles also appeared in the multitasking view. I think we probably want to retain this behavior


danrabbit:
- [ ] Bubbles should be persistent across workspaces. At the moment if I switch workspaces, I lose the bubble
- [ ] In the Gala plugin bubbles also appeared in the multitasking view. I think we probably want to retain this behavior

cassidyjames:
@danrabbit for the first item, can/should you just set that property on the bubble window itself?

I'm kind of indifferent about the multitasking view. I honestly think it would be better to delay the notification until out of the multitasking view, as there's not anything you can really do from inside it without exiting anyway, right? That would be different for confirmations I guess, though.

danrabbit:
@cassidyjames You can request a `stick` yeah, but it seems like Gala should just handle the window management here since I don't think we want notifications windows to ever not stick

danrabbit:
Even though it's not perfect yet, I'm inclined to merge this since it makes it possible to test the Gtk notification server and with the new key it doesn't effect existing users

danrabbit:
If you switch workspaces while a notification is animating, you get stuck with the notification locked into a skew/rotate state :)

PR #556, opened 2019-06-16

peteruithoven:
Fixes: #445
Supersedes: https://github.com/elementary/gala/pull/476


peteruithoven:
Hm, not sure what happened here, I should have checked the changes.

PR #559, opened 2019-06-17
search in window overview

### LINKED ISSUE #218: Windows overview mode should be searchable like Unity
Unity can search and filter windows in the overview mode like this.
screenshot from 2018-03-30 13-14-49
screenshot from 2018-03-30 13-16-26

It is extremely helpful when a lot of windows are open. It would be great if elementary also had this feature.

ranfdev:
This closes #218 
There are still some problems:
- If there are 2 windows with the same title, you can only search for the last used one.
- Probably the design must be improved.
search_in_overview


danirabbit:

Issue #561, opened 2019-06-22
BackgroundCache broken


peteruithoven:
While looking into #13 I noticed that when I connected a second monitor the BackgroundManagers where destroyed twice which breaks the `use_count` of the BackgroundCache instance. This leads to BackgroundSource instance being discarded and recreated needlessly and possibly causing #13

1. When a monitor is connected `screen.monitors_changed` is triggered.
2. BackgroundContainer is connected to this and will destroy all existing BackgroundManager (`destroy_all_children`) and re-create one per monitor. 
3. This `destroy_all_children` calls `BackgroundManager.destroy`, but I noticed it was called twice. Once during the `BackgroundContainer.update` and once directly afterwards. 

Looking into this destroy using gdb ([more info](https://github.com/elementary/gala/issues/13#issuecomment-504475478)) I figured out `BackgroundManager.destroy` was called twice from `BackgroundContainer.update` First one through a `clutter_actor_destroy` > `g_object_run_dispose` and the second time through a `g_object_unref`. 
The second time seems like some kind of automatic memory cleanup step?

Anyway what this triggers is:
1. `BackgroundManager.destroy` will release the BackgroundCache instance from `BackgroundCache`. (which decrements `use_count`)
2. A new `BackgroundManager` is created which retrieves the BackgroundCache instance from `BackgroundCache`(which increments `use_count`)
3. `BackgroundManager.destroy` is called again, releasing the BackgroundCache instance from `BackgroundCache`. (which decrements `use_count` to a invalid number)

This invalid lower `use_count` value means that at some point the BackgroundSource instance is discarded and needs to be recreated, possibly causing #13.

For reference, the relevant Background classes and how they relate:
![Gala background uml](https://user-images.githubusercontent.com/523210/59969128-1b046400-9547-11e9-9430-b7fda05ffd5e.png)

Issue #13, opened 2017-05-23
Selected wallpaper is not applied 


x1unix:
This issue is redirected from https://github.com/elementary/switchboard-plug-pantheon-shell/issues/61#event-1094577982.

Hardware acceleration and virtualization are enabled.
Video memory: 768 MB.

Nine-H:
this has been happening to me too.

Nine-H:
steps to reproduce:
1. plug in external vga monitor while running.

I'm curious what happens with hdmi and multi-monitor setups.

nkkollaw:
I was actually never able to set one in Loki (ok in Freya), even without a monitor attached.

jlnr:
I've had it happen both in VMware Fusion, but also in my native/real installation that uses the proprietary Nvidia drivers.

cassidyjames:
This seems to be remedied by restarting Gala. It's like Gala stops properly updating the wallpaper, no matter which source you're setting it from. I've seen this when switching displays (plugging/unplugging a monitor), but not reliably enough to track it down.

x1unix:
Have to check it this issue is applicable to the latest EOS release.

rugglcon:
@x1unix this is confirmed an issue on the latest release

J0Ar:
This is really a weird bug...
I saw this bug report and tried myself to change the wallpaper...and came to this...
https://streamable.com/s4ily

(do not know if this works with the video upload...if anybody know a better alternative...tell me!^^ Produced GIF was too big =/ )

rugglcon:
That happens to me sometimes as well.

peteruithoven:
> This seems to be remedied by restarting Gala.

This can be easily done by executing `gala -r` in the Applications menu.

@donadigo do you perhaps know the chain of events of how this should happen? Is it just Gala? If Gala is responsible I'd love to add some logging so that we can check journalctl afterwards.

donadigo:
I'm not really into wallpaper code for gala, but what I do know is I've experienced some issues as well with changing them. Sometimes gala even crashes. I'm pretty sure it is an issue within gala though.

There's a report with some logs above, with a method that fails an assertion, maybe we should look at the produced C code and go into that direction.

peteruithoven:
The logs from other (duplicate) issues:

Using Desktop system settings, syslog:
```
[...]

PR #562, opened 2019-06-22
Prevent double BackgroundManager destroy

peteruithoven:
Fixes: #561
Fixes: #13

So this feels a bit hacky, but I don't know of a better way to prevent this automatic second destroy call. See #561 for more information.


peteruithoven:
@donago That seems like a way nicer solution and it also fixes the issue.

PR #564, opened 2019-06-25

peteruithoven:
Fixes: #445
Supersedes: #556


plugins/notify/NotificationStack.vala
```diff
@@ -31,15 +31,22 @@ namespace Gala.Plugins.Notify
 
 		public Screen screen { get; construct; }
 
+		public int width
```
donadigo: If you solved it this way, you should hide the original property and also change the type to match.

```suggestion
		public new float width
```

Issue #573, opened 2019-07-01
Windows that start maximized or fullscreened should respect “Move windows to a new workspace” setting


Blast-City:
1- Open Terminal and other app.
2- Press F11 to put Terminal in fullscreen mode. Terminal moves to a new workspace.
3- Close Terminal.
4- Return to workspace in step 1 and open terminal again.

**Result:** Terminal opens in fullscreen mode, but doesn't move to a new workspace.

Relevant information:
#570
#571

cassidyjames:
I _think_ I agree to this, especially now that we have optional maximize-to-new-workspace behavior; it's weird how sometimes (when explicitly maximizing) a window will go to a new workspace, but when opening a previously-maximized app it stays on the current workspace on top of my other windows.

Could use a second opinion from @elementary/ux though.

cassidyjames:
I came to report this today after being annoyed by it again. Based on @danrabbit’s :+1: I'm gonna mark this as confirmed.

Issue #567, opened 2019-06-27
Allow fullscreen apps have their CSD


aggalex:
When apps go fullscreen (Like if you press f11 on elementary code) the apps are stripped off their CSD, which contain functions citical to the UI of the app. Please make it so that the app will have its CSD while fullscreen.

Blast-City:
Hi.

Did you set a fullscreen shortcut in the keyboard settings?

More info: elementary/switchboard-plug-keyboard#206

aggalex:
No
![image](https://user-images.githubusercontent.com/11951328/60351090-afe6e180-99cd-11e9-9a35-d6af65ed07e7.png)

Blast-City:
Oh Ok. I've only seen this issue when force fullscreen is set. For example Files. When an user force fullscreen, the csd disappears, but because by default it doesn't have a fullscreen shortcut (F11) this problem doesn't happen.

What other native apps did you saw this behaviour?

aggalex:
code, terminal, GIMP, inkscape, aisleriot. Also firefox, epiphany and photos, but these are intentional I believe.

Blast-City:
> code, terminal, GIMP, inkscape, aisleriot. Also firefox, epiphany and photos, but these are intentional I believe.

Hi.

Some of those apps are not native, but i can see that terminal doesn't show the csd in fullscreen.

The problem is, wouldn't show csd make fullscreen mode and maximize the same?

You should see the following thread. I think it resolves your issue.
elementary/wingpanel#93

aggalex:
I'm not saying to show csd at all times, just when the cursor goes to the top of the screen, only then will the CSD controls come down from the top like plank, just like in Mac OS

Blast-City:
> I'm not saying to show csd at all times, just when the cursor goes to the top of the screen, only then will the CSD controls come down from the top like plank, just like in Mac OS
> 
> BTW, I take it that by native you mean curated I guess

Hi.

Yeah i meant apps that are made by elementary. Sorry.
The link that I provided above seems to correspond to your issue.

quequotion:
> I'm not saying to show csd at all times, just when the cursor goes to the top of the screen, only then will the CSD controls come down from the top like plank, just like in Mac OS

[...]

Issue #231, opened 2018-04-21
Fullscreen to new workspace feature should be optional


codygarver:
Would be nice to be able to disable this feature, it's not great on multi-monitor setups because I treat them as static workspaces.

PR #578, opened 2019-07-01
Move maximized windows to their own workspace

### LINKED ISSUE #570: Move maximized windows to their own workspace
[Blockbuster](https://github.com/donadigo/blockbuster) has a feature where you can set an option to have maximized windows automatically moved to another workspace.

This requires working out some details, namely:
* Move the window to a new workspace that's created next to the current one or
* Move the window to the last workspace
* Closing that maximized window probably should remove the workspace it's on, just as it happens in the mentioned app

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/76195972-move-maximized-windows-to-their-own-workspace?utm_campaign=plugin&utm_content=tracker%2F62872462&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F62872462&utm_medium=issues&utm_source=github).
</bountysource-plugin>

donadigo:
Fixes #570. 

Uses the exact same behaviour as fullscreen windows do. Maximizing a window moves it to the next nearest new workspace, unmaximizing it moves back to the previous workspace.

The behaviour got moved to a new function and therefore there is one difference in fullscreen behaviour:
in master a fullscreen window would *always* be moved to the next workspace, even if the original workspace would be empty after moving, this doesn't happen anymore and is more inline with the maximize behaviour.

Testable also with other experimental features in https://github.com/elementary/gala/tree/intelligent-workspaces.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,1 +1,0 @@
-Contributors to this repo agree to be bound by the [Code of Conduct](https://elementary.io/code-of-conduct) as made available on our website.
--- a/daemon/Main.vala
+++ b/daemon/Main.vala
@@ -15,132 +15,132 @@
 //  along with this program.  If not, see <http://www.gnu.org/licenses/>.
 //
 
-namespace Gala
-{
-	[DBus (name = "org.gnome.SessionManager")]
-	public interface SessionManager : Object
-	{
-		public abstract async ObjectPath register_client (string app_id, string client_start_id) throws DBusError, IOError;
-	}
+namespace Gala {
+    [DBus (name = "org.gnome.SessionManager")]
+    public interface SessionManager : Object {
+        public abstract async ObjectPath register_client (
+            string app_id,
+            string client_start_id
+        ) throws DBusError, IOError;
+    }
 
-	[DBus (name = "org.gnome.SessionManager.ClientPrivate")]
-	public interface SessionClient : Object
-	{
-		public abstract void end_session_response (bool is_ok, string reason) throws DBusError, IOError;
+    [DBus (name = "org.gnome.SessionManager.ClientPrivate")]
+    public interface SessionClient : Object {
+        public abstract void end_session_response (bool is_ok, string reason) throws DBusError, IOError;
 
-		public signal void stop () ;
-		public signal void query_end_session (uint flags);
-		public signal void end_session (uint flags);
-		public signal void cancel_end_session ();
-	}
+        public signal void stop () ;
+        public signal void query_end_session (uint flags);
+        public signal void end_session (uint flags);
+        public signal void cancel_end_session ();
+    }
 
-	public class Daemon
-	{
-		SessionClient? sclient = null;
+    public class Daemon {
+        SessionClient? sclient = null;
 
-		public Daemon ()
-		{
-			register.begin ((o, res)=> {
-				bool success = register.end (res);
-				if (!success) {
-					message ("Failed to register with Session manager");
-				}
-			});
+        public Daemon () {
+            register.begin ((o, res)=> {
+                bool success = register.end (res);
+                if (!success) {
+                    message ("Failed to register with Session manager");
+                }
+            });
 
-			var menu_daemon = new MenuDaemon ();
-			menu_daemon.setup_dbus ();
-		}
+            var menu_daemon = new MenuDaemon ();
+            menu_daemon.setup_dbus ();
+        }
 
-		public void run () {
-			Gtk.main ();
-		}
+        public void run () {
+            Gtk.main ();
+        }
 
-		public static async SessionClient? register_with_session (string app_id)
-		{
-			ObjectPath? path = null;
-			string? msg = null;
-			string? start_id = null;
+        public static async SessionClient? register_with_session (string app_id) {
+            ObjectPath? path = null;
+            string? msg = null;
+            string? start_id = null;
 
-			SessionManager? session = null;
-			SessionClient? session_client = null;
+            SessionManager? session = null;
+            SessionClient? session_client = null;
 
-			start_id = Environment.get_variable ("DESKTOP_AUTOSTART_ID");
-			if (start_id != null) {
-				Environment.unset_variable ("DESKTOP_AUTOSTART_ID");
-			} else {
-				start_id = "";
-				warning ("DESKTOP_AUTOSTART_ID not set, session registration may be broken (not running via session?)");
-			}
+            start_id = Environment.get_variable ("DESKTOP_AUTOSTART_ID");
+            if (start_id != null) {
+                Environment.unset_variable ("DESKTOP_AUTOSTART_ID");
+            } else {
+                start_id = "";
+                warning (
+                    "DESKTOP_AUTOSTART_ID not set, session registration may be broken (not running via session?)"
+                );
+            }
 
-			try {
-				session = yield Bus.get_proxy (BusType.SESSION, "org.gnome.SessionManager", "/org/gnome/SessionManager");
-			} catch (Error e) {
-				warning ("Unable to connect to session manager: %s", e.message);
-				return null;
-			}
+            try {
+                session = yield Bus.get_proxy (
+                    BusType.SESSION,
+                    "org.gnome.SessionManager",
+                    "/org/gnome/SessionManager"
+                );
+            } catch (Error e) {
+                warning ("Unable to connect to session manager: %s", e.message);
+                return null;
+            }
 
-			try {
-				path = yield session.register_client (app_id, start_id);
-			} catch (Error e) {
-				msg = e.message;
-				warning ("Error registering with session manager: %s", e.message);
-				return null;
-			}
+            try {
+                path = yield session.register_client (app_id, start_id);
+            } catch (Error e) {
+                msg = e.message;
+                warning ("Error registering with session manager: %s", e.message);
+                return null;
+            }
 
-			try {
-				session_client = yield Bus.get_proxy (BusType.SESSION, "org.gnome.SessionManager", path);
-			} catch (Error e) {
-				warning ("Unable to get private sessions client proxy: %s", e.message);
-				return null;
-			}
+            try {
+                session_client = yield Bus.get_proxy (BusType.SESSION, "org.gnome.SessionManager", path);
+            } catch (Error e) {
+                warning ("Unable to get private sessions client proxy: %s", e.message);
+                return null;
+            }
 
-			return session_client;
-		}
+            return session_client;
+        }
 
-		async bool register ()
-		{
-			sclient = yield register_with_session ("org.pantheon.gala.daemon");
+        async bool register () {
+            sclient = yield register_with_session ("org.pantheon.gala.daemon");
 
-			sclient.query_end_session.connect (() => end_session (false));
-			sclient.end_session.connect (() => end_session (false));
-			sclient.stop.connect (() => end_session (true));
+            sclient.query_end_session.connect (() => end_session (false));
+            sclient.end_session.connect (() => end_session (false));
+            sclient.stop.connect (() => end_session (true));
 
-			return true;
-		}
+            return true;
+        }
 
-		void end_session (bool quit)
-		{
-			if (quit) {
-				Gtk.main_quit ();
-				return;
-			}
+        void end_session (bool quit) {
+            if (quit) {
+                Gtk.main_quit ();
+                return;
+            }
 
-			try {
-				sclient.end_session_response (true, "");
-			} catch (Error e) {
-				warning ("Unable to respond to session manager: %s", e.message);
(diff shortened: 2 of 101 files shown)


danrabbit:
This works as expected as far as I can tell. Great improvement! UX Approve :)

worldofpeace:
I've noticed that I'd expect a a window to moved to it's own workspace if it's launched maximized.
For example when I launch `firefox` the window is already maximized but it isn't moved to a separate workspace. I would then have to unmaximize it and maximize it to get the desired behavior.

(can reproduce with actual gtk applications as well).

Blast-City:
> I've noticed that I'd expect a a window to moved to it's own workspace if it's launched maximized.
> For example when I launch `firefox` the window is already maximized but it isn't moved to a separate workspace. I would then have to unmaximize it and maximize it to get the desired behavior.
> 
> (can reproduce with actual gtk applications as well).

#573

ryanprior:
---
in-response-to: https://github.com/elementary/gala/pull/578
---

I think we should not adopt this behavior. Since no rationale is given for adopting it, either in this this PR or in the linked issue, I'll begin with my best arguments in favor of adopting it, then explain why I think the arguments against are stronger. Perhaps my understanding of the arguments in favor is flawed, so feel free to expand.

#### Reasons in favor of moving maximized windows into their own workspaces

The first that comes to mind is that macOS does this. The general principle that it's beneficial to behave consistently with another popular OS does favor this approach.

The next thing I think about comes from the domain of mobile devices, where it's common for apps to run full-screen and to exclude anything else. If we make that principle applicable to the desktop—that an app which runs full screen should exclude everything else and stand on its own—then we arrive at something like the proposed behavior.

Finally we might argue, from a spatial metaphor, that laying out full screen windows side-by-side across the virtual table of our workspaces is superior to creating a big pile of full screen windows in front of the user's chair. In this way, switching between full screen apps is like rolling your chair back and forth rather than shuffling through the pile to get the app you want.

#### Reasons against

In practice this complicates the spatial model for workspaces and introduces workflow-breaking behaviors. Returning to the spatial metaphor: when you organize sheets on a table, whether into piles or into a layout spread, you make this choice by selecting the number of piles you want and dragging things into them, then flipping through whatever pile you're interested in. This flow is what we do with the multitasking view and window switcher at present.

Maximizing a window is like turning special attention to it. The user signifies that they want this content to take up maximum screen space. In terms of the spatial metaphor, that's like picking up a sheet of material and holding it in front of you so you so it takes up more space.

Suppose, though, that you want to take a peek at something that was in the same pile as the currently-maximized window; but without you asking, the pile has been swept to the side, so when you go to grab from it, there's no pile there. Your normal focus-switching mechanism is broken. Instead, you have to go look for another pile.

It gets worse. Suppose you've created two workspaces for yourself, the first for accounting and the second for communicating. Occasionally, you need to flip between these, such as when you're communicating the result of some accounting task. Because the workspaces use a spatial metaphor, you quickly form the connection that "switch left" brings you to your accounting view, and "switch right" brings you to your communicating view. However, if you maximize the spreadsheet in your accounting view, it immediately breaks the spatial metaphor by introducing a new third pile that you didn't intend to create. Now, to switch between tasks, you have to think about whether you're in the spreadsheet or not, and use either one or two switch commands. The more maximized windows you use, the more complicated this model becomes, and you're more likely to enter the multitasking view and hunt for the workspace you want rather than switching confidently.

The model that auto-moves maximized windows into their own workspace is also inconsistent with the use of the tiling window manager. If I tile one window to the left half of my screen, it doesn't and shouldn't flip to a new workspace. If I tile a window to the right half of my screen instead, it similarly shouldn't do that. But if I tile windows both left and right, creating what's essentially a full-screen display, the logic of maximized-window-as-workspace vaguely suggests that those two windows should now comprise their own workspace.

#### Analysis

The spatial metaphor for workspace organization works if users are intentional about when they create new spaces and switch confidently between them. Creating a new workspace without the user's explicit direction undermines the user's confidence and complicates the metaphor, making it less useful.

When a user maximizes a window, that act communicates the intent that the window should take up as much screen space as possible. It doesn't communicate the intent that all other windows should be swept off of the current workspace. It doesn't communicate the intent that alt-tab should now behave differently from before, nor that the user now wants to organize their work in two spatial piles instead of one (or, inductively, N+1 instead of N.)

Many common workflows flip between a maximized window (like a browser) and another maximized window (like a text editor.) The task switcher's behavior of switching to the last-viewed window typically does what the user wants in a quick motion, while switching workspaces often requires thinking for a moment to decide which direction you want to switch to, making it relatively inconvenient. The biggest weakness of the task switcher is when you have many groups of tasks, which means the last-used window isn't the one you want to use when switching tasks. Here, workspaces compliment the task switcher's functionality–if any only if the user is confident in switching between workspaces and managing them logically according to task. Automatically creating new workspaces for maximized windows ruins the excellent workflow synergy between the task switcher and the workspace switcher.

#### Recommendation

Leave a maximized window in its current workspace.

When the user wants to move a maximized window to a new workspace, they should be able to do so easily–consider a specific keyboard shortcut for "Workspaces -> Move to new"

jlnr:
> Reasons in favor of moving maximized windows into their own workspaces
> The first that comes to mind is that macOS does this.

It only does this for _fullscreen_ apps, same as elementary OS. If you alt-click the green bubble in a Terminal window, the closest thing that exists to _maximization_ in macOS, the window stays on the same workspace.

I have no strong opinion either way. But if maximization now also creates a new workspace, isn't maximization close enough to fullscreen mode that we should merge both concepts? It seems to me that the biggest difference is #567, there is no way to reach header bars in fullscreen mode because they aren't brought into view when the mouse is at the top of the screen (like browsers and macOS do it).

> When the user wants to move a maximized window to a new workspace, they should be able to do so easily–consider a specific keyboard shortcut for "Workspaces -> Move to new"

If it's only "move to new", then it's not immediately clear if the window should be maximized on the new workspace or not.

ryanprior:
>But if maximization now also creates a new workspace, isn't maximization close enough to fullscreen mode that we should merge both concepts?

That's interesting. It could make sense, in that by choosing to maximize a window the user has communicated that they want it to take up as much space as possible. Whatever we can do to give a maximized window more space, we probably ought to do, within limits: if maximizing a window were to make the top bar disappear, that might be disorienting.

Where this leads my mind is that only certain apps should have a "full screen" state- games, movies, etc come to mind. Apps which really don't intend to cooperate with any other app, where if you're in that app you probably intend to stay there until you quit. Any app that permits cooperation should probably retain some window decoration.

>If it's only "move to new", then it's not immediately clear if the window should be maximized on the new workspace or not.

We have a keyboard shortcut for "Windows -> Toggle maximized" so if the user intends to maximize the window and move it to a new workspace, hopefully that would be easy to accomplish with two keystrokes in succession, similar to how maximizing a window and moving it to the workspace on the right is already easy to accomplish.

The difference would be the creation of a new workspace immediately to the right to receive the focused window, rather than moving the window to the workspace that's already to the right. In the case that the workspace to the right is already the "new workspace," the function would be identical.

ryanprior:
Some frames I use to understand the implications of this pull request are: implicit vs explicit, confidence-inspiring vs confidence-undermining, surprising vs expected, synergistic vs competitive.

So to sum up my view in terms of those frames: maximized-windows-as-workspaces takes the user's instruction to maximize a window and reads into that an implied instruction to create a new workspace and move the window there, rather than waiting for the user to explicitly do that. This can undermine the user's confidence in the workspace system, because until the user understands all the implications of what actions create workspaces and what ones don't, it may seem that workspaces are arbitrarily created and deleted.

Only once the user understands the logic behind it, which is rather surprising at first, will they regain their footing and be able to use the workspace interface confidently. Even then, with a user who understands what they're doing, the maximized-windows-as-workspaces model pits workspaces against the window switcher as it moves important windows out into their own workspaces where the window switcher does nothing, rather than exploiting the natural synergy between workspaces and the window switcher when each workspace contains semantically-related windows.

cassidyjames:
I'll need to play with this a lot more and see how it feels, especially across multiple devices setups: a laptop, a laptop with an external display, a single-display desktop, multi-display desktop, etc.

But my instinct is that for such a large, behavior-changing feature like this we need to make it an optional, off-by-default feature for now and let adventurous users try it out first. Based on their feedback and more extensive dogfooding, we can decide if we want it on by default in future releases.

Blast-City:
> I'll need to play with this a lot more and see how it feels, especially across multiple devices setups: a laptop, a laptop with an external display, a single-display desktop, multi-display desktop, etc.
> 
> But my instinct is that for such a large, behavior-changing feature like this we need to make it an optional, off-by-default feature for now and let adventurous users try it out first. Based on their feedback and more extensive dogfooding, we can decide if we want it on by default in future releases.

What if it was an "Isolate Windows" gsetting that had the following range:

- Disabled
- Fullscreen Windows
- Maximized and Fullscreen Windows.

This could also be a nice way to test #231

andreasfelix:
Just wanted to mention that in the current implementation, maximizing a window on the secondary screen switches the workspace on the primary screen. Also this change would imo introduce a slight inconsistency for multiple monitor setups as there are no workspaces for secondary screens.

After using this branch for about an hour I felt a bit confused and also (as @ryanprior described) undermined. I think there are workflows where it is reasonable to have multiple maximized windows on the same workspace. I think of workspaces as a logical container for multiple windows. When each window gets assigned to its own workspace it imo kind of defeats the purpose of workspaces as the workspace overview just becomes a dock.

But I also think that there are workflows where this behavior could beneficial so I would prefer to see this an explicit action. Maybe Alt+click-on-maximize or Alt+Drag-to-top??

donadigo:
@cassidyjames GSetting added (it's off by default).

crocodile2u:
That fullscreen windows are moved to a new workspace without me giving the order for this, is exactly why I stopped using this feature, although I would very much like to continue using it. I dont like multiple workspaces, I tend to keep the number of open apps small and I want to be able to Alt-Tab between **all** of them. Fullscreen windows just make me sad and frustrated, I always forget that this is a special window and I have to use a different shortcut if I want to switch apps. Please leave an option to **not** open extra workspaces unless the user has requested this (something like a checkbox in settings "Move fullscreen/maximized windows to a separate workspace" which can be checked by default)

ryanprior:
I also notice I've been avoiding ever using full-screen windows since this merged.

I do like & make use of multiple desktops but I'm very particular about what windows I put on each one, and I use numbered shortcuts (eg <kbd>⌘</kbd><kbd>2</kbd>) to jump to a specific group of windows. Since this merged, whenever I full-screen a window it messes up my layout in two ways:
* numbered shortcuts for desktops to the right of the full-screened window are now offset by one, which means I'll repeatedly jump to the wrong desktop
* the window I full-screened is now removed from the alt-tab pool it was in, leading me to look for it in the window switcher only to remember a second later that it's on another desktop now

My workaround has been to use maximized windows for all use cases where I used to use full-screen, except for true single-screen apps like Videos where I'm not interacting with a workflow. I know I can go full-screen and then use <kbd>⌘</kbd><kbd>Alt</kbd><kbd>←</kbd> to move the window back, but I haven't got in the habit of doing that and there's no sensation of window motion to reassure me that the move happened, so I haven't been comfortable with that as a workaround.

crocodile2u:
By the way.. With this feature merged, it means that maximized windows will now behave much like fullscreen, right? And there's no way to omit this behavior. I don't want to sound offensive but that's simply going to make me switch OS. I honestly tried to get used to fullscreen behavior in elementary, and I just can't get along with it. I deeply appreciate the efforts in this otherwise great software, but things like this - I use them **all the time** and if I can't get a behavior that feels natural to me, I'm gonna stick to UI/UX inconsistencies of some other distro which does not irritate me every other minute :-(

BTW, [this is a good read on the subject](https://www.reddit.com/r/Ubuntu/comments/ef1ck4/mutiny_in_pantheon_ubuntu_mate_gala_plank/) and I agree to each of the guy's requirements, and it makes me feel like I'm not alone.

I would also be keen to help implement the new setting and keep the window where it belongs (downside being that I have zero experience with Vala in which I guess the WM is written or the Gala development).

donadigo:
@crocodile2u This behaviour is completely optional right now and will be for foreseeable future. If you didn't touch any settings, nothing should change. I don't think we have any plans to make this behaviour locked-in without a setting.

crocodile2u:
Yeah, thanks for the heads-up! I already took a look at the PR and found out that there is a setting. One small question: it doesn't seem like it would be a big deal to implement the same kind of setting for fullscreen windows (although there might be certain hidden pitfalls). In fact, I myself feel like investing some time in this, maybe I'll be able to show some result in a week or two.

Blast-City:
Anyone who wants Fullscreen workspaces to be optional should give a thumbs up (or a heart if you feel generous) to #231

PR #579, opened 2019-07-02
Drag the window down to open workspace view

### LINKED ISSUE #569: Drag the window down to move it to a workspace
[Blockbuster](https://github.com/donadigo/blockbuster) had a feature (removed due to crashing gala on some systems) to drag the window down to the bottom of the screen to open the workspace view with the window being already moved. This allowed to move the window quickly to another workspace.

donadigo:
Fixes #569.

Enables dragging the window down to the bottom screen edge to open workspace view and immediately move the window elsewhere.

Currently based on screen percentage (more than 98% of the height), we may rethink if this should happen.

### DIFF
--- a/src/WindowMovementTracker.vala
+++ b/src/WindowMovementTracker.vala
@@ -1,0 +1,110 @@
+//
+//  Copyright (C) 2019 Adam Bieńkowski
+//
+//  This program is free software: you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License as published by
+//  the Free Software Foundation, either version 3 of the License, or
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//
+
+using Meta;
+
+namespace Gala
+{
+	public class WindowMovementTracker : Object {
+		public weak Meta.Display display { get; construct; }
+		public signal void open (Meta.Window window, int x, int y);
+		private Meta.Window? current_window;
+		private const float TRIGGER_RATIO = 0.98f;
+
+		private float start_x;
+		private float start_y;
+		private Meta.MaximizeFlags maximize_flags;
+
+		public WindowMovementTracker (Meta.Display display)
+		{
+			Object (display: display);
+		}
+
+		public void watch ()
+		{
+			display.grab_op_begin.connect (on_grab_op_begin);
+			display.grab_op_end.connect (on_grab_op_end);
+		}
+
+		public void unwatch () 
+		{
+			display.grab_op_begin.disconnect (on_grab_op_begin);
+			display.grab_op_end.disconnect (on_grab_op_end); 
+			
+			if (current_window != null) {
+				current_window.position_changed.disconnect (on_position_changed);
+			}
+		}
+
+		public void restore_window_state ()
+		{
+			var actor = (Meta.WindowActor)current_window.get_compositor_private ();
+			current_window.move_frame (false, (int)start_x, (int)start_y);
+			if (maximize_flags != 0) {
+				unowned AnimationSettings settings = AnimationSettings.get_default ();
+				int previous = settings.snap_duration;
+				settings.snap_duration = 0;
+				current_window.maximize (maximize_flags);
+				settings.snap_duration = previous;
+
+				/**
+				 * kill_window_effects does not reset the translation
+				 * and that's the only thing we want to do
+				 */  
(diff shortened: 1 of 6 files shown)


danrabbit:
This is a cool idea and it works smoothly. I'm into it.

Something brought up in slack was making sure that we only do this for bottom displays in a vertical multi-display setup.

the percentage works fine for me on a 13" display, but maybe we should get feedback from someone with a rather large display to make sure that's still sane

worldofpeace:
How's this going to fit in when Pantheon is gesture aware?

jlnr:
@donadigo I've rebased this on master, but besides the git conflicts, I've had to change the handling of `AnimationSettings`. Also, `start_motion` returns a bool now because the extracted branch on master has started to `return false` when `handle == null`. I've compared the diff before and after rebasing, but please also double-check that the animation changes make sense :pray: 

As for the percentage, it feels good to me on a large 3840x1600px screen :+1:

src/WindowMovementTracker.vala
```diff
@@ -0,0 +1,106 @@
...
+			unowned Meta.Screen screen = window.get_screen ();
+			unowned Meta.CursorTracker ct = Meta.CursorTracker.get_for_screen (screen);
+			int x, y;
+			Clutter.ModifierType type;
+			ct.get_pointer (out x, out y, out type);
+
+			int height;
+			screen.get_size (null, out height);
+
+			if (y > (float)height * TRIGGER_RATIO) {
```
andreasfelix: - Does currently not work for multiple monitors if monitors have not the same height or are not aligned. Instead of using the `screen ` height I think it should be better to use the height of the current monitor. This should fix the Problem:

```vala
Meta.Rectangle wa = window.get_work_area_for_monitor (screen.get_current_monitor ());

if (y - wa.y > wa.height * TRIGGER_RATIO) {
  window.position_changed.disconnect (on_position_changed);
  open (window, x, y);
}
```
- Also wouldn't it be reasonable for the `TRIGGER_RATIO` to take the current scaling factor into account, so that it would be larger on HiDPI?

andreasfelix:
To check if there is no monitor below in a vertical multi monitor setup you could use the code below.
I have implement this and changes from the comment above in https://github.com/andreasfelix/gala/commit/c5709005ecce3b7baba5e391ea11bfc34b8e9e9c. Feel free to merge them if you agree.

```vala
bool has_monitor_below (Meta.Window window) {
  unowned Meta.Screen screen = window.get_screen ();
  var n_monitors = screen.get_n_monitors ();
  var current_monitor = screen.get_current_monitor ();
  var current_rect = window.get_work_area_for_monitor (current_monitor);
  var current_end_x = current_rect.x + current_rect.width - 1;
  var current_end_y = current_rect.y + current_rect.height - 1;
  for (int i = 0; i < n_monitors; i++) {
    if (i == current_monitor) {
      continue;
    }

    var other_rect = window.get_work_area_for_monitor (i);
    if (current_end_y < other_rect.y) {
        if (current_rect.x <= other_rect.x + other_rect.width - 1 && current_end_x >= other_rect.x) {
            return true;
        }
    }
  }

  return false;
}
```

donadigo:
@andreasfelix do I understand correctly that your code disables the behaviour completely when the monitors are stacked up vertically? If that's the case then it's not what we really want. Ideally, we would like to have always the most bottom edge of the monitor setup to be the trigger instead of disabling the feature completely.

andreasfelix:
@donadigo 
> We would like to have always the most bottom edge of the monitor setup to be the trigger instead of disabling the feature completely.

This is what I was trying to achieve: 

1. Currently you are using the `screen` height to test if the cursor is at the bottom on the screen. This logic fails for a horizontal monitor setup, if the bottom of the monitors are not algined. I think, therefore it would be reasonable to use the `monitor` height instead. See https://github.com/andreasfelix/gala/commit/88c5b7ef18acc860bb709f8c61b376ccaf2f848d

2. But using the `monitor` height, the workspace overview gets also triggered when trying to move a window from the upper monitor to the bottom monitor in a vertical setup. To prevent the drag-to-bottom-action should not be triggered for all monitors which have a monitor below. See https://github.com/andreasfelix/gala/commit/c5709005ecce3b7baba5e391ea11bfc34b8e9e9c

The function `has_monitor_below` only evaluates to `true` if the current monitor has a monitor below not if any monitor has a monitor below. This is not so easy to test because of #7. You have to horizontally offset the windows so that the wingpanel struts do not block the area or install the branch with the fix.

donadigo:
@danrabbit resolved conflicts and merged the multi-monitor fixes from @andreasfelix. Ready for review again.

andreasfelix:
Another multi-monitor issue occurs when dragging a window from a secondary monitor to the bottom.
The workspace overview gets triggered, but the window does not transform into the icon and instead moves back to its initial monitor.

issue

donadigo:
@andreasfelix unfortunately I won't be able to solve these issues by myself. I do not have the setup to properly test these cases so there's little here to do for me.

danrabbit:
Converting to draft since there are conflicts and there hasn't been any movement here for a while. Feel free to mark ready for review if this is updated :)

PR #581, opened 2019-07-12
Add issue templates

peteruithoven:
First example of https://github.com/elementary/os/issues/209

Goal is to improve the quality of issues, for example making sure people include all the relevant information from the get go. 

Inspired by:
- Github's default issue templates
- @kgrubb proposal
- [Signal's issue template](https://github.com/signalapp/Signal-Desktop/issues/new)

Tried to include as many existing resources as possible. 

I'm hoping to copy these templates to almost all repo's with minor customizations.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -2,7 +2,7 @@
 name: Bug report
 about: Create a report to help us improve
 title: ''
-labels: bug
+labels: ''
 assignees: ''
 
 ---
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -46,12 +46,12 @@ If applicable, add screenshots or screen recordings to help explain your problem
 
 ### Logs
 <!--
-If applicable, share logs.
+If applicable, please share logs.
 
-If you notice erroneous behavior share logs, how:
+Notice erroneous behavior? Get and share debug logs by following:
 https://elementary.io/docs/code/os-dev#debug-logs
 
-For example when an application crashes please share crash logs, how:
+The application crashes? Get and share crash logs by following:
 https://elementary.io/docs/code/os-dev#inspecting-crashes
 -->
 
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,72 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: bug
+assignees: ''
+
+---
+
+<!--
+By filing an Issue, you are expected to comply with the elementary code of conduct: https://elementary.io/code-of-conduct
+
+Please fill out this template with all the information you have. We can't do much without a detailed description of what you've encountered. Please do your best!
+
+Please note that this tracker is only for bugs and feature requests. Please try these locations if you have a question or comment:
+
+  https://elementaryos.stackexchange.com/
+  https://www.reddit.com/r/elementaryos/
+
+Please read and follow these tips:
+https://elementary.io/docs/code/reference#be-prepared-to-provide-more-information
+
+Lastly, be sure to preview your issue before saving. Thanks!
+-->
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,72 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: bug
```
danrabbit: I don't think we want to use this label since everything in the issue tracker that a user reports can be described as a bug

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,72 @@
...
+  https://www.reddit.com/r/elementaryos/
+
+Please read and follow these tips:
+https://elementary.io/docs/code/reference#be-prepared-to-provide-more-information
+
+Lastly, be sure to preview your issue before saving. Thanks!
+-->
+
+## Prerequisites
+- [ ] I have searched open and closed issues for duplicates.
```
danrabbit: Do we need this anymore since GitHub includes search when creating a title now?

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,43 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+title: ''
+labels: enhancement
```
danrabbit: We use this label for feature requests.

```suggestion
labels: Priority: Wishlist
```

I wonder if we should also add the "needs design" label here by default?

cassidyjames: @danrabbit yeah it makes sense that a feature request/wishlist item would require design input.

.github/ISSUE_TEMPLATE/bug_report.md
cassidyjames: I don't think we need it. I know some third-party clients might not search for issues ahead of time (i.e. an Android app or whatever) but that doesn't really seem like what we'd be catering to with issue templates anyway.

.github/ISSUE_TEMPLATE/bug_report.md
cassidyjames: Would it be evil to use the Incomplete label until a maintainer/somebody comes in and properly confirms/labels it?

.github/ISSUE_TEMPLATE/bug_report.md
peteruithoven: I would mainly include it to make people think twice. It kind of reiterate that people should really do this. 

Also these relates issues list only looks at the title and as we all know there are usually many ways to word things. When searching for existing issues I usually try 2 or 3 terms. 

In some cases it's useful to search through all the issues of the elementary org. For example when people open an issue in the nightlight indicator repo which is a duplicate of something in the nightlight settings repo. But I'm not sure how to explain that and if that applies often enough.

.github/ISSUE_TEMPLATE/bug_report.md
peteruithoven: We'll except for feature requests... 
But yeah bug is used way more often so I get why that might just add noise.

jlnr:
What do you think about linking to [Screen Recorder](https://appcenter.elementary.io/com.github.mohelm97.screenrecorder/) right from the template? It's a bit unfair because there several screen recording and GIF conversion apps on AppCenter, but the alternative is that users have to app hunting just to report a bug.

peteruithoven:
I've improved the default labels. Added `Needs design` to Feature requests and removed `bug` from bug report.

Even though I love Screen recorder it seems weird to recommend a specific non core application.

cassidyjames:
@peteruithoven it sounds like we can add these to the new `.github` repo, actually, and have them be the default for all repos. We should move them there, and then yeah I'm fine starting here since it'll be easier to update.

danrabbit:
Moved to https://github.com/elementary/.github/pull/1

peteruithoven:
Thanks for picking this up! 
It's awesome that github supports these org wide settings.

Issue #582, opened 2019-07-13
No animation for tiled windows with tiled match


andreasfelix:
![animation4](https://user-images.githubusercontent.com/24651767/61176486-035a5180-a5c2-11e9-948b-b3ee397e511c.gif)
*(slowed down animation)*

It seems like there is no animation when tiling an already tiled window on top of another tiled window. This happens because of line 969:

https://github.com/elementary/gala/blob/a790d2d0f80be8c12a5afe2534ea17bd825f3a63/src/WindowManager.vala#L969-L975

Even I was the last one how edited this line, I just copied it from the previous code. The condition `window.get_tile_match () == null` introduced in #239. 

@donadigo was the idea of introducing this condition? I assume you wanted rule out animations when resizing of two tiled windows?

If this is the case, I think it could be simply fixed by using this condition only on the `size_changed` signal:
```vala
} else  { 
  ulong size_signal_id = 0U;
  ulong position_signal_id = 0U;
  if (window.get_tile_match () == null) { // don't animate resizing of two tiled windows
    size_signal_id = window.size_changed.connect (() => window_change_complete (actor, which_change, size_signal_id, position_signal_id));
  }
  position_signal_id = window.position_changed.connect (() => window_change_complete (actor, which_change, size_signal_id, position_signal_id));
  return; // must wait for size/position-changed-signal to get updated rect_frame
}
```

donadigo:
@andreasfelix to be honest with you, I really don't know why I introduced this there. I predict that it was something wrong with how the animations worked when this wasn't ruled out. Either way there was some issue that I can't remember. 

I'm not sure if this happens still today with the current mutter version and I think it's worth a try.

andreasfelix:
@donadigo, could it maybe be better to do this like GNOME shell
https://gitlab.gnome.org/GNOME/gnome-shell/blob/master/js/ui/windowManager.js#L1468

Instead of using callbacks for different signals of the individual windows, they use the "size-change" signal of the WM to prepare the animation and than play the animation when a "size-changed" signal of the WM is emitted. Also at the beginning of the `size_changed` method they rule out all resizing events, which is probably what you wanted to achieve with `if (window.get_tile_match () != null) return;` condition.

I tested to log the rectangle that bounds window in the `size_changed` method:
```vala
public override void size_changed (Meta.WindowActor actor) {
  unowned Meta.Window window = actor.get_meta_window ();
  var r = window.get_frame_rect ();
[...]

PR #583, opened 2019-07-13
Fix: Animations for two tiled windows

andreasfelix:
Fixes: #582

There was no animation when moving a tiled window on top of another tiled window or when maximizing one of two tiled windows ...

**Before** (slowed down animations)
ani_bug

**After** (slowed down animations)
ani_fixed

### DIFF
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -962,14 +962,30 @@ namespace Gala
 
 		public override void size_change (Meta.WindowActor actor, Meta.SizeChange which_change, Meta.Rectangle old_frame_rect, Meta.Rectangle old_buffer_rect)
 		{
+			debug("size_change");
 			unowned Meta.Window window = actor.get_meta_window ();
+			switch (which_change) {
+				case Meta.SizeChange.MAXIMIZE:
+					debug("maximize");
+					break;
+				case Meta.SizeChange.UNMAXIMIZE:
+					debug("minimize");
+					break;
+				case Meta.SizeChange.FULLSCREEN:
+				case Meta.SizeChange.UNFULLSCREEN:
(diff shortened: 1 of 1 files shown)


andreasfelix:
Does currently introduce other glitches! Has to be changed!

andreasfelix:
The problem is that in the event of resizing two tiled windows the value of `which_change` is `Meta.SizeChange.MAXIMIZE`. Therefore (in the previous commits) the maximize-animation is also played back when resizing two tiled windows, which lead to the above mentioned glitch.

In Master this is currently ruled out by:
https://github.com/elementary/gala/blob/50694796d4c8f0ca92517d5a628b0efdf748279c/src/WindowManager.vala#L979

But this does also disable the following window animations for two tiled windows:

1. Maximizing one of two tiled windows
2. Moving one of two tiled window (see GIFs above)
3. Unmaximizing one of two tiled windows

A solution would be to rule out all resizing events, but I did not find a way to recognize them.

---

To a least enable the unmaxizing animations (3.) one can use the condition `window.get_tile_match () == null` only for maximizing events. This is what I did in the last commit.

Additionally I removed the callbacks of the position/size-changed signals of the window and instead used the `size_changed` signal of the WM similar to how [GNOME](https://gitlab.gnome.org/GNOME/gnome-shell/blob/master/js/ui/windowManager.js#L1470) does it.

andreasfelix:
I think I now found a proper fix. To detect if one of two windows was just resized and not moved one can use the position of its left and right edges: If one of these position does not change and the window was not maximized horizontally one can assume that the window as just resized. Using this condition one can rule out the resizing events, which enables the animation for the cases 1-3.

Issue #536, opened 2019-05-14
Enable "Alt" + "~" to switch between windows of the same application


uncertainquark:
Right now, there's no quick way to switch between multiple windows of the same application. In most operating systems, one does that with `Alt` + `~`. It seems like the feature has been left amiss in elementary.

Blast-City:
I like this idea, but what would happen if the windows from the same app were spread through multiple workspaces?

I guess this would only be possible if there was a window switcher revamp. #72

janxkoci:
IMHO the switcher should stick to the current behaviour of showing only windows from current workspace. If you change it to show windows from all workspaces, you will:

1. surprise people with unexpected behaviour
2. break workflows of people used to current behavior
3. make workspaces less useful (lot of people use them to prevent endless alt-tabbing)

Blast-City:
> IMHO the switcher should stick to the current behaviour of showing only windows from current workspace. If you change it to show windows from all workspaces, you will:
> 
>     1. surprise people with unexpected behaviour
> 
>     2. break workflows of people used to current behavior
> 
>     3. make workspaces less useful (lot of people use them to prevent endless alt-tabbing)

Yeah, i agree.

Maybe limit Alt+Key above tab to all the windows of the app that are present on the current workspace.

The same could be also applied to middle scroll (or in the future left clicking when multiple windows are open) on the app dock icon. Leaving right click to select if you want a window on other workspaces. Maybe an indicator showing which windows correspond to which workspace.

peteruithoven:
On OS X this seems to be done using <kbd>Super</kbd> + <kbd>`</kbd>:
https://apple.stackexchange.com/questions/193937/shortcut-for-toggling-between-different-windows-of-same-app
Which on my Dell XPS 15 would be easier since the <kbd>~</kbd> would requite pressing <kbd>Shift</kbd> as well.

agrawalsourav98:
I used to do this on ubuntu and i guess all windows of an application should just be limited to the workspace.
Is there some progress on this issue?

techdev5521:
I'm visually impaired and I'd also like to see this implemented. I use the <kbd>Super</kbd> + <kbd>`</kbd> in macOS to switch between windows of the same app because it is visually easier than either:

- Looking for the application icon(s) and guessing which is the window I want in the all windows view,
- Looking for the correct window preview in the all window view.

I'm working on my first elementaryOS application right now and switching between Code windows is proving to be a pain without this feature.

[...]

PR #585, opened 2019-07-18
Introduce shortcuts to focus/open applications

andreasfelix:
Fixes https://github.com/elementary/terminal/issues/319

**TL;DR**: A simple plugin to bind keyboard shortcuts to open/focus specific applications.

> ⚠️ Disclaimer:
>Remove this plugin when switching back to master:
> ``` rm /usr/lib/x86_64-linux-gnu/gala/plugins/libgala-appshortcuts.so ```
>Otherwise you cannot login again!

## Motivation
With increasing number of different windows (often multiple instances of the same application), I always feel (in all major desktops, not only elementary) not being able to efficiently switch focus between them anymore.
As 95% of the time I only use 5 different applications, I always wanted a shortcut to directly switch focus to one of them. *Alt+Esc* and *Alt+Tab* are great for the most recent applications but get tedious for applications more than 3 key presses away.

The idea is to introduce the possibility to bind a keyboard shortcut to a single application. Pressing this shortcut would focus this application. Pressing it multiple times would toggle focus between the different windows of this application. In other words: an *application-specific Alt-Tab* shortcut.

## Behavior
The *Application* shortcut behaves in that way that if ..

- **Case 1:** ... the application is not running, it opens a new instance.
- **Case 2:** ... the application is running but is not focused, it focuses the instance with highest z-index.
- **Case 3:** ... the application is running and has focus, it focuses another instance (lowest z-index).

The most important applications (like files, terminal, browser, ...) get all bond to a combination of Super+Key.

Additionally *Super+Ctrl+Key* will always open a new window of the application regardless if another instance is already running. (not implemented yet)

In this first implementation the default applications are bind to the following shortcuts:
- Text **E**ditor - **Super+E**
- File **M**anger - **Super+M**
- Web **B**rowser - **Super+B**
- **T**erminal - **Super+T**

## Productivity/Workflow
This feature would imo increase the productivity for different scenarios:

-  Quickly show/hide a browser or terminal to do a simple web search or to quickly execute a command.
- Toggling focus between two windows of a text-editor or terminal.
- Navigating between many different windows especially when dealing with multiple workspaces.

Also I think this feature would not only be useful for a keyboard-driven workflows as it also can reduce the mouse movement in some cases. For example, if the mouse is on top of the screen and one wants to do a web search, one has to move the cursor down to plank and than again up to click the search bar. With the *Application-Shortcut* one has to press the shortcut and then move the cursor only a short distance to the search bar. 

I also recorded a simple [Demo video](https://www.youtube.com/watch?v=0ES5pTrbQ3w) to showcase the workflow for these cases.

## Consistency
As it is already possible to open some applications with a shortcut, this feature would simple extend this functionality by the focusing/toggling component. The current behavior which always opens a new window would get bound to *Super+Ctrl+Key*.

This feature would also be consistent with the current behavior of plank, where a left-click opens a new window when the application is not running and focuses the highest window if the application is running. Also in plank *Ctrl-Click* opens a new instance of the application regardless if another instance is already running and the toggling behavior between windows of the same application is achieved by scrolling over an item.

This feature could made be more consistent to *Alt+Tab* by allowing *Super+Shift+Key* to toggle backwards.

## This PR
This PR is only a first implementation to demonstrate this feature. It is not ready for review, but suggestions are welcome! 

If the idea gets accepted I would additionally implement, *Shift* to toggle backwards, *Ctrl* to always open a new window and maybe make the code a bit more efficient.


jlnr:
I agree with the use case and I like the idea of toggling through windows. But in terms of shortcuts, I prefer the way this is handled in Ubuntu (I think starting in Unity?): the first dock icon is always cmd+1, the second is cmd+2, etc. That way would be workflow-agnostic and even easier to customize. And you can assist the user by showing tiny numbers in the dock while cmd is held down.

cmd+n is already bound to workspace switching in elementary; maybe cmd+F1, cmd+F2, ... would work?

andreasfelix:
Thanks for the feedback! Would it be crazy to allow both? *Super+Key* for the favorite applications and *Super+FKey* for the current applications. 

Some reasons why I think the numerical *Super+FKey* approach alone would not be enough:
- not able launch applications which are not in the dock
- if the dock is not visible *Super+F3* is harder to remember than *Super+T*
- if you reorder your dock, you will loose the muscle memory

The idea with the tiny indicators sounds nice, but I could imagine it could be irritating when pressing other shortcuts. Maybe one could implement these number-indicators in the *Alt-Tab* view or *window overview*?

Could you maybe explain why you think the numerical approach would be easier to customize? All the keyboard shortcuts would also be customize-able through settings. Also the shortcuts already use the current default applications. You could even change these to manually *Super+FKey* so the will visually match order of the dock... 

Also the 
- Text **E**ditor - **Super+E**
- File **M**anger - **Super+M**
- Web **B**rowser - **Super+B**
- **T**erminal - **Super+T**

shortcuts were just for demonstration purposes. I personally wouldn't mind to change them (or even deactivated them by default) as long as I can change them through settings.

jlnr:
My default editor (Code) is not an app I jump to very often, instead I'd love to have shortcuts for Android Studio and gitg. With the "dock position" approach, I only need to make sure they are among the first twelve dock icons, and my shortcut setup is done. That's the sort of "customization" I was thinking of.

But sorry, you are of course right that we could totally have both. I'll open a separate ticket after some more consideration.

By the way, what do you think about adding case 4: when the app only has one window and it is already focused, then pressing the shortcut again hides the window -- similar to clicking a Plank icon twice. That could be great to bring up an app and dismiss it with minimal finger movements, like a schedule or Slack.

andreasfelix:
> I don't think that everyone lives in the same set of apps. I don't jump to my default editor (Code) very often, instead I'd love to have shortcuts for Android Studio and gitg. With the "dock position" approach, I only need to make sure they are among the first twelve dock icons, and my shortcut setup is done. That's the sort of "customization" I was thinking of.

That is a good point, I will additionally to the default applications introduce an option to bind this shortcut directly to a given `.desktop` file. But this will also need a s

> By the way, what do you think about adding case 4: when the app only has one window and it is already focused, then pressing the shortcut again hides the window -- similar to clicking a Plank icon twice. That could be great to bring up an app and dismiss it with minimal finger movements, just to check a log in the Terminal etc.

This would not be difficult to implement, but a fourth case could also make this feature more confusing for a normal user. BTW, You can already achieve the same behavior by focusing the application with `Super+key` and than switch back with `Alt+Esc`. But I am open to implement that ...

andreasfelix:
@jlnr I have now added a possibility to bind custom applications and I also did a first implementation for the settings-keyboard plug. It would be nice to get some feedback!

When switching back to master don't forget to remove the plugin:
```
rm /usr/lib/x86_64-linux-gnu/gala/plugins/libgala-appshortcuts.so
```

applications-menu
 to

Serkan80:
I also have the urgent need for a shortcut to switch between windows of the same app. I have the same frustrations as the topic opener. 

If Im not wrong, you can do this in OSX by pressing super+’. 

Anyways lets not discuss which keycombinations we should use for this functionallity. Just take a sensible default and give the option to change it in the settings.

andreasfelix:
Last commit also fixes #536

andreasfelix:
similar to https://github.com/mkropat/jumpapp

Issue #591, opened 2019-08-06
Mouse Quarter Tiling / Ctrl-Drag-Tile


andreasfelix:
I think it would be a useful addition if **Ctrl-Drag** could be used to tile windows. I did a [first implementation](https://github.com/andreasfelix/gala/commit/16be4b9673a9db784c59a7a7f4600696fce42600) which is based on https://github.com/elementary/gala/pull/579.

The ctrl-tile-left/right action is especially useful for multi-monitor setups where the edge-tiling can get a bit inconvenient.

A problem is that [Mutter does not expose an API for tiling](https://gitlab.gnome.org/GNOME/mutter/issues/657), so the windows are not "really" tiled (compared to the edge tiling).

If the idea gets accepted, I could do a PR.  

![mouse_quarter_tiling](https://user-images.githubusercontent.com/24651767/62541776-bf720980-b85a-11e9-830c-430c12cec990.gif)
*Ctrl is pressed while dragging the window*

![tile_sketch](https://user-images.githubusercontent.com/24651767/62542523-36f46880-b85c-11e9-9de0-a91db6876e17.png)

donadigo:
I personally think that tile top shouldn't be the default action for dragging to the top as this action is much less desirable than actually maximzing the window.
From the above GIF I conclude that your tiling works based on the window position. I suggest that it should work on mouse position and also make the margin for triggering the tile much lower, it should not cause accidental tiles.

Anyway, I think we would definitely merge it if it worked well.

andreasfelix:
@donadigo The tiling is based on the mouse position. Maybe I should be more specific, but this does not aim to improve/replace the edge tiling. Instead when pressing down **Ctrl** while dragging a window, it gets resized to the area where the cursor is currently in. (see second picture).

cassidyjames:
Duplicate of #105; let's try to keep discussions on original issues/discussions instead of opening new ones. ;)

andreasfelix:
I opened a new issue, because I proposed a different tiling mechanism than in the mentioned issue: Instead of moving the window to an edge of the screen, the window is tiled corresponding to the area which the mouse pointer is currently in.

Anyway for the current implementation see #963  :)

cassidyjames:
@andreasfelix I understand, but the underlying "want" is the same: for quarter tiling. It's easier to track if the discussion is in one place instead of spread across several different issues.

PR #595, opened 2019-08-11
Introduce a Tiling-Mode

andreasfelix:
Fixes: #591

I think it would be a useful addition if <kbd>Ctrl</kbd><kbd>Drag</kbd> could be used to tile windows. I did a implementation which is based on https://github.com/elementary/gala/pull/579.

**Behavior:**
The screen gets divided into 9 different areas which all correspond to a different tile-action (see second picture). Dragging a window with <kbd>Ctrl</kbd> pressed to one of these areas will activate the tile preview (blue rectangle). Releasing the  <kbd>Left-Mouse-Button</kbd> will tile the window corresponding to area where the mouse pointer is currently in.

**Use-cases:**
- This PR will enable quarter and vertical-half tiling (which is super useful on vertical monitors).
- The ctrl-tile-left/right action is also useful for horizontal multi-monitor setups where the edge-tiling can get a bit inconvenient.

A problem is that [Mutter does not expose an API for tiling](https://gitlab.gnome.org/GNOME/mutter/issues/657), so the windows are not "really" tiled (compared to the edge tiling).

ctrl-drag-2
*Ctrl is pressed while dragging the window*

tile_sketch


Serkan80:
It would also be cool if:

1) you could easily navigate with your keyboard between the tiled apps
2) when you tile an app, a second suggestion should appear to let you choose another app to tile your app against. 
In W10, when you tile an app to the left, then automatically on the right side of the screen a suggestion appears with miniature apps to choose from (when you press esc then the suggestion disappears).  This is especially handy for only half tiling. For quarter and other tilings this suggestion should not appear.

Serkan80:
It would also be cool if:

1) you could easily navigate with your keyboard between the tiled apps
2) when you tile an app, a second suggestion should appear to let you choose another app to tile your app against. 
In W10, when you tile an app to the left, then automatically on the right side of the screen a suggestion appears with miniature apps to choose from (when you press esc then the suggestion disappears).  This is especially handy for only half tiling. For quarter and other tilings this suggestion should not appear.

andreasfelix:
@Serkan80 this PR only targets mouse tiling. Maybe you should open separate issues for keyboard navigation and tile suggestions.

cassidyjames:
So I think the main issue is that this is not "real" tiling, i.e. the window manager and thus stylesheet has no way of knowing a window is considered tiled in this way. *If* we're fine with that (which it seems we'll have to be until there are some big tiling improvements to Mutter), then I would be open to pursuing tiling that uses the existing Super+Drag mouse shortcut as well as some keyboard shortcuts.

Some things to keep in mind with this sort of tiling: 
- What happens when a window is moved across to another display of a different size (or the resolution is changed)? Is it tiled relative to the new display, or does it retain its pixel-size? I believe the former is what Mutter attempts to do.
- How does this interact with Mutter's actual side-by-side tiling? Especially since Mutter has a concept of resizable tiling; I think you'd want this tiling to be aware of that and thus not always be split 50/50.
- Is this written in a way that will be maintainable, support both mouse and keyboard (and probably window-menu) tiling?

andreasfelix:
> What happens when a window is moved across to another display of a different size (or the resolution is changed)? Is it tiled relative to the new display, or does it retain its pixel-size? I believe the former is what Mutter attempts to do.

Yes, windows which are tiled via edge tiling will be tiled relative to the new display when moved via <kbd>Super</kbd><kbd>Shift</kbd><kbd>Right</kbd>:

edge_move_window

Windows which are (fake-)tiled via the new <kbd>Ctrl</kbd>-Drag feature will retain their pixel size when moved via <kbd>Super</kbd><kbd>Shift</kbd><kbd>Right</kbd>. But this is also an issue for normal untiled windows (see same video below):
ctrl_drag_move-window

> How does this interact with Mutter's actual side-by-side tiling? Especially since Mutter has a concept of resizable tiling; I think you'd want this tiling to be aware of that and thus not always be split 50/50.

This is currently not possible. [Mutter does not expose an API for tiling](https://gitlab.gnome.org/GNOME/mutter/issues/657), it only exposes the tiled match. The windows will always be tiled with 50%. I could implement, that the maximizing action would use the mutter API, so that a least maximized windows are "really" maximized.

> Is this written in a way that will be maintainable?

I think so. This branch is based on @donadigo's work of #579 . He implemented a `WindowMovementTracker` class which makes it able to hook into the window movement. I just implemented the tiling logic into this class. As soon as mutter exposes an API for tiling it should be pretty straight forward to replace the my custom implemented tiling logic (it is only 21 lines of code).

> Is this written in a way that will be support both mouse and keyboard (and probably window-menu) tiling?
I do not think that `WindowMovementTracker` should support keyboard tiling. At some point the tiling logic should be implemented in mutter. And this class will be just a user of the tiling API.

What is window-menu tiling?

cassidyjames:
I still have mixed feelings about this. I appreciate the work that has gone into it so far and feel like it's close, but fundamentally I think we need to be careful not to pour resources (and maintenance) into a somewhat hidden, non-default behavior. I would much rather arrive at a solution that we could just enable by default on normal window dragging instead of requiring a modifier. I also really want to think about keyboard tiling with this, as I think people will expect to be able to do all the same tiling from a keyboard as soon as we would release this.

Would this approach be possible: implement this on normal window drags. pushing against the sides or top of the display would half-tile or maximize using Mutter (as they do today). But pushing into the corners would corner-tile using this sort of fake tiling (positioning and sizing). Then we add keyboard shortcuts to tile windows into the corners as well.

It might be a much smaller step towards true tiling, but I think it would be more intercompatible with the existing tiling workflow, plus it seems to be closer to what i.e. Windows 10 is shipping. So there's at least a precedent there for that sort of corner vs side/maximized tiling on drag.

andreasfelix:
@cassidyjames, thanks for the feedback!

I would argue, that there is currently no widely accepted workflow for quarter-tiling, so all types of quarter-tiling are kind of non-default. On the other hand the edge-tiling feature is default in many Linux-Shells and Windows. I think, the strength of edge-tiling is that there is one distinct action for each edge. IMO, introducing 6 more actions (4 quarter + 2 vertical-half) would make edge-tiling less accessible (because one has to be more precise) and would break existing workflows. I have no numbers to back this, but I think there are many users which expect edge-tiling to work as it currently does and do not necessarily need quarter-tiling. I imagine it would be frustrating to them, if dragging a window near a corner would suddenly quarter-tile it.

IMO, for more complex tiling actions a area-based approach is less fiddly than dividing each edge into 3 parts. Another point is that edge-tiling is somewhat inconvenient on multi-monitor setups (because one has to precisely move the window to the edge of the monitor). This would get more tedious if drag-to-corner would be used for quarter-tiling.

<kbd>Ctrl-drag</kbd> and <kbd>ctrl-super-drag</kbd> are a super convenient actions and are currently unused. Furthermore they are consistent with the existing keyboard shortcuts <kbd>Ctrl</kbd><kbd>super</kbd><kbd>arrow-keys</kbd>. And in the future, one could add an alt-drag action to move windows between workspaces, which would be consistent with the existing <kbd>alt</kbd><kbd>super</kbd><kbd>arrow-keys</kbd> shortcuts.

Regarding your point of a hidden feature: What would you think if  - in addition to ctrl-drag tiling - holding the maximize button would also activate the area-based tiling-mechanism. (Maybe with a little animation which shrinks the window and adds a tiling icon to the cursor) This would be similar to the window-tiling in macOS, but which more than two tiling actions. If the edge-tiling would be kept the same, than elementary-tiling would be a windows-macOS-hybrid: Windows-tiling for the standard horizontal tiling and macOS-style for more complex tiling actions.

To make the feature more discover-able, it could be also promoted in Onboarding?

andreasfelix:
I have added a shrink animation.

ctrl-drag-2

andreasfelix:
Another super crazy idea. What if this would be the default drag action? Than dragging to edge could be used to move a window to a different work-space.

tort-dla-psa:
I love this feature! Any chances it will be in master someday?

cassidyjames:
@andreasfelix okay, a but more feedback: is it possible to show the window itself moving instead of shrinking down to an app icon? I think it would feel pretty jarring if the default dragging shrunk your app window down (basically making freeform movement impossible).

I think we do need to require pushing against (or at least being very near) display edges to trigger tiling, though. 

>I would argue, that there is currently no widely accepted workflow for quarter-tiling, so all types of quarter-tiling are kind of non-default. On the other hand the edge-tiling feature is default in many Linux-Shells and Windows. 

Windows 10 does have quarter-tiling out of the box, though, and it works well enough in my experience.

At this point, I wonder if I need to mock up my ideal tiling behavior graphically somehow to see if we can get Gala as close to that as possible—while obviously keeping the current Mutter restrictions in mind. If we do this, I want to make sure we do it right and don't end up shipping a half-thought-through feature that we end up changing significantly.

andreasfelix:
@cassidyjames 
1. How it looks with a shrinked window instead of an icon: (can be tested at 29057cc)
shrink-window
If this would become default behavior, freeform could be made accessible to a modifier key. I am aware that might be too radical. I just thought it was an interesting idea...

> I think we do need to require pushing against (or at least being very near) display edges to trigger tiling, though.

2. Why do you think so? I think an area-based approach needs less mouse movement and less precision. It also allows vertical tiling and it could be extended to any grid (e.g. 3x2 for ultra-wide monitors).

3. I wasn't aware windows has implemented this. I tested it, but I think the quarter-tiling is too fiddly on multiple monitors. And if I am not mistaken, it has no vertical tiling. 

IMO, the edge-tiling gets to complex, if one wants to implement vertical tiling. I feel like the area-based tiling is a more natural metaphor. And it isn't that "non-default", because area-based tiling is already available in some code editor (e.g. vscode): https://code.visualstudio.com/docs/getstarted/userinterface#_grid-editor-layout

I think something similar would be great for gala!

kmal-kenneth:
I have an ultra wide monitor. This functionality would be very useful ... I hope its integration.

andreasfelix:
I have added the possibility to divide the display into an arbitrary grid, which is useful for ultra-wide monitors. Currently the grid has a hardcoded 2x2 value, but it should not difficult to make this configurable via a gsettings entry.  

Here an example of an 3x2 grid:
area_tiling_other_grids

andreasfelix:
I refactored the code by moving the tiling logic into its own class.

aenrione:
I only wanted to pass by and say, Amazing work @andreasfelix ! I've been waiting for this functionality for a long time. Can't wait for it to get implemented, or just to try it for my self. Congrats!

cassidyjames:
@andreasfelix I'm getting build errors on hera:

```
../src/AreaTiling.vala:63.63-63.89: error: The name `get_current_monitor' does not exist in the context of `Meta.Display'
        Meta.Rectangle wa = window.get_work_area_for_monitor (display.get_current_monitor ());
                                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^
../src/AreaTiling.vala:71.13-71.51: error: expression type not allowed as initializer
            wa.x + col / 2 * monitor_width / grid_x,
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../src/AreaTiling.vala:72.13-72.52: error: expression type not allowed as initializer
            wa.y + row / 2 * monitor_height / grid_y,
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../src/AreaTiling.vala:73.13-73.50: error: expression type not allowed as initializer
            (1 + col % 2) * monitor_width / grid_x,
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../src/AreaTiling.vala:74.13-74.51: error: expression type not allowed as initializer
            (1 + row % 2) * monitor_height / grid_y,
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../src/AreaTiling.vala:51.50-51.76: error: The name `get_current_monitor' does not exist in the context of `Meta.Display'
        wm.show_tile_preview (window, tile_rect, display.get_current_monitor ());
                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
../src/WindowMovementTracker.vala:60.45-60.70: error: The name `get_cursor_tracker' does not exist in the context of `Meta.Display'
            unowned Meta.CursorTracker ct = display.get_cursor_tracker ();
                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^
../src/WindowMovementTracker.vala:32.44-32.59: error: Argument 1: Cannot convert from `void Gala.WindowMovementTracker.on_grab_op_begin (Meta.Display, Meta.Window)' to `void Meta.Display.grab_op_begin (Meta.Display, Meta.Screen, Meta.Window, Meta.GrabOp)'
            display.grab_op_begin.connect (on_grab_op_begin);
                                           ^^^^^^^^^^^^^^^^
../src/WindowMovementTracker.vala:50.49-50.74: error: The name `get_cursor_tracker' does not exist in the context of `Meta.Display'
                unowned Meta.CursorTracker ct = display.get_cursor_tracker ();
                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^
../src/WindowMovementTracker.vala:33.42-33.55: error: Argument 1: Cannot convert from `void Gala.WindowMovementTracker.on_grab_op_end (Meta.Display, Meta.Window)' to `void Meta.Display.grab_op_end (Meta.Display, Meta.Screen, Meta.Window, Meta.GrabOp)'
            display.grab_op_end.connect (on_grab_op_end);
                                         ^^^^^^^^^^^^^^
../src/WindowMovementTracker.vala:37.47-37.62: error: Argument 1: Cannot convert from `void Gala.WindowMovementTracker.on_grab_op_begin (Meta.Display, Meta.Window)' to `void Meta.Display.grab_op_begin (Meta.Display, Meta.Screen, Meta.Window, Meta.GrabOp)'
            display.grab_op_begin.disconnect (on_grab_op_begin);
                                              ^^^^^^^^^^^^^^^^
../src/WindowMovementTracker.vala:38.45-38.58: error: Argument 1: Cannot convert from `void Gala.WindowMovementTracker.on_grab_op_end (Meta.Display, Meta.Window)' to `void Meta.Display.grab_op_end (Meta.Display, Meta.Screen, Meta.Window, Meta.GrabOp)'
            display.grab_op_end.disconnect (on_grab_op_end);
                                            ^^^^^^^^^^^^^^
```

That aside, I still am uncertain about this. Collapsing windows to their app icon feels weird to me outside of the overiew; I think I'd still prefer moving the window "directly" (or a thumnail of the window at least), but I would be open to feedback here. I also think this needs to be a separate mode from normal window moving; I guess something like Ctrl+Super+Drag might make sense (since we already use Super+Drag for moving freeform and Shift+Super+Drag for snapping to edges). That way it's a separate opt-in tiling mode that bypasses Mutter's built-in tiling.

@donadigo have you had a chance to take a look at this code-wise?

andreasfelix:
@cassidyjames Is it required to built on 5.1? I thought, because you changed the CI to 6.0, it wouldn't be necessary to make it work on mutter-3.28. If it is necessary I can just add the compiler macros to make it work.

> That aside, I still am uncertain about this. Collapsing windows to their app icon feels weird to me outside of the overiew; I think I'd still prefer moving the window "directly" (or a thumnail of the window at least), but I would be open to feedback here. 

My motivation to shrink the window is to make sure that the window does not hide the tiling-preview. This was an issue in the first version. Also it gives the user a visual feedback that they entered the tiling-mode. But I am okay to change to whatever you want. I just tried to make the tiling mechanic more obvious.

>I also think this needs to be a separate mode from normal window moving; I guess something like Ctrl+Super+Drag might make sense (since we already use Super+Drag for moving freeform and Shift+Super+Drag for snapping to edges). That way it's a separate opt-in tiling mode that bypasses Mutter's built-in tiling.

What would you think about a desktop action similar to the window overview. Right click could be used to set the grid and a left-click would toggle an auto-tiling mode:

* If auto-tiling is enabled than windows will be tiled whenever they are moved. Free-form movement is accessible via <kbd>Ctrl</kbd>
* If auto-tiling is disabled than can be moved freely. Area-tiling is accessible via <kbd>Ctrl</kbd>

**Example icon:**


Screenshot from 2020-06-24 02 34 34

**Video: (10MB might take a second to load)**

toggle_auto_tiling-min

glimpster:
This is a really cool feature and it would look great on my dell xps 15 2020 !

danrabbit:
Hm I kind of feel like maybe this is getting to the point where we need to conduct a survey of users who already regularly use tiling window managers to find out what the expectations and conventions are and what pain points they have with existing solutions. These users probably already know what kind of grids they expect and how their window manager adapts to their display size.

Something to maybe consider is that for a lot of apps there will be a minimum size where that app is usable, so we may be able to infer what kind of tiling grid to offer based on display size instead of having the user select it manually.

Another alternative would be to have the grid dynamically add rows and columns as the user tries to tile more applications

alirezaly:
This is definitely amazing if we can get the tiling and the screen scaling to just work (specially screen scaling xrandr is kiling me :D) elementary OS is literally an macOS alternative.

HansJK:
The user should be able to change the grid layout. I for example like to have the middle part of the grid wider than what's on these pictures, but not so wide the it covers the whole screen. And tiling should be optional.

ToxicDragon:
Would be nice if this code were to code merged for the new release :)

fabiozaramella:
@danrabbit Is it time for a user survey? This PR seems to have gathered lot of attention.

danrabbit:
I'll defer to @cassidyjames here since he's done a lot of UX work on tiling already

andreasfelix:
Within the last commits I tried PR to address some points of criticism:

* Introduce keyboard shortcuts
    > I also really want to think about keyboard tiling with this, as I think people will expect to be able to do all the same tiling from a keyboard as soon as we would release this.
* Make behavior more consistent with existing tiling window manager - See 1.
     > a somewhat hidden, non-default behavior.
* Introduce minimum tile size (currently 600x400) 
    > Something to maybe consider is that for a lot of apps there will be a minimum size where that app is usable
* Don't require a modifier - See 5.
    >  I would much rather arrive at a solution that we could just enable by default on normal window dragging instead of requiring a modifier.

#### New Behavior

1. Similar to other tiling window managers a *Tiling-Mode*  can be activated by pressing <kbd>Super</kbd><kbd>Return</kbd>: This will shrink down the current window, which can then be moved and tiled using <kbd>↑</kbd><kbd>←</kbd><kbd>↓</kbd><kbd>→</kbd>, <kbd>w</kbd><kbd>a</kbd><kbd>s</kbd><kbd>d</kbd>, <kbd>h</kbd><kbd>j</kbd><kbd>k</kbd><kbd>l</kbd> or the mouse. Pressing <kbd>Return</kbd> or <kbd>space</kbd> will confirm the current preview and will tile the window accordingly. (Windows can also be moved between multiple monitors)
2. In the future this mode could be complement with a *Window-Selection-Mode*: Here windows can be selected using the arrow keys (similar to the Multitasking view).
3. The grid can be changed by pressing the number keys when in *Tiling-Mode*. For example pressing <kbd>3</kbd><kbd>2</kbd> will set a 3 by 2 grid.
4. Inspired by the [latest blog post](https://blog.elementary.io/dark-style-progress/) I also added optional gaps, which can be toggled by pressing <kbd>g</kbd> when in tiling mode.
5. The <kbd>Drag</kbd><kbd>Ctrl</kbd> tile action is still available. Additionally it is also possible to activate the *Tiling-Mode* by <kbd>Drag</kbd><kbd>Secondary-Click</kbd> which does not require any modifier key. I admit this also a bit hidden/non-default. Has someone a better idea how to activate the *tiling-mode* without using a modifier? Maybe be shaking the window? But this seems also weird ...

andreasfelix:
@cassidyjames What would you think about merging this as experimental feature, which could be used for the user survey. (After the last couple of changes this PR isn't even production-ready enough for an experimental mode. So there is still some cleanup needed.)

I already added an `"experimental-tiling-mode"` flag which toggles the features introduced in this PR. It could be set to `false` by default, so that adventurous users would be able to test the behavior. If it turns out that the introduced workflow isn't beneficial or there are better alternatives, it should be easy to revert this PR as changes made are relatively self-contained.

andreasfelix:
As this PR is in its current state much different from its initial implementation, I will close this PR. I also revert this branch to a [previous commit](https://github.com/elementary/gala/pull/595/commits/84aa12bf32903eae3177be98857313406abab0dc) in case someone wants to test the initial behavior. See #963 for the new implementation.

Issue #626, opened 2019-11-04
Crash when taking window screenshot when there are no windows


Exalm:
## Prerequisites
- [x] I have searched open and closed issues for duplicates.

## Describe the bug
The steps explain it.

### To Reproduce
1. Use https://github.com/elementary/screenshot/pull/147 branch of Screenshot
2. Go to an empty (aside from the Screenshot window) workspace
3. Take window screenshot

### Logs

![Screenshot from 2019-11-04 19-58-47](https://user-images.githubusercontent.com/1510457/68130643-a0dfe500-ff13-11e9-9bea-4dfa5aae7de5.png)

Here's a screenshot of a blurred backtrace without debug symbols from tty. No proper backtrace, but it looks like it's just missing a null check and trying to access a window anyway.

- [ ] I'm using the latest version from git that I've manually compiled
- [ ] I'm using the latest released stable version
- [x] I'm using the version from elementary daily repo

PR #627, opened 2019-11-04
ScreenshotManager: Null-check the window in screenshot_window()

Exalm:
Throw an error instead of crashing if there's no window to screenshot.

Fixes https://github.com/elementary/gala/issues/626


tintou:
Ideally this error would have been translated

Exalm:
Maybe, but none of the other errors there are translated either. 🤷

PR #633, opened 2019-11-07
ScreenshotManager: Add ScreenshotAreaWithCursor() and ConcealText()

Exalm:
See https://github.com/elementary/screenshot/pull/147

`ScreenshotAreaWithCursor()` is just like `ScreenshotArea()`, but it adds `include_cursor` parameter that is missing from `ScreenshotArea()` for some reason.

`ConcealText()` changes font to "Redacted Script", then after a screenshot is taken or after a 2s timeout changes it back. Subsequent calls while the text is concealed renew the timeout.


src/ScreenshotManager.vala
```diff
@@ -162,6 +178,44 @@ namespace Gala
 			selection_area.get_selection_rectangle (out x, out y, out width, out height);
 		}
 
+		private void unconceal_text ()
+		{
+			if (!text_concealed) {
+				return;
+			}
+
+			var desktop_settings = new Settings ("org.gnome.desktop.interface");
```
donadigo: Do not re-create settings everytime this method is called, make them a static member inside the `ScreenshotManager` class and initialize in `static construct`.

src/ScreenshotManager.vala
```diff
@@ -162,6 +178,44 @@ namespace Gala
...
+			if (text_concealed) {
+				return;
+			}
+
+			Timeout.add (UNCONCEAL_TEXT_TIMEOUT, () => {
+				unconceal_text ();
+				return Source.REMOVE;
+			});
+
+			var desktop_settings = new Settings ("org.gnome.desktop.interface");
```
donadigo: Exactly the same thing here.

src/ScreenshotManager.vala
```diff
@@ -162,6 +178,44 @@ namespace Gala
...
+			text_concealed = false;
+		}
+
+		public async void conceal_text () throws DBusError, IOError
+		{
+			if (text_concealed) {
+				return;
+			}
+
+			Timeout.add (UNCONCEAL_TEXT_TIMEOUT, () => {
```
donadigo: I'm not sure if the choice to ignore subsequent calls to `conceal_text` is correct here. I feel like instead, a subsequent call should cancel the current timeout and add a new one with a duration of 2s.

src/ScreenshotManager.vala
Exalm: Considering `ScreenshotManager` already a singleton, I think a non-static field and `construct` makes more sense.

Issue #634, opened 2019-11-08
Gala crash on startup with mutter 3.3x


Tireg:
## Prerequisites
- [X] I have searched open and closed issues for duplicates.

## Describe the bug
When gala is built with mutter 3.3x, it crashes on startup.
This bug report follows the discussion in https://github.com/elementary/gala/pull/566.

### To Reproduce
1. Start gala compiled with mutter 3.3x

### Expected behavior
Gala should start properly

### Logs
See logs reported after https://github.com/elementary/gala/pull/566 was merged.

## Platform Information
OS: Gentoo
Mutter version: 3.32.2

- [X] I'm using the latest version from git that I've manually compiled
- [ ] I'm using the latest released stable version

## Additional context
I've already triggered the bug(s) origin. I opened this bug report to track issues related to building against mutter 3.3x.
Within the comment, I'll explained each new bug, what patch is used in order to fix it, and why the original code was buggy.

Tireg:
So the PR related to building with mutter 3.3x has been merged. We can finally build gala against it, yay !
It seems however some critical bugs are still present and prevent using gala when build with mutter 3.3x.
After some investigation, I finally got it to work almost entirely properly.

In this post, I'll detail all the steps I used in order to make everything work.

## Steps to fix gala with Mutter 3.3x

### Building against Mutter 3.32
It seems the upstream gala builds fine agains mutter 3.34 and mutter 3.30 but fails to build agains mutter 3.32.
It looks like the PR has some changed code for `MUTTER_334` that should also be made for `MUTTER_332`, concerning background management. Rather simple, just replace the `if HAS_MUTTER334` with `HAS_MUTTER332` for things related to background management (the related errors are reported by your compiler when building fails).
In order to fix this, I use the following patch: https://gist.github.com/Tireg/40a88ef867e978644ef3024d3d28ec55

### Getting dynamic workspaces to work
Ok, so I have my gala working, but there is no animations, no dynamic workspace, ... This is not my gala ! It's almost unusable this way :(
It seems like applying gsettings has changes with mutter 3.30. We need to use a schema override. Here is a simple fix: https://gist.github.com/Tireg/7af22e747425b5a4f52b055a7b878a73

### Fix gala crash on startup
Oh crap ! I can't even start a session with my newly built gala. I get a stracktrace looking like https://gist.github.com/worldofpeace/8e2d3ff4a442a09454a6b91e69af290b.
[...]

PR #635, opened 2019-11-08
Fix mutter 3.3x crashes

Tireg:
This PR fixes issues against running upstream gala with mutter 3.3x.
This should fix https://github.com/elementary/gala/issues/634

Every single patch is detailed in this post: https://github.com/elementary/gala/issues/634#issuecomment-551851873
I reported every step as a commit comment in order to understand it more easily.

### DIFF
--- a/src/Widgets/MultitaskingView.vala
+++ b/src/Widgets/MultitaskingView.vala
@@ -92,6 +92,9 @@ namespace Gala
 
 			manager.workspace_added.connect (add_workspace);
 			manager.workspace_removed.connect (remove_workspace);
+#if HAS_MUTTER334
+			manager.workspaces_reordered.connect (() => update_positions (false));
+#endif
 			manager.workspace_switched.connect_after ((from, to, direction) => {
 				update_positions (opened);
 			});
--- a/src/Widgets/MultitaskingView.vala
+++ b/src/Widgets/MultitaskingView.vala
@@ -392,7 +392,7 @@ namespace Gala
 			// FIXME is there a better way to get the removed workspace?
 #if HAS_MUTTER330
             unowned Meta.WorkspaceManager manager = display.get_workspace_manager ();
-			unowned List<Workspace> existing_workspaces = null;
+			List<Workspace> existing_workspaces = null;
 			for (int i = 0; i < manager.get_n_workspaces (); i++) {
 				existing_workspaces.append (manager.get_workspace_by_index (i));
 			}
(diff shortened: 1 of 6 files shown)


src/Widgets/MultitaskingView.vala
```diff
@@ -389,29 +392,20 @@ namespace Gala
 			// FIXME is there a better way to get the removed workspace?
 #if HAS_MUTTER330
             unowned Meta.WorkspaceManager manager = display.get_workspace_manager ();
+			unowned List<Workspace> existing_workspaces = null;
```
tintou:
```suggestion
			List<Workspace> existing_workspaces = null;
```

src/WorkspaceManager.vala
```diff
@@ -132,10 +132,16 @@ namespace Gala
 
 		void workspace_removed (Meta.WorkspaceManager manager, int index)
 		{
+			unowned List<Workspace> existing_workspaces = null;
```
tintou:
```suggestion
			List<Workspace> existing_workspaces = null;
```

src/WorkspaceManager.vala
```diff
@@ -389,9 +395,13 @@ namespace Gala
 #if HAS_MUTTER330
             unowned Meta.Display display = wm.get_display ();
 			unowned Meta.WorkspaceManager manager = display.get_workspace_manager ();
-			var last_index = manager.get_n_workspaces () - 1;
+			unowned List<Meta.Workspace> workspaces = null;
```
tintou:
```suggestion
			List<Meta.Workspace> workspaces = null;
```

Tireg:
> By creating an `unowned List` you're creating a leak here

For each one of the suggestions, we fill the list using `manager.get_workspace_by_index (i)`. I don't know Vala but isn't there something like `List<unowned Workspace>` (instead of `List<Workspace>`) that I should use in order to avoid destroying the `Workspace`s objects when the `List` is destroyed ?

Just to be sure there is no mistake.

tintou:
@Tireg a new commit is fine too, no worries :)
The difference here:
`List<unowned Workspace>` means that we have to free the List but not the elements
`List<Workspace>` means that we have to free the list and unref the elements
Defined as `List<Workspace>`, Vala will take care to increase the reference count before adding an element to the list, when `List<unowned Workspace>` would not ref the workspace, so `List<Workspace>` is safer for us in this case.

decathorpe:
I'm looking forward to testing this on fedora :)

Just one question / nit-pick: The name for the gsettings override file seems rather odd:

`20_elementary.pantheon.wm.gschema.override`

Can we rename it to something like `20_io.elementary.gala.gschema.override` instead?

tintou:
@decathorpe Yeah, would make sense too :)

decathorpe:
Hm. I'm still getting a reproducible crash on startup on fedora 31 :( I'll try to file a new bug report for it, once I'm able to provide a useful traceback :(

PR #640, opened 2019-11-19
Make Mutter 3.28 the minimum libmutter version

tintou:
We don't even expect gala to work with something lower to 3.28


tintou:
@ricotz done, applied the same heuristic with the other vapi

Issue #637, opened 2019-11-13
mutter 3.34: windows not drawn in Windows Overview or Multitasking view


worldofpeace:
## Describe the bug
When using Gala with Mutter 3.34, windows are visible in Multitasking View or Windows Overview.
The only visible elements for a window would be its workspace icon, and the close window button.
(I'm assuming those are independent here)

### To Reproduce
You'll need to be running gala built with mutter 3.34, which will likely imply the entire Gnome 3.34 stack. Please make sure you've also compiled wingpanel and greeter from master, which has patches for newer mutter compatibility.

1. Open a window
2. Open Windows Overview or Multitasking View
3. You can't see the windows

### Expected behavior
The windows content will be visible in both Multitasking and Overview views.

###### Multitasking View
![Screenshot from 2019-11-10 21-26-33](https://user-images.githubusercontent.com/28888242/68556896-06dfd700-0402-11ea-9347-c58aab318c82.png)

###### Show all windows/Windows Overview
![Screenshot from 2019-11-13 15-29-50](https://user-images.githubusercontent.com/28888242/68802012-1ee66f00-062b-11ea-842e-af414245e1a7.png)

### Logs
I see things like
```
.gala-wrapped[1375]: g_object_new_valist: invalid object type 'MetaShapedTexture' for value type 'ClutterActor
(gala:1622): Clutter-CRITICAL **: 20:45:35.633: _clutter_stage_queue_event: assertion 'CLUTTER_IS_STAGE (stage)' failed
(gala:1622): GLib-GObject-WARNING **: 20:45:36.446: invalid cast from 'CoglTexture2D' to '(null)'
```

in the logs. Nothing special when trying to reproduce this bug however.

## Platform Information

- OS: NixOS
- OS Version: 20.03 (development)

Gnome 3.34 stack, using wingpanel, gala, and greeter from git master.

- [x] I'm using the latest version from git that I've manually compiled
- [ ] I'm using the latest released stable version

## Additional context
I originally raised this issue in #634. 
Here's @Tireg comments:
> @worldofpeace I'm afraid I don't have enough Vala/Gobject/Clutter experience to understand what's actually responsible for drawing the objects.
The src/Widgets/WindowClone.vala seems to be the Clutter Actor responsible of the windows when you open multitasking view and windows-overview. It contains the window icon and the window close button, as well as the box responsible to catch mouse events and the window shadow.
[...]

PR #642, opened 2019-11-21
Regenerate the vapi using the .metadata files

tintou:
We are now sure that we have the right symbols available with the right version

### DIFF
--- a/plugins/notify/NormalNotification.vala
+++ b/plugins/notify/NormalNotification.vala
@@ -343,9 +343,9 @@ namespace Gala.Plugins.Notify
 				return null;
 
 #if HAS_MUTTER330
-		    unowned GLib.List<weak Meta.WindowActor>? actors = Meta.Compositor.get_window_actors (display);
+		    unowned GLib.List<Meta.WindowActor> actors = display.get_window_actors ();
 #else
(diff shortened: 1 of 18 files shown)


tintou:
@worldofpeace can you see if that fixes #637

worldofpeace:
> @worldofpeace can you see if that fixes #637

Sure, will check :+1:

worldofpeace:
Ok, compilation fails on NixOS Gnome 3.34 with these regenerated vapi's
* https://gist.github.com/worldofpeace/c4d7f2afc0d00faee3d9ee4987db1926

lots with `error: too few arguments to function 'g_boxed_free'`.

@tintou What do you think about having a never ubuntu target in CI for the mutter using projects?
Just think it would make things easier to test against multiple mutter versions.

tintou:
@worldofpeace it looks like an issue with Vala, can you try with Vala master to see if it has been fixed and otherwise open a bug there? I think that I was using latest Vala to build it here

worldofpeace:
> @worldofpeace it looks like an issue with Vala, can you try with Vala master to see if it has been fixed and otherwise open a bug there? I think that I was using latest Vala to build it here

I suspected similar, will try.

worldofpeace:
Thanks, works :+1: 

Screenshot from 2019-11-22 03-34-57

Tested overview mode also.

PR #711, opened 2020-01-17
Fix broken state of notification windows when switching workspaces

### LINKED ISSUE #706: Notification windows get stuck if you change workspaces during their opening animation
1. Receive a notification
2. Change workspaces during the notification window's opening animation
3. See that the notification freezes in an awkward skewed state

donadigo:
Fixes #706.

Of course Clutter does not expose a way to specify flags to remove the child from a parent without also stopping all the running transitions so we have to do some manual restoring to make this work properly.


danrabbit:
This definitely fixes the issue.

Requesting a review from @tintou to make sure we don't break Mutter 3.30

PR #741, opened 2020-03-07
Add LibMutter 6 to build deps

danrabbit:


tintou:
Blocked until proper support lands in master

PR #760, opened 2020-04-01
Support g-s-d UngrabAccelerators API

worldofpeace:
This was added in g-s-d 3.33.0 [0].

Without this I was seeing:
```
gsd-media-keys[1927]: Failed to ungrab accelerators: GDBus.Error:org.freedesktop.DBus.Error.UnknownMethod: No such method “UngrabAccelerators”
```

in the journal from g-s-d.

[0]: https://gitlab.gnome.org/GNOME/gnome-settings-daemon/-/blob/GNOME_SETTINGS_DAEMON_3_33_0/NEWS#L8

### DIFF
--- a/src/DBusAccelerator.vala
+++ b/src/DBusAccelerator.vala
@@ -127,6 +127,15 @@ namespace Gala {
             return ret;
         }
 
+#if HAS_MUTTER334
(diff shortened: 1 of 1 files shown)


worldofpeace:
cc @tintou @davidmhewitt

PR #784, opened 2020-04-26
fix Gala.Utils.create_close_button and Gala.Utils.create_refresh_butt…

jgoday:
I'm using pantheon on fedora 32 and ubuntu 20.04 (manually compiled in both),
and I'm facing the same issue with PIP (awesome feature BTW), close and resize handles are both invisible, and pip windows cannot be closed.

This issue seems to disappear with the following changes in lib/Utils.vala.

It's like the texture does not take the current size of the pixmap.

Is anyone else affected by this problem?

Mutter version: 3.36.1
Clutter version: 1.26.4


tintou:
Yes it is also an issue on ubuntu 20.04, thank you for the fix :)

PR #795, opened 2020-05-07
Close multitasking view if workspace is selected

### LINKED ISSUE #776: Multitasking View. Clicking on an already selected workspace (bottom square) should exit overview.
1- Open an App.
2- Open multitasking view. Click on the "+" square on the bottom.
3- Click again on the "+" square on the bottom.

**Result:** An error sound is heard and nothing happens.

Wouldn't it be more beneficial if clicking on an already selected workspace would exit multitasking view like it happens when Enter/Esc/Super+S/Super+Down is pressed?

P.S: I'm sorry if this issue has been brought before.

andreasfelix:
Fixes #776


andreasfelix:
Oooops, not exactly sure how this happened 🙄️ 

Initially I used more verbose code to actually close the multitasking-view instead of just the workspace clone. Then I found the existing

https://github.com/elementary/gala/blob/9897f02702cd0547c74eeb07af132af629cf89f8/src/Widgets/MultitaskingView.vala#L556

method of the `MultiTaskingView` and must have confused it with the close method of the `Workspace-Clone`: 

https://github.com/elementary/gala/blob/9897f02702cd0547c74eeb07af132af629cf89f8/src/Widgets/WorkspaceClone.vala#L419

However, I think the easiest solution is to just emit the `selected` signal in the same way when the background is clicked:

https://github.com/elementary/gala/blob/9897f02702cd0547c74eeb07af132af629cf89f8/src/Widgets/WorkspaceClone.vala#L170-L173


BTW, thanks for the invitation!

PR #809, opened 2020-05-23
Implement ScreenSaver interfaces

### LINKED ISSUE #716: Provide an embedded screen-saver
We need to provide our own screensaver embedded into Gala as light-locker is not being actively maintained anymore and it won't be secured enough for Wayland anyway.

davidmhewitt:
Fixes #716 

This implements a screen saver/screen locker embedded in the compositor, effectively replacing `light-locker`.

To test:
* Install this branch
* Disable `light-locker` by renaming it's autostart file at `/etc/xdg/autostart/light-locker.desktop` to `/etc/xdg/autostart/light-locker.desktop.disabled`
* Reboot
* Set `/org/pantheon/desktop/gala/behavior/lock-on-suspend` and `/org/pantheon/desktop/gala/behavior/lock-on-screen-off` to your desired values. 

Notes:
The "turn off display when inactive for" option in the power plug still controls the session timeout (i.e. when the screensaver activates), and the suspend settings should work as expected.

The two locking options in the privacy plug will no longer do anything due to them controlling `light-locker` dconf keys, hence the manual setting of the two keys above. We'll need to follow up with the necessary changes in the switchboard plug once we're happy with this.

I'm hoping this will help with resolving a lot of sleep/resume problems people have been experiencing, especially with laptops with external displays and with cases of devices suspending when they shouldn't. This does true late locking now and stays in your user session, not switching to the greeter until there's activity, rather than switching instantly to the lightdm session running as a different user.

Plus there's now a nice fade out animation (300ms on manual lock and 3 seconds on session timeout, which can be interrupted). This also allows music to continue playing in the background when the screensaver comes on and gives us the potential to display stuff on the "lock screen" in future if we wish.


src/ScreenSaverManager.vala
```diff
@@ -0,0 +1,119 @@
+//
+//  Copyright (C) 2020 elementary, Inc. (https://elementary.io)
```
tintou, line 2:
```suggestion
//  Copyright 2020 elementary, Inc. (https://elementary.io)
```

src/ScreenSaverManager.vala
```diff
@@ -0,0 +1,119 @@
...
+            if (started > 0) {
+                return (uint)Math.floor ((GLib.get_monotonic_time () - started) / 1000000);
+            } else {
+                return 0;
+            }
+        }
+    }
+
+    [DBus (name="org.gnome.ScreenSaver")]
+    public class GNOMEScreenSaverManager : ScreenSaverManager {
```
tintou, line 73:
```suggestion
    public class GNOMEScreenSaverManager : GLib.Object {
```

tintou: And better let it be an implementation detail of the ScreenSaverManager itself (i.e. let it be instantiated by the ScreenSaverManager itself and let this class be private)

src/Widgets/ScreenShield.vala
```diff
@@ -0,0 +1,433 @@
...
+    enum PresenceStatus {
+        AVAILABLE = 0,
+        INVISIBLE = 1,
+        BUSY = 2,
+        IDLE = 3
+    }
+
+    [DBus (name = "org.gnome.SessionManager.Presence")]
+    interface SessionPresence : Object {
+        public abstract uint status { get; }
```
tintou, line 57:
```suggestion
        public abstract PresenceStatus status { get; }
```

src/Widgets/ScreenShield.vala
```diff
@@ -0,0 +1,433 @@
...
+        public signal void status_changed (uint new_status);
```
tintou, line 58:
```suggestion
        public signal void status_changed (PresenceStatus new_status);
```

src/Widgets/ScreenShield.vala
```diff
@@ -0,0 +1,433 @@
...
+    public struct LoginDisplay {
+        string session;
+        GLib.ObjectPath objectpath;
+    }
+
+    [DBus (name = "org.freedesktop.login1.User")]
+    interface LoginUserManager : Object {
+        public abstract LoginDisplay? display { owned get; }
+    }
+
```
tintou, line 47: [DBus (signature = "u")]

src/Widgets/ScreenShield.vala
```diff
@@ -0,0 +1,433 @@
...
+                in_greeter = true;
+            // Otherwise, we're in screensaver mode, just deactivate
+            } else if (!is_locked) {
+                debug ("user became active in unlocked session, closing screensaver");
+                deactivate (false);
+            }
+        }
+
+        private LoginSessionManager? get_current_session_manager () throws GLib.Error {
+            string? session_id = GLib.Environment.get_variable ("XDG_SESSION_ID");
```
tintou, line 292:
```suggestion
            unowned string? session_id = GLib.Environment.get_variable ("XDG_SESSION_ID");
```
Unsure if that's possible with `login_user_manager.display.session` though

src/Widgets/ScreenShield.vala
```diff
@@ -0,0 +1,433 @@
...
+
+                restore_easing_state ();
+
+                _set_active (true);
+
+                if (screensaver_settings.get_boolean (LOCK_ENABLED_KEY)) {
+                    @lock (false);
+                }
+
+                return false;
```
tintou, line 377:
```suggestion
                return GLib.Source.REMOVE;
```

src/Widgets/ScreenShield.vala
```diff
@@ -0,0 +1,433 @@
...
+        public signal void status_changed (uint new_status);
+    }
+
+    [DBus (name = "org.freedesktop.DisplayManager.Seat")]
+    interface DisplayManagerSeat : Object {
+        public abstract void switch_to_greeter () throws GLib.Error;
+    }
+
+    public class ScreenShield : Clutter.Actor {
+        static ScreenShield? instance;
```
tintou, line 67:
```suggestion
        private ScreenShield? instance;
```
You don't really need it to be static, you'll always have just one ScreenShied per Gala process anyway, won't you?

tintou:
Thanks for this work, that should really fix a bunch of issues indeed.

After looking at the code I'm unsure that we want ScreenSaverManager and GNOMEScreenSaverManager to be singletons, the GNOMEScreenSaverManager one can be a private instance of the ScreenSaverManager object and the ScreenSaverManager can just be instantiated by the WindowManager class and passed as argument when creating the ScreenShied object. Which would make true ownership more transparent.

src/Widgets/ScreenShield.vala
davidmhewitt: Yeah, unfortunately this doesn't work with `login_user_manager.display.session`

danrabbit:
I can confirm this is largely working as expected. The one problem I am noticing is that there is no fade out when manually locking. (the fade out for session timeout works as expected). When locking manually, after logging back in I get a black flash.

sbarrow:
Totally locks up my machine, won't let me log back in. Attached syslog. In all likelihood I'm doing something wrong here compiling and installing, I'm somewhat out of my element but if these logs mean anything to you guys here they are (they're in reverse, used tac instead of cat).

[syslog.txt](https://github.com/elementary/gala/files/4679140/syslog.txt)

davidmhewitt:
@sbarrow There's a lot of errors about "bamf" in that log which makes me think something unrelated to this branch has gone a bit awry.

Have you rebooted since you installed this branch?

sbarrow:
@davidmhewitt I have not rebooted. I will try that, and I'll try a guest session. Also, I didn't think to ask if this will run on v5, or is this relying on v6? I'm running v5.1.4. And I'm using master branch as it looks like this was merged.

sbarrow:
Restarting or using another user account still crashes. Here's a new log.

[newlog.txt](https://github.com/elementary/gala/files/4679206/newlog.txt)

davidmhewitt:
@sbarrow I did some light testing of this branch on 5.1.4 and it definitely didn't break the session and appeared to work.

However, there have been some other branches merged recently that are likely to cause breakage on 5.1, related to notifications. 

On 6.0 , notifications are handled by a separate process, no longer a plugin to Gala. I suspect what's happened is you still have the notifications plugin installed at `/usr/lib/x86_64-linux-gnu/gala/plugins/libgala-notify.so` and the rest of Gala that it needs to hook into is now gone.

Deleting that plugin file may help, and it may not. At the very least you'll have no notifications. If you want to test out this branch, then I'd recommend checking out the specific commit: https://github.com/elementary/gala/commit/28afa1f9fdaa584c04c86812c85f26918b849026

This was after this branch was merged, but before any of the notification stuff changed.

Otherwise, there may or may not be some 6.0 daily (testing) builds coming soon that you may or may not be able to get access to :wink: 

I guess it depends on what you're trying to achieve and how stable you want the end result to be, because unfortunately this screen locker code hasn't had very wide testing yet and I think it's unlikely it'll make it into 5.1, so trying anything built on top of 5.1 isn't really going to be a supported configuration.

sbarrow:
Got it working by checkout out that commit you mentioned. No problem with external monitor, works great. Can open and close my laptop and everything. However, I now have the opposite problem - if I unplug the external monitor and then lock, I get a blank screen and/or it kills my session.

That being said, probably not worth debugging on 5 if there are no plans to release for 5. I was expecting on waiting for 6.0 for this to work reliably anyway. Happy to help with testing if you guys put anything out.

davidmhewitt:
So getting a blank screen when you lock is the (new) intended behaviour. Perhaps we need to look at putting *something* on the screen so it's less concerning that it's just a blank screen. But on the other hand, leaving it blank allows the monitor to be put into power saving mode to save power.

So the actual problem in your case is that lightdm appears when you have an external monitor plugged in which isn't something I've found during my testing. But I haven't tested that case on 5.1 since I only have a VM install of 5.1.

Watch out for some 6.0 testing opportunities and I'd be interested to hear how it goes there.

sbarrow:
Regarding behavior, here's my take, just speaking as a user.

My use case for locking is that I get up from my desk at least 20 or 30 times a day. I hit Super+L and when I come back, the greeter is there waiting for me, no delays. If I don't come back for say 30 minutes, the system goes into suspend automatically, which is fine (this is how Elementary works on my desktop where these lid open/close issues aren't present and it's perfect).

If power is a concern, it probably means I'm using my laptop away from my desk anyway so I won't even need to do that, I can just close the lid and get a real power-saving suspend mode rather than just a blank screen. If I was at my desk but not plugged in (never happens), I would just hit suspend in the power menu instead of lock.

In any case, I would definitely put something on the screen. Something like Android phones do would be cool, but that's probably a project in itself, and I think the greeter would be just as good.

A blank screen is going to be very confusing to people - it didn't even occur to me to press a key for a few minutes to see if that woke things up, since the screen was clearly fully powered up, just blank, which with with Linux tends to mean something is wrong since there are a lot of very annoying bugs that cause this exact symptom. Come to think of it, the light-locker bug that brought us here did exactly this.

Again, just my take as a user, obviously it's not my call and anything beats broken lol. It does seem to work reliably now. Thanks again for the hard work.

Also, upon further testing, I'm needing to press a key with the external monitor plugged in as well now. Might have just been a fluke.

davidmhewitt:
@sbarrow We purposefully don't switch to the greeter immediately as we believe that was the cause of a bunch of other issues like devices suspending when they shouldn't. The greeter runs as a different user, so as soon as you switch to it, you haven't got any access to user specific settings, playing music stops etc...

I suspect there's some residual bits of config on your machine that are preventing the monitor going into power save when the screen is blank and they won't be present in 6.0.

However, it'd be cool if you could take the gist of your comment above and open a separate issue here in the gala repo. I think there's probably a desire to display some stuff on the lock screen but it'll need input from the design team before anyone can even think about developing it. And as you say, it's probably not a small project.

Anyway, thanks again for sponsoring me on this. I think I spent way more than an hour looking into it. Then again, I do say a minimum of an hour and it was an interesting issue to look at 😂

sbarrow:
@davidmhewitt Sounds good, I'll open an issue. If the screen had gone into power save, that would be much better than what happened on mine. My only concern would be speed of bringing up the greeter as I lock/unlock my system anywhere from 20 to 50 times a day.

>We purposefully don't switch to the greeter immediately as we believe that was the cause of a bunch of other issues like devices suspending when they shouldn't. The greeter runs as a different user, so as soon as you switch to it, you haven't got any access to user specific settings, playing music stops etc...

Interesting. My audio has always continued playing at the greeter when using light-locker. Should this not be the case?

Im actually now having an intermittent issue where maybe 1 in every 5 times I lock using this new version of Gala, it suspends me (lid closed, external monitor, but my settings in logind.conf should handle this properly). I can open an issue for it if you want but as I said earlier maybe it's not worth the trouble to debug on version 5.

Like I said I don't mind just paying for your time per job since obviously it's hard to predict how long things will take. I've got some more issues so we'll discuss.

sbarrow:
Did some more testing, this may or may not be useful but here it is. Seems I was mistaken earlier about having to press a key w/ monitor plugged in, at least reliably. I can do it again and get logs if it's helpful.

Laptop open - no external monitor: 10 unlock cycles. Screen goes blank, I press a key, greeter comes up. All good.

Laptop closed - external monitor plugged in: 20 unlock cycles.
16 - Went straight to greeter. Did not have to press a key.
2 - Screen went blank for a second, then system suspended, then woke itself up, went to greeter. Did not have to press a key.
2 - Went to blank screen. Waited a few seconds. Had to press a key for it to do anything. System then suspended. Had to open and close laptop to get to the greeter.

I'm not doing anything inbetween these. Logging in, locking, logging in, locking again, etc etc.

Rantyrant:
> So getting a blank screen when you lock is the (new) intended behaviour. Perhaps we need to look at putting _something_ on the screen so it's less concerning that it's just a blank screen. But on the other hand, leaving it blank allows the monitor to be put into power saving mode to save power.
> 
If I may butt in here: If I lock the screen, manually(Super + L), I would expect it to go out on the user switcher, and then blank the screen maybe after a minute. Just to show the user it's actually locked and requires a password to get back in.

Also, I'd be happy to test once there's some alpha/beta/whatever testing available. :)

sbarrow:
@Rantyrant 

You can test it on 5.1.4 by running off of that commit David mentioned. I've been running it for the last few days. I think there will be laptop lid + external monitor issues, but those were there before too.

git clone https://github.com/elementary/gala
cd gala
git checkout 28afa1f9fdaa584c04c86812c85f26918b849026

Then follow these instructions: https://github.com/elementary/gala

If it breaks, hit Ctrl+Alt+F1, log in to console and "sudo apt reinstall gala" then Ctrl+Alt+F7 to go back.

Rantyrant:
@sbarrow I'm not currently running elementary, and won't be until 6.0, so if I have to start up a testing drive, I'm going to wait for the full 20.04/Odin base. :p

sbarrow:
Went out and got a new laptop just to test some of this stuff, and it works fine there. Still does the powered on blank screen which I don't like, but it does seem to work consistently.

PSA, don't try to run Elementary on an HP Spectre, especially not the 2017 model. This wasn't even worst problem I had with it.

PR #813, opened 2020-05-25
Gschema: Move screensaver keys to io.elementary.desktop

danrabbit:
Since these are new keys, it would be nice to have them RDNN'd correctly so we don't have to migrate them later

### DIFF
--- a/data/gala.gschema.xml
+++ b/data/gala.gschema.xml
@@ -16,7 +16,7 @@
 		<value nick='natural' value='1'/>
 	</enum>
 
-	<schema path="/io/elementary/desktop/screensaver/" id="io.elementary.desktop.screensaver/">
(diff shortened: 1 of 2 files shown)


data/gala.gschema.xml
```diff
@@ -15,7 +15,20 @@
 		<value nick='grid' value='0'/>
 		<value nick='natural' value='1'/>
 	</enum>
-	
+
+	<schema path="/io/elementary/desktop/screensaver/" id="io.elementary.desktop.screensaver/">
```
davidmhewitt, line 19: Is that slash supposed to be on the end of the ID?

danrabbit: I believe so. It looks like that's how the other schema paths are done

davidmhewitt: I was looking at the `id` parameter, not the `path` one.

Issue #758, opened 2020-03-16
20_elementary.pantheon.wm.gschema.override refers to non-existent key


jeremypw:
I have searched open and closed issues for duplicates.

## Describe the bug

The gsettings override file 20_elementary.pantheon.wm.gschema.override refers to the key `button-layout` in the schema `[org.gnome.mutter:Pantheon]`.  This does not exist in Hera 5.1 daily.

### To Reproduce
1. Run `sudo glib-compile-schemas --dry-run --strict /usr/share/glib-2.0/schemas
` in a terminal.
2. Output `No such key 'button-layout' in schema 'org.gnome.mutter:Pantheon' as specified in override file '/usr/share/glib-2.0/schemas/20_elementary.pantheon.wm.gschema.override'`

### Expected behavior
No error

## Platform Information
Hera 5.1 daily.

I'm using the latest version from daily repository. Updated through AppCenter

oberon-manjaro:
Confirming. Additionally also key `enable-animations` does not exist:
```
No such key “button-layout” in schema “org.gnome.mutter:Pantheon” as specified in override file “/usr/share/glib-2.0/schemas/20_elementary.pantheon.wm.gschema.override”; ignoring override for this key.
No such key “enable-animations” in schema “org.gnome.mutter:Pantheon” as specified in override file “/usr/share/glib-2.0/schemas/20_elementary.pantheon.wm.gschema.override”; ignoring override for this key.
```

PR #852, opened 2020-06-15
override: Update the mutter schemas

tintou:
Fixes https://github.com/elementary/gala/issues/758

### DIFF
--- a/data/20_elementary.pantheon.wm.gschema.override
+++ b/data/20_elementary.pantheon.wm.gschema.override
@@ -1,6 +1,7 @@
 [org.gnome.mutter:Pantheon]
 dynamic-workspaces = true
 attach-modal-dialogs = false
-button-layout = 'close:maximize'
 edge-tiling = true
-enable-animations = true
+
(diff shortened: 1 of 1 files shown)


data/20_elementary.pantheon.wm.gschema.override
```diff
@@ -1,6 +1,7 @@
 [org.gnome.mutter:Pantheon]
 dynamic-workspaces = true
 attach-modal-dialogs = false
-button-layout = 'close:maximize'
 edge-tiling = true
-enable-animations = true
+
+[org.gnome.desktop.wm.preferences]
```
danrabbit, line 6: This should still be 
```suggestion
[org.gnome.desktop.wm.preferences:Pantheon]
```
right?

worldofpeace: Yep, you only want that preference on Pantheon (it's probably the most pantheon specific thing). That way it can co-exist with other session with different button-layout

PR #859, opened 2020-06-25
MaskCorners: Remove radius setting

danrabbit:
And bump to 6px to match Greenfield

There's no real reason this should be configurable


tintou:
Don't remove the setting now, just change the description to "DEPRECATED, not used anymore"

PR #869, opened 2020-07-09
BackgroundSource: use map values, not map entries

danrabbit:


andreasfelix:
I can confirm, that this builds correctly. As far as I understand this was caused by the newer Vala handling iterations over the Gee.HashMap differently: It now seems to loop over the entries instead of the values.

Issue #876, opened 2020-07-18
Increase zoom factor in Gala's Zoom plugin


mmohsinkhan:
I am a software developer with eye sight problem, I need screen magnification to carry out routine work on my computer.

Screen magnification works on Elementary OS but current max magnification factor (2.5 times) is not enough for me. Please make it at least 6 times. If required I can open Pull Request with increased magnification factor.

For now I have build and installed gala from source after increasing magnification factor. 

Best Regards

meisenzahl:
I saw your post on Reddit some time ago and actually wanted to create an issue here. Good that you were faster 😀️

How did you implement it? Did you just increase the maximum zoom factor to `6.0f`? https://github.com/elementary/gala/blob/266bc587e9aa2ade7a14347212a6f19041b03a6d/plugins/zoom/Main.vala#L85-L86

So you can now increase the zoom factor with a step size from `0.5f` to `6.0f`? https://github.com/elementary/gala/blob/266bc587e9aa2ade7a14347212a6f19041b03a6d/plugins/zoom/Main.vala#L124

Maybe it would be a good idea to make the maximum zoom factor and step size adjustable in [Switchboard Display Plug](https://github.com/elementary/switchboard-plug-display)?

mmohsinkhan:
Good to see you here, because of you I'm using EOS as my daily driver.

I made the simplest change possible to make it work on my system, changed 2.5f to 6.0f on line # 85 :).

+1 for Switchboard Display Plug option.

meisenzahl:
> Good to see you here, because of you I'm using EOS as my daily driver.

Because of me? What have I done? 😅️

mmohsinkhan:
> Because of me? What have I done?

It was not possible for me to use EOS with out of the box magnification factor. Because of your answer on Reddit I was able to increase magnification factor.

meisenzahl:
I didn't answer your question at Reddit. Another user was faster 😁️

mmohsinkhan:
> I didn't answer your question at Reddit. Another user was faster

Ohh. Hehe.
But still I must say thanks, you are marking required updates in source code here.

PR #877, opened 2020-07-19
Increase maximum for zoom plugin

meisenzahl:
Closes https://github.com/elementary/gala/issues/876


meisenzahl:
@mmohsinkhan has described that he needs at least a zoom factor of `6.0f`. Should the maximum then be even higher, e.g. `10.0f`? Should the step size still be `0.5f`?

mmohsinkhan:
Max magnification factor 2.5f set currently is not enough for people like me. I need to magnify desktop more than 2.5 times to make it usable for me.
I don't think setting max factor to 10f would do any harm, on the bright people like me would be able to use Elementary out of the box.

Thanks

plugins/zoom/Main.vala
```diff
@@ -82,7 +82,7 @@ namespace Gala.Plugins.Zoom {
             // Nothing to do if zooming out of our bounds is requested
             if (current_zoom <= 1.0f && !@in)
                 return;
-            else if (current_zoom >= 2.5f && @in)
+            else if (current_zoom >= 10.0f && @in)
                 return;
```
danrabbit, line 86: While we're here, a little bit of feedback couldn't hurt :)
```suggestion
                Gdk.beep ();
                return;
```

meisenzahl: Many thanks for the feedback :) It is your love for detail that makes elementary OS so unique.

meisenzahl:
The solution was much easier than I thought #Bitesize

mmohsinkhan:
Thanks guys for the support.

PR #879, opened 2020-07-20
Desktop context menu

### LINKED ISSUE #521: Expose Wallpaper and Display settings on the Desktop like GNOME
GNOME, like elementary, doesn't have Desktop icons too. But it does make the menu useful by providing only three settings there:
1. Display settings
2. Change wallpaper
3. System settings

This falls in line with elementary's style of exposing relevant settings throughout the OS. For instance, the wingpanel indicators expose relevant settings rather conveniently. It would be great to extend this to the Desktop as well. It can also help set expectations of users coming from traditional desktop paradigms who might at first be confused as to why the right click isn't working!

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/72968228-expose-wallpaper-and-display-settings-on-the-desktop-like-gnome?utm_campaign=plugin&utm_content=tracker%2F62872462&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F62872462&utm_medium=issues&utm_source=github).
</bountysource-plugin>

lainsce:
The goal here is to have a desktop context menu.

Looks like this:
Screenshot from 2020-07-20 15 49 09

Fixes #521

### DIFF
--- a/daemon/MenuDaemon.vala
+++ b/daemon/MenuDaemon.vala
@@ -270,7 +270,6 @@ namespace Gala {
                     try {
                         AppInfo.launch_default_for_uri ("settings://desktop/appearance/wallpaper", null);
                     } catch (Error e) {
-                        warning ("Failed to open wallpaper settings: %s", e.message);
                         var message_dialog  = new Granite.MessageDialog.with_image_from_icon_name (
                             "Failed to Open Wallpaper Settings",
                             "The error was: Could not launch Switchboard.",
--- a/daemon/MenuDaemon.vala
+++ b/daemon/MenuDaemon.vala
@@ -314,7 +314,7 @@ namespace Gala {
             desktop_menu.show_all ();
         }
 
-        public void show_desktop_menu (Gala.WindowFlags flags, int x, int y) throws DBusError, IOError {
+        public void show_desktop_menu (int x, int y) throws DBusError, IOError {
             if (desktop_menu == null) {
                 init_desktop_menu ();
             }
--- a/daemon/MenuDaemon.vala
+++ b/daemon/MenuDaemon.vala
@@ -275,7 +275,11 @@ namespace Gala {
             change_wallpaper = new Gtk.MenuItem ();
             change_wallpaper.add (change_wallpaper_accellabel);
             change_wallpaper.activate.connect (() => {
-                // Link to Switchboard Appearance Wallpaper tab
+                try {
+                    AppInfo.launch_default_for_uri ("settings://desktop/appearance/wallpaper", null);
+                } catch (Error e) {
+                    warning ("Failed to open network settings: %s", e.message);
+                }
             });
 
             display_settings_accellabel = new Granite.AccelLabel (_("Display Settings…"));
@@ -283,7 +287,11 @@ namespace Gala {
             display_settings = new Gtk.MenuItem ();
             display_settings.add (display_settings_accellabel);
             display_settings.activate.connect (() => {
-                // Link to Switchboard Display plug
+                try {
+                    AppInfo.launch_default_for_uri ("settings://display", null);
+                } catch (Error e) {
+                    warning ("Failed to open network settings: %s", e.message);
+                }
             });
 
             system_settings_accellabel = new Granite.AccelLabel (_("System Settings…"));
@@ -291,7 +299,11 @@ namespace Gala {
             system_settings = new Gtk.MenuItem ();
             system_settings.add (system_settings_accellabel);
             system_settings.activate.connect (() => {
-                // Link to Switchboard itself
+                try {
+                    AppInfo.launch_default_for_uri ("settings://", null);
+                } catch (Error e) {
+                    warning ("Failed to open network settings: %s", e.message);
+                }
             });
 
 
--- a/daemon/MenuDaemon.vala
+++ b/daemon/MenuDaemon.vala
@@ -306,10 +306,12 @@ namespace Gala {
                 }
             });
 
+            var separator = new Gtk.Separator (Gtk.Orientation.HORIZONTAL);
 
             desktop_menu = new Gtk.Menu ();
             desktop_menu.append (change_wallpaper);
             desktop_menu.append (display_settings);
+            desktop_menu.append ((Gtk.MenuItem) separator);
             desktop_menu.append (system_settings);
             desktop_menu.show_all ();
         }
--- a/daemon/MenuDaemon.vala
+++ b/daemon/MenuDaemon.vala
@@ -274,7 +274,7 @@ namespace Gala {
                         var message_dialog  = new Granite.MessageDialog.with_image_from_icon_name (
                             "Failed to Open Wallpaper Settings",
                             "The error was: Could not launch Switchboard.",
-                            "dialog-warning",
+                            "dialog-error",
                             Gtk.ButtonsType.CLOSE
                         );
                         message_dialog.run();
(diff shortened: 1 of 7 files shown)


lainsce:
Ready to review! :rocket:

daemon/MenuDaemon.vala
```diff
@@ -258,5 +268,65 @@ namespace Gala {
                 push_in = true;
             }, 3, Gdk.CURRENT_TIME);
         }
+
+        private void init_desktop_menu () {
+            change_wallpaper_accellabel = new Granite.AccelLabel (_("Change Wallpaper…"));
```
danrabbit, line 273: Since none of these have accelerators, you could probably just do `Gtk.MenuItem.MenuItem.with_label`

daemon/MenuDaemon.vala
```diff
@@ -258,5 +268,65 @@ namespace Gala {
                 push_in = true;
             }, 3, Gdk.CURRENT_TIME);
         }
+
+        private void init_desktop_menu () {
```
danrabbit, line 272: Since this is only ever done once, it probably doesn't need to be a separate method

daemon/MenuDaemon.vala
```diff
@@ -48,6 +49,15 @@ namespace Gala {
...
         Gtk.MenuItem close;
 
+        // Desktop Menu
+        private Granite.AccelLabel change_wallpaper_accellabel;
+        private Granite.AccelLabel display_settings_accellabel;
+        private Granite.AccelLabel system_settings_accellabel;
+        Gtk.Menu? desktop_menu = null;
+        Gtk.MenuItem change_wallpaper;
+        Gtk.MenuItem display_settings;
+        Gtk.MenuItem system_settings;
```
danrabbit, line 59: It doesn't appear that these are ever modified outside of the method that constructs them, so they probably don't need to be accessible by the whole class. You could construct them with `var foo`

daemon/MenuDaemon.vala
```diff
@@ -258,5 +268,65 @@ namespace Gala {
...
+        private void init_desktop_menu () {
+            change_wallpaper_accellabel = new Granite.AccelLabel (_("Change Wallpaper…"));
+
+            change_wallpaper = new Gtk.MenuItem ();
+            change_wallpaper.add (change_wallpaper_accellabel);
+            change_wallpaper.activate.connect (() => {
+                try {
+                    AppInfo.launch_default_for_uri ("settings://desktop/appearance/wallpaper", null);
+                } catch (Error e) {
+                    warning ("Failed to open wallpaper settings: %s", e.message);
```
danrabbit, line 281: We should probably throw a dialog if these launches fail and not fail silently

lainsce: What would the corpus be?

daemon/MenuDaemon.vala
lainsce: Solved!

daemon/MenuDaemon.vala
```diff
@@ -258,5 +262,54 @@ namespace Gala {
...
+                        warning ("Failed to open settings: %s", e.message);
+                    }
+                });
+    
+                var separator = new Gtk.Separator (Gtk.Orientation.HORIZONTAL);
+    
+                desktop_menu = new Gtk.Menu ();
+                desktop_menu.append (change_wallpaper);
+                desktop_menu.append (display_settings);
+                desktop_menu.append ((Gtk.MenuItem) separator);
```
danrabbit, line 300: `Gtk.SeparatorMenuItem` exists for this :)

src/Background/BackgroundContainer.vala
```diff
@@ -29,6 +30,13 @@ namespace Gala {
         construct {
             Meta.MonitorManager.@get ().monitors_changed.connect (update);
 
+            reactive = true;
+            button_release_event.connect ((event) => {
+                if (event.button == 3) {
```
danrabbit, line 35: There's a convenient constant for this to make it more clear to read :)
```suggestion
                if (event.button == Gdk.BUTTON_SECONDARY) {
```

daemon/MenuDaemon.vala
lainsce: I did some dialogs

daemon/MenuDaemon.vala
```diff
@@ -258,5 +262,78 @@ namespace Gala {
...
+                var change_wallpaper = new Gtk.MenuItem.with_label (_("Change Wallpaper…"));
+                change_wallpaper.activate.connect (() => {
+                    try {
+                        AppInfo.launch_default_for_uri ("settings://desktop/appearance/wallpaper", null);
+                    } catch (Error e) {
+                        warning ("Failed to open wallpaper settings: %s", e.message);
+                        var message_dialog  = new Granite.MessageDialog.with_image_from_icon_name (
+                            "Failed to Open Wallpaper Settings",
+                            "The error was: Could not launch Switchboard.",
+                            "dialog-warning",
```
danrabbit, line 277: This isn't really a warning right? Should probably be
```suggestion
                            "dialog-error",
```

daemon/MenuDaemon.vala
```diff
@@ -258,5 +262,78 @@ namespace Gala {
...
         }
+
+        public void show_desktop_menu (int x, int y) throws DBusError, IOError {
+            if (desktop_menu == null) {
+                var change_wallpaper = new Gtk.MenuItem.with_label (_("Change Wallpaper…"));
+                change_wallpaper.activate.connect (() => {
+                    try {
+                        AppInfo.launch_default_for_uri ("settings://desktop/appearance/wallpaper", null);
+                    } catch (Error e) {
+                        warning ("Failed to open wallpaper settings: %s", e.message);
```
danrabbit, line 273: We probably don't need this since we throw a dialog now
```suggestion
```

lainsce:
Any last points for review? :)

daemon/MenuDaemon.vala
```diff
@@ -258,5 +262,78 @@ namespace Gala {
...
+                        var message_dialog  = new Granite.MessageDialog.with_image_from_icon_name (
+                            "Failed to Open Wallpaper Settings",
+                            "The error was: Could not launch Switchboard.",
```
danrabbit, line 276: We don't ever use "Switchboard" in the UI, so we probably don't want to reference that here. We also want to try to provide some kind of further information about how to resolve the issue or at least begin investigating it. How about
```suggestion
                            "Unable to open System Settings. A handler for the `settings://` URI scheme must be installed.",
```

daemon/MenuDaemon.vala
```diff
@@ -258,5 +262,78 @@ namespace Gala {
...
+                            "dialog-warning",
+                            Gtk.ButtonsType.CLOSE
+                        );
```
danrabbit, line 279: Since we get some error messaging from GLib if this fails, this would be a good time to include that:
```suggestion
                        );
                        message_dialog.show_error_details (e.message);
```

daemon/MenuDaemon.vala
```diff
@@ -258,5 +262,78 @@ namespace Gala {
...
+                        message_dialog.run();
+                        message_dialog.destroy();
```
danrabbit, line 281: Code style :)
```suggestion
                        message_dialog.run ();
                        message_dialog.destroy ();
```

lainsce:
Should be good to go. :rocket:

PR #880, opened 2020-07-20
GSchema: revert removing notification schema

danrabbit:
So we don't accidentally hose your session if you have this plugin still

### DIFF
--- a/data/gala.gschema.xml
+++ b/data/gala.gschema.xml
@@ -260,6 +260,33 @@
 		</key>
 	</schema>
 
+	<schema path="/org/pantheon/desktop/gala/notifications/" id="org.pantheon.desktop.gala.notifications">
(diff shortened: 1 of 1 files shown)


data/gala.gschema.xml
```diff
@@ -255,6 +260,33 @@
...
+	<schema id="org.pantheon.desktop.gala.notifications.application">
+		<key type="b" name="bubbles">
+			<default>true</default>
+			<summary>Enable bubbles</summary>
+			<description>DEPRECATED: This key is deprecated and ignored.</description>
+		</key>
+		<key type="b" name="sounds">
+			<default>true</default>
+			<summary>Enable sounds</summary>
+						<description>DEPRECATED: This key is deprecated and ignored.</description>
```
cassidyjames, line 281: Accidentally some tabs

PR #881, opened 2020-07-20
Add editorconfig; fix whitespace

cassidyjames:
I noticed we stil had a lot of mixed tabs and spaces throughout the codebase. This PR adds our regular `.editorconfig` file and replaces tab characters with spaces.

It looks massive, but view without whitespace changes and you'll see it's tiny. :)


tintou:
I'd prefer to not touch anything in the .vapi folder as the .vapi files are always using tabs

cassidyjames:
@tintou :+1: reverted those changes

PR #888, opened 2020-07-29
Enable "Alt" + "~" to switch between windows of the same application

### LINKED ISSUE #536: Enable "Alt" + "~" to switch between windows of the same application

andreasfelix:
Fixes #536

Switching between different instances of the same applications via <kbd>Alt</kbd><kbd>Above_tab</kbd> is already implemented in mutter, but is - at least to my understanding - disabled by the following lines:

https://github.com/elementary/gala/blob/517748bcac9484df6a1f03c84f77c9a066ed8793/src/WindowManager.vala#L296-L297

Is their reason behind this?


jeremypw:
@andreasfelix  The source repository appears to be deleted - did you intend to close this PR?

andreasfelix:
@jeremypw, no I just deleted most of my elementary forks when I was added to the Github org. As far as I understand GitHub does some magic and still stores these commits somewhere. I can still fetch the most recent commit with:

```git fetch origin pull/888/head:switch-windows-between-same-application```

jeremypw:
@andreasfelix If this is still active, maybe it would be better to close this and reopen a new one from the new location?

andreasfelix:
> @andreasfelix If this is still active, maybe it would be better to close this and reopen a new one from the new location?

done

PR #890, opened 2020-08-02
Add Blender-like pie menus

andreasfelix:
Many games and  [Blender 2.8](https://developer.blender.org/T56881#:~:text=Blender%202.8%20now%20includes%20pie,key%20to%20set%20the%20viewpoint.) have pie-menus, which are really fast and allow for quick interactions similar to gestures. I think it would be cool if Gala would have them!

I did an initial implementation, which is not perfect, but should be good enough for demonstration purposes. 

###### Tile Windows <kbd>Super</kbd><kbd>Z</kbd>
tiling

*BTW this is also super convenient on multiple monitors. You don't have to drag the window to another monitor, just move the mouse and select the tile-action.*

###### Maximize/Fullscreen/Center/Move <kbd>Super</kbd><kbd>X</kbd>

_(No screen capture)_ 

###### Switch and move windows to Workspaces <kbd>Super</kbd><kbd>C</kbd>

workspaces

######  Move to workspace <kbd>Super</kbd><kbd>Q</kbd>

 _(No screen capture)_

 The screen flickering is only visible in the screen capture. I don't know what exactly causes this problem. I used Cairo to draw these menus, but this is the first time I used it, so it is likely, i have messed something up. Any help is appreciated!

Blender also allows to press a number on the number-pad instead of moving the mouse, which is not implemented yet!


andreasfelix:
**ToDo**
- [ ] Pre-calculated positions so they, do not have to be calculated on every draw
- [ ] Remove flickering in screen captures
- [ ] Better Design ?

fabiozaramella:
Hey, @andreasfelix are you still working on this? 😊

andreasfelix:
@fabiozaramella, in my opinion the implementation is good enough for a prototype :) I think in the future, when Gala will have gestures (which only work on track-pads), pie-menus could be used to make those actions available on a usual mouse. So that both 3 and 4 finger swipes would have an equivalent pie-menu. 

If the elementary team agrees with this, I would make this branch more production ready (but not before next year).

danirabbit:

PR #894, opened 2020-08-03
KeyboardManager: simplify the memory management by not making it a singleton

tintou:


src/KeyboardManager.vala
```diff
@@ -17,41 +17,32 @@
...
-        public static void init (Meta.Display display) {
-            if (instance != null)
-                return;
-
-            instance = new KeyboardManager ();
-
-            display.modifiers_accelerator_activated.connect (instance.handle_modifiers_accelerator_activated);
-        }
+        public weak Meta.Display display { get; construct; }
+        private GLib.Settings settings;
```
jeremypw, line 23: If `schema == null` settings does not get explicitly initialized (see line 37). Should be `private Glib.Settings? settings = null;` ?

tintou: Fields are automatically set to zero at creation time but it might be worth setting it to `GLib.Settings?`

src/KeyboardManager.vala
```diff
@@ -17,43 +17,38 @@
...
-        public static void init (Meta.Display display) {
-            if (instance != null)
-                return;
-
-            instance = new KeyboardManager ();
-
-            display.modifiers_accelerator_activated.connect (instance.handle_modifiers_accelerator_activated);
-        }
+        public weak Meta.Display display { get; construct; }
+        private GLib.Settings? settings;
```
jeremypw, line 23: This may not get initialised explicitly - is it OK to assume it gets automatically initialised to `null`?

jeremypw (commented):
One  inline comment.

I am not familiar with the code or window managers in general but I did not find any regressions in keyboard handling after installing and running `gala --replace` or after rebooting (in VirtualBox).  Except that I am not sure how to confirm that it is the new code that is running - inserted terminal output did not appear - any suggestions?

Does only one WindowManager instance ever get created, in which case there should not be any problem with not ensuring the KeyboardManager is a singleton.

The code looks good.

PR #904, opened 2020-08-25
Add visual indicator for dwell click

### LINKED ISSUE #893: Add Dwell Click
The current interface for this was removed from GSD: https://github.com/GNOME/gnome-settings-daemon/commit/9d5e1d9c0f7b0b702372c47fb81fad7cf3a4085f

And added to GNOME Shell: https://gitlab.gnome.org/GNOME/gnome-shell/merge_requests/474/diffs

davidmhewitt:
Fixes #893 

This uses a hardcoded color (blueberry 300) for now. It may be possible to do something clever and get the accent color from the theme if desired, but I was struggling a bit with figuring out how to get the StyleContext, CssProvider and everything set up to do this.

Note: There's been a PR merged into the a11y indicator that adds a switch to set the necessary dconf key to enable this. But for reference it's `org.gnome.desktop.a11y.mouse.dwell-clicks-enabled`

### DIFF
--- a/src/Widgets/DwellClickTimer.vala
+++ b/src/Widgets/DwellClickTimer.vala
@@ -1,0 +1,129 @@
+//
+//  Copyright 2020 elementary, Inc. (https://elementary.io)
+//
+//  This program is free software: you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License as published by
+//  the Free Software Foundation, either version 3 of the License, or
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//
+
+namespace Gala {
+    public class DwellClickTimer : Clutter.Actor, Clutter.Animatable {
+        private const int WIDTH_PX = 60;
+        private const int HEIGHT_PX = 60;
+
+        private const uint BORDER_WIDTH_PX = 3;
+
+        private const double START_ANGLE = 3 * Math.PI / 2;
+
+        private int scaling_factor = 1;
+
+        private Cogl.Pipeline pipeline;
+        private Clutter.PropertyTransition transition;
+
+        public weak WindowManager wm { get; construct; }
+
+        public double angle { get; set; }
+
+        public DwellClickTimer (WindowManager wm) {
+            Object (wm: wm);
+        }
+
+        construct {
+            visible = false;
+            reactive = false;
+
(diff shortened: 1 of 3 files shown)


src/Widgets/DwellClickTimer.vala
```diff
@@ -0,0 +1,139 @@
...
+            transition.set_animatable (this);
+            transition.set_from_value (START_ANGLE);
+            transition.set_to_value (START_ANGLE + (2 * Math.PI));
+
+            transition.new_frame.connect (() => {
+                queue_redraw ();
+            });
+
+            var rgba = Gdk.RGBA ();
+            rgba.parse (BACKGROUND_COLOR);
```
donadigo, line 60: As for now, it's better to hardcode the RGB values of the color and feed set them directly into the struct, rather than having the class parse the same string everytime on construction.

src/Widgets/DwellClickTimer.vala
```diff
@@ -0,0 +1,139 @@
...
+            cr.fill_preserve ();
+
+            cr.set_line_width (BORDER_WIDTH_PX * scaling_factor);
+            cr.set_source (stroke_color);
+            cr.stroke ();
+
+            var cogl_context = context.get_framebuffer ().get_context ();
+
+            try {
+                var texture = new Cogl.Texture2D.from_data (cogl_context, width, height, Cogl.PixelFormat.BGRA_8888_PRE,
```
donadigo, line 118: Creating a new texture like this should really only happen when the width or height changes, otherwise there's no reason to recreate it everytime on paint.

donadigo (changes requested):
Two diff comments.

src/Widgets/DwellClickTimer.vala
davidmhewitt: This class is only constructed once when Gala starts. I felt it was more beneficial to have a string constant at the top that was easily copied and pasted from a palette etc than it was to need to change 3 numbers somewhere in the file when you need to change the colour. Happy to change it if you feel it's necessary though.

src/Widgets/DwellClickTimer.vala
davidmhewitt: Thanks, yeah, that makes sense. Done!

src/Widgets/DwellClickTimer.vala
danrabbit: I think since we'll want to get the color from the stylesheet anyways that this is probably okay for now

worldofpeace:
I'm using mutter 3.34 and I can no longer compile gala with this PR.
```
../src/Widgets/DwellClickTimer.vala:28.17-28.29: error: The type name `Cogl.Pipeline' could not be found
        private Cogl.Pipeline pipeline;
                ^^^^^^^^^^^^^
../src/Widgets/DwellClickTimer.vala:96.37-96.56: error: The type name `Clutter.PaintContext' could not be found
```

I would think those should be conditionally complied for mutter 3.36, but I don't think `DwellClickTimer.vala` should be compiled at all if you're not using mutter 3.36?

PR #905, opened 2020-08-25
Locate pointer

### LINKED ISSUE #715: Implement reveal pointer
It looks like this was removed from GSD:

https://github.com/GNOME/gnome-settings-daemon/commit/6dff93a0dfa80db3f481eed5d6ed689bf469aa1b

https://gitlab.gnome.org/GNOME/gnome-shell/merge_requests/413

davidmhewitt:
Fixes #715 

Works when `org.gnome.desktop.interface.locate-pointer`  is set. Again, this uses a hardcoded color, but the animation on this one is much more up for discussion I guess.

Maybe we just want an outline shrinking towards the cursor, or something else completely. But this covers the basic implementation so we can discuss potential changes.

### DIFF
--- a/src/Widgets/PointerLocator.vala
+++ b/src/Widgets/PointerLocator.vala
@@ -33,7 +33,7 @@ namespace Gala {
         private GLib.Settings settings;
         private Cogl.Pipeline pipeline;
         private Cairo.Pattern stroke_color;
-        private Cairo.Pattern background_color;
+        private Cairo.Pattern fill_color;
 
         private uint timeout_id;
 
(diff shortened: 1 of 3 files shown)


worldofpeace:
I'm using mutter 3.34 and I can no longer compile gala with this PR. 
```
../src/Widgets/PointerLocator.vala:35.17-35.29: error: The type name `Cogl.Pipeline' could not be found
        private Cogl.Pipeline pipeline;
                ^^^^^^^^^^^^^
../src/Widgets/PointerLocator.vala:70.37-70.56: error: The type name `Clutter.PaintContext' could not be found
```

I would think those should be conditionally complied for mutter 3.36, but I don't think `PointerLocator.vala` should be compiled at all if you're not using mutter 3.36?

PR #907, opened 2020-08-25
Scale hover click timer to cursor size

danrabbit:
Instead of hardcoding the size of the click time, scale it to the cursor size


davidmhewitt (approved):
Derp, that makes way more sense than hard coding it.

If you're happy with the size, I'm happy with the code.

It feels to me like the timer can be obscured by the pointer slightly, especially the taller pointers, like the text entry caret. So whether we wanted to increase it above the base size of the pointer a bit?

danrabbit:
@davidmhewitt good call, how's 1.25 look?

davidmhewitt:
Yep! Looks good

PR #909, opened 2020-08-26
Set accent color in DwellClick and PointerLocator

### LINKED ISSUE #906: Reveal pointer color should match theme accent color
Not sure exactly the easiest/best way to get this but right now we're currently hardcoding blue

danrabbit:
Fixes #906

### DIFF
--- a/src/Widgets/DwellClickTimer.vala
+++ b/src/Widgets/DwellClickTimer.vala
@@ -31,6 +31,7 @@ namespace Gala {
         private Cairo.Pattern fill_color;
         private GLib.Settings interface_settings;
         private Cairo.ImageSurface surface;
+        private Gtk.StyleContext selection_style_context;
 
         public weak WindowManager wm { get; construct; }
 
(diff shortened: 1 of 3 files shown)


donadigo (changes requested):
`paint` is a very frequently called function and there should be as little work done in it as possible. I recommend creating the style context in the constructor and then only querying the color in paint. That way we keep allocations to a minimum and reduce the amount of work that is otherwise constant. 

If creating the context only in the constructor doesn't work here, we could look at other methods of querying for the color.

src/Widgets/DwellClickTimer.vala
```diff
@@ -100,6 +94,27 @@ namespace Gala {
...
+            style_context.set_parent (label_style_context);
+            style_context.set_state (Gtk.StateFlags.SELECTED);
+
+            var rgba = (Gdk.RGBA) style_context.get_property (
+                Gtk.STYLE_PROPERTY_BACKGROUND_COLOR,
+                Gtk.StateFlags.NORMAL
+            );
+
+            stroke_color = new Cairo.Pattern.rgb (rgba.red, rgba.green, rgba.blue);
+            fill_color = new Cairo.Pattern.rgba (rgba.red, rgba.green, rgba.blue, BACKGROUND_OPACITY);
```
tintou, line 116: Just for clarity, why aren't you using rgba.alpha here instead of BACKGROUND_OPACITY ?

danrabbit: Added a comment. It's just too low contrast in many situations. We can predict the background for a label, but not the cursor

danrabbit:
@donadigo good call! Moved everything except grabbing the color to construct

donadigo (commented):
This looks good now. Because the same code now lives in two classes, an ideal solution however would be to create a utility function in `InternalUtils` to initialize the context only once and have it return the RGBA, separating the code that is used for this purpose.

danrabbit:
@donadigo done :)

donadigo (approved):
Looks good to me.

Issue #917, opened 2020-09-09
Notification appear on left when window animations turned off


jeremypw:
I have searched open and closed issues for duplicates.

## Describe the bug
If the window animations are turned off in Settings | Desktop | Appearance notifications appear on the left not withstanding the setting in Settings | Notifications being for them to appear on the right.

### To Reproduce
1. Settings | Desktop | Appearance
2. Turn off window animations
3. Generate a notification with `notify-send` or otherwise.
4. Notification bubble appears on the left.
5. Turn window animations on
6. Generate another notification
7. Bubble appears in the expected position.

### Expected behavior
The notification bubble position does not depend on any setting apart from that in Settings | Notifications.

## Platform Information
Odin on Virtual Box VM
I am using the version from the Odin repository.

davidmhewitt:
As far as I can tell, this is resolved. The linked PR just wasn't marked to close the issue. Please re-open if you can still reproduce this.

PR #920, opened 2020-09-19
Fix notifications being unhandled when animations disabled

donadigo:
This caused #917 and having them completely non functional when animations were disabled.


danrabbit (approved):
Fixes the linked issue and works as expected. Nice job :)

PR #930, opened 2020-10-11
WindowManager: fixed nudge flicker and improved performance

Dreckr:
When scrolling past the first/last workspace, Gala nudges the current workspace to indicate that there's no other workspace in the direction that you are scrolling. This nudge sometimes flickers Gala's background on top of the workspace's clone.

This changes fix this by simplifying the animation. I also replace animating `x` with animating `translation-x`, which improves performance by avoiding any relayout calculations made by Clutter.


danrabbit (approved):
LGTM and fixes the issue. Nice work!

PR #938, opened 2020-10-19
InternalUtils: Prevent X11 specific calls on wayland

davidmhewitt:
This method causes a segfault when trying to use Gala as a wayland compositor.

I'm reasonably sure it's not needed under Wayland, and in the case that it is needed, it looks like it would just break the hot corners until we write an alternative.


tintou (approved):
Wayland is better than this feature

PR #939, opened 2020-10-20
Replace BAMF for icon loading

davidmhewitt:
BAMF is used in Gala to get the icon belonging to a window so it can be shown in the window switcher/multitasking view, etc...

We also relied on the xids of windows. Neither of these things are relevant in the Wayland world.

This implements some Wayland-compatible window matching code based on code in gnome-shell.

As mentioned in a comment below, there are some applications that this can't match an icon for (which is a regression from the current use of BAMF), but we need to weigh up how much effort we put into matching every single edge case:
https://github.com/elementary/gala/pull/939#issuecomment-716567268


jassmith:
How do I merge this 5 years ago?

davidmhewitt:
I think this is getting to the point where it's ready for some testing.

@cassidyjames I noticed you've been playing with some Edge web apps turned into desktop files recently. I think they'd be an interesting use case for this branch to see if Gala can still pick up the correct icons for them in the window switcher and the multitasking view.

I think testing this is just going to involve getting loads of weird and wonderful apps and seeing if any of them don't get icons in the compositor.

tintou:
I'm in favor of merging this early and dealing with regressions later :+1:

davidmhewitt:
So, I've been doing some testing here and there are definitely a lot of non-native apps (packaged as debs) that this can't pick up icons for. Notably these are Java or wxWidgets apps (where have we heard that before?!) and a few KDE apps. Two examples are PlayOnLinux and Netbeans

GNOME has an advantage here in that it's often the shell itself that launches applications, so it can keep a record of which pids belong to which desktop files. Since it's wingpanel (or the dock) that launches the applications here, gala doesn't have oversight of what the pids are and can't match them to desktop files in these edge cases.

However, if the same applications that can't be matched are packaged as a snap/flatpak, then gala can get an icon because we can use the sandbox application id.

As far as I can tell so far, this is working well for GTK apps and a number of other popular apps (Firefox, LibreOffice, etc...), so I guess we just need to make a call on how much we worry about the other non-native cases. In theory, we can make wingpanel and plank notify gala when it launches an app to pick up some of these other edge cases.

danrabbit:
My personal feeling is that our recommended method to sideload apps is via Flatpak and we don't provide any tools to load apps as a Deb, so users should expect they're going outside of what the system provides for. It's probably not worth the additional effort when the majority of apps will work as expected

worldofpeace:
>  we don't provide any tools to load apps as a Deb,

elementary OS doesn't have apt or dpkg in PATH? Or do you mean graphical tools?

davidmhewitt:
When I say packaged as deb, I mean stuff from the Ubuntu repos too. So I can install playonlinux from appcenter and not get an icon in multitasking view.

cassidyjames:
If anything, this seems more reliable in my use than the matching in Plank, for the most part. The one caveat I have is Flatpaked apps that provide multiple .desktop files, like LibreOffice; LibreOffice Writer correctly shows in Plank as the Writer icon, but it shows as the generic LibreOffice icon in Gala now.

That seems like an edge case and not a huge deal to me personally, as imho LibreOffice apps should be bundled separately in Flatpaks, but that's a whole different issue. That's at least an example of an app where users will see a visible change.

cassidyjames:
As a note for Edge web apps, they work as expected in this branch. Something else to test would be WebPin where it creates a new .desktop file for web apps that users pick inside the app itself.

cassidyjames:
Granite Demo is also not showing the icon correctly in Gala with this branch. Is there something we can/should be doing differently in its .desktop file or in the app itself to resolve that?

davidmhewitt:
The GApplication ID has to either match the filename of the desktop file. Or, the desktop file needs to provide a StartupWMClass key that contains the GApplication ID.

cassidyjames:
@davidmhewitt hm, it _looks_ to me like that should already be the case in Granite Demo: .desktop file is [`io.elementary.granite.demo.desktop`](https://github.com/elementary/granite/blob/ee33b5c77f3a392534594978c009614bc6ac2546/demo/io.elementary.granite.demo.desktop) and [application ID is `io.elementary.granite.demo`](https://github.com/elementary/granite/blob/ee33b5c77f3a392534594978c009614bc6ac2546/demo/GraniteDemo.vala#L29
). Could it be that it's set to `NoDisplay=true` or something?

Regardless, I don't consider this a blocker. I'm in favor of getting this rolling sooner rather than later.

davidmhewitt:
Hm, interesting. I don't think NoDisplay should affect it. I'll look into that and hopefully someone can do a code review in the mean time.

stsdc:
Would an api be available for 3rd party apps? I did use bamf to display apps (basically their window title and icon) in process view of Monitor and separate them from rest of the background processes.

lib/Utils.vala
```diff
@@ -17,147 +17,234 @@
 
 namespace Gala {
     public class Utils {
-        // Cache xid:pixbuf and icon:pixbuf pairs to provide a faster way aquiring icons
-        static HashTable<string, Gdk.Pixbuf> xid_pixbuf_cache;
-        static HashTable<string, Gdk.Pixbuf> icon_pixbuf_cache;
-        static uint cache_clear_timeout = 0;
+        private class CachedIcon {
```
donadigo, line 20: I think a struct or even a `[Compact]` class would be a better fit for this kind of container.

donadigo (changes requested):
Only one comment here, good job.

davidmhewitt:
@cassidyjames So I looked at the granite-demo issue. It turns out that Mutter doesn't do anything smart under X11 and can't actually get the GApplication ID, but it can under Wayland, so the only thing we have left to try and match here is the "WM_CLASS".

GTK apps by default set their "WM_CLASS" to the name of the executable. So given that `granite-demo` does not match the desktop file, there are 3 options:

- Rename the executable to `io.elementary.granite.demo`
- Rename the desktop file to `granite-demo.desktop`
- Add `StartupWMClass=granite-demo` to the desktop file

So we need to watch out for any cases where the desktop file doesn't match the exec name.

cassidyjames:
@davidmhewitt dope. Thanks for looking at that. It sounds to me like adding `StartupWMClass=granite-demo` would be the most sane because it's a developer tool designed to be launched from the Terminal, so `granite-demo` is super nice to have for the executable.

donadigo (approved):
I think at some point we have to start considering making this a public API that gala exposes that you can query for windows, otherwise we're left with Plank being completely incompatible with Wayland among other things. This looks good for the time being.

jassmith:
Congratulations! :)

PR #949, opened 2020-10-27
meson: Compare libmutter_dep version instead of checking for mutter336_dep

tintou:
Using libmutter_dep allows to have a future-proof checking for this.


davidmhewitt (approved):
This makes a lot more sense to me, but we can wait to have confirmation that this fixes mutter-7 if you like.

decathorpe:
LGTM

PR #953, opened 2020-10-31
Rename 'Minimize' to 'Hide'

jlnr:
Rationale:

* There is no "minimized" version of a window anywhere on the screen, so the name as I understand it doesn't make sense
* The new name also matches the existing meta+H shortcut (and hopefully makes it easier to memorize)

Bildschirmfoto 2020-10-21 um 13 07 05

Instead of only changing the localizable string, I've also renamed all occurences of the word "minimized" that weren't inherited from base classes. Are there any third-party projects that might be affected by this? Should I bump an so version number or similar to indicate a breaking change; or should I revert some of the renames?

I've kept the internal name of the "minimize" shortcut to avoid breaking existing setups.

### DIFF
--- a/daemon/MenuDaemon.vala
+++ b/daemon/MenuDaemon.vala
@@ -32,14 +32,14 @@ namespace Gala {
         // Window Menu
         private Granite.AccelLabel always_on_top_accellabel;
         private Granite.AccelLabel close_accellabel;
-        private Granite.AccelLabel minimize_accellabel;
+        private Granite.AccelLabel hide_accellabel;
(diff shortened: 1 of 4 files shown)


jlnr:
Corresponding keyboard plug update: https://github.com/elementary/switchboard-plug-keyboard/pull/332

Because Switchboard has already dropped support for Juno (through libhandy-1-dev), I guess that this should also be merged _after_ the gala/juno branch has been created to keep things consistent.

danrabbit (approved):
LFG

PR #958, opened 2020-11-07
Multi-touch support: Touchégg integration + Multitasking View 1:1 gesture

JoseExposito:
Hi!

About a month ago, we discussed in the #gala channel about the best way of integrating multi-touch support in Gala but we didn't reach a conclusion. I have decided to create this proof of concept so we can continue the discussion with some real code to play with.

### How it works:

Touchpad and touchscreen input is handled by libinput. I have done quite a bit of work on this project:
https://github.com/JoseExposito/touchegg
This pull request add some glue code to connect to the Touchégg daemon and get gesture events.

In order to test it, install Touchégg and run Gala as usual. You'll see the gesture information on the terminal.

### Why a daemon?

On Wayland the compositor has access to libinput. However, on X11 it is only possible to access libinput if the user running the WM (i.e. your user) is part of the `input` group. This is not safe as any process could access your all your input: keyboard, mouse, touchpad, etc.

I solved this issue by creating a `systemd` daemon that exposes only the required information:
https://github.com/JoseExposito/touchegg/blob/master/installation/touchegg.service

### What is included out of the box?

The daemon does a lot more than just proxy gesture events through an Unix socket. Here are some of those extra features:

 - Gestures work in small an big touchpads. This is something GNOME is missing at the moment and it makes difficult to use gestures if your touchpad size is not similar to the values hardcoded: https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/master/js/ui/swipeTracker.js#L9
 - Touchscreen support. libinput does not support touchscreen gestures out of the box, you have to implement them your self from raw data. The daemon abstracts this and exposes touchscreen gestures using a single interface.
 - Gestures doesn't trigger until a threshold is reached. Otherwise it is just too easy to trigger gestures just by resting your fingers on the touchpad/touchscreen. This threshold is calculated automatically to the size of your touchpad/touchscreen.
 - Low memory and CPU usage (daemon on the top, client on the bottom)
resource_usage
 - Allow to include features in Gala while falling back to Touchégg in the meanwhile. Of course this is your decision, but it is possible to, for example, handle multitasking view in Gala with a smooth animation (we need this :smile:) and fallback to Touchégg (client) to for example, maximize windows while it gets ported to Gala.

### What is missing?

This is a work in progress pull request, so a lot is missing:

 - [X] Connect the signals emitted by the `GestureManager` and do something useful with them
 - [ ] ~~System settings. It'll be nice to be able to configure some stuff. For example, 4 fingers swipe to switch workspaces is fine on big touchpads, but it feels awkward on small touchpads.~~ Fix it here: https://github.com/elementary/switchboard-plug-mouse-touchpad/issues/16
 - [X] On the Touchégg daemon side, I need to manually set numbers to the enums used in Gala to ensure the protocol is not break accidentally in an update. Fixed on this commit: https://github.com/JoseExposito/touchegg/commit/881bbcd4b3ca0424f53f18f955007c42142546ab
 - [x] If you decide to also run Touchégg client for missing features, this needs to be done before elementary OS 6 gets released:
   https://github.com/JoseExposito/touchegg/issues/360
 - [X] Follow natural scrolling. Gala will need access to this property as well. The daemon is the one that knows if the gesture was done in a touchscreen (always natural scrolling) or in a touchpad (configurable). https://github.com/JoseExposito/touchegg/issues/382

### :warning: Gala noob ahead :warning:

This is the first time I write Vala, so do not hesitate to suggest changes, I'll use your help.


Dirli:
I'll assume that it would look a bit neater as a plugin.

hanaral:
I think it would be optimal to have a stripped down nameless version of touchegg installed by default, since then it would actually create a push to get it as seamless as possible. There could be a global toggle to enable/disable it, but I'm on the side that thinks that this is something that should have been there to begin with. Users _could_ also customise whether its 3 or 4 fingers, but I think it should just be 4.

JoseExposito:
Ok, this is ready!

What is included:

- Integration with Touchégg.
- 1:1 multitasking view gesture, as this was the only one without an animation. Swipe up with 3 or 4 fingers to open it and down with 3 or 4 fingers to close it.
- The multitasking view is not open/close if the gesture doesn't reach the 20% threshold.

Let me explain the changes I made so the code review is easier :wink: 

- As suggested by @Dirli, I integrated Touchégg as a plugin. You can find it in the `plugins/touchegg` folder.
  There is not much going on there, just the connection with the daemon and a `TODO` for my future self to fix https://github.com/elementary/switchboard-plug-mouse-touchpad/issues/16 as my next task.
- The plugin needs to access `wm.workspace_view`, so I exposed that. Let me now if that's a problem.
- I splitted the current animations in 3 parts: being, update, end.
- I created a class named `GestureAnimationDirector`. This class is used to start/update/finish the animation using the hints passed by the plugin.
- I'm not quiet happy having to pass the `GestureAnimationDirector` a couple of levels down to the `WindowClone`, but I didn't find a better solution. I'm open to suggestions and refactors.

And finally, this is how it looks: 

multitasking_view_touchegg

plugins/touchegg/Client.vala
```diff
@@ -0,0 +1,175 @@
...
+
+        /**
+         * Maximum number of reconnection attemps to the daemon.
+         */
+        private const int MAX_RECONNECTION_ATTEMPS = 5;
+
+        /**
+         * Socket used to connect to the daemon.
+         */
+        private Socket socket = null;
```
donadigo, line 61: The member should be marked as nullable.
```suggestion
        private Socket? socket = null;
```

plugins/touchegg/Main.vala
```diff
@@ -0,0 +1,103 @@
...
+            hints.insert ("cancel_action", new Variant.boolean (gesture.percentage < SUCCEESS_THRESHOLD));
+        }
+
+        return hints;
+    }
+
+    // TODO (José Expósito) It'll be nice to be able to configure this from Switchboard
+    // https://github.com/elementary/switchboard-plug-mouse-touchpad/issues/16
+
+    private bool is_open_workspace_gesture (Gesture gesture) {
```
donadigo, line 81: This simply can be a `static` method or part of the `Gesture` class.

plugins/touchegg/Main.vala
```diff
@@ -0,0 +1,103 @@
...
+        return gesture.type == GestureType.SWIPE
+            && gesture.direction == GestureDirection.UP
+            && (gesture.fingers == 3 || gesture.fingers == 4);
+    }
+
+    private bool is_close_workspace_gesture (Gesture gesture) {
```
donadigo, line 87: This method as well.

src/GestureAnimationDirector.vala
```diff
@@ -0,0 +1,53 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+ namespace Gala {
+    public class GestureAnimationDirector {
```
donadigo, line 20: This class should inherit from `Object`.
```suggestion
    public class GestureAnimationDirector : Object {
```

src/Widgets/MultitaskingView.vala
```diff
@@ -27,6 +27,9 @@ namespace Gala {
     public class MultitaskingView : Actor, ActivatableComponent {
         public const int ANIMATION_DURATION = 250;
         public const AnimationMode ANIMATION_MODE = AnimationMode.EASE_OUT_QUAD;
+
+        private GestureAnimationDirector gesture_animation_director = null;
```
donadigo, line 31: This should be marked nullable.
```suggestion
        private GestureAnimationDirector gesture_animation_director? = null;
```

src/Widgets/MultitaskingView.vala
```diff
@@ -544,28 +547,48 @@ namespace Gala {
...
          */
         public void open (HashTable<string,Variant>? hints = null) {
-            if (opened)
-                return;
+            bool manual_animation = hints != null && hints.get ("manual_animation").get_boolean ();
+
+            if (!opened) {
+                if (manual_animation) {
+                    debug ("Starting MultitaskingView manual open animation");
+                    this.gesture_animation_director = new GestureAnimationDirector();
```
donadigo, line 555: I'm not sure why is there a need to create a new instance of this class everytime open/close is called here? If the intention is to manage some kind of state, then perhaps it would be better to initialize the director only once and then store the needed flag in it?

src/Widgets/MultitaskingView.vala
```diff
@@ -626,95 +649,177 @@ namespace Gala {
...
                 unowned WorkspaceClone workspace = (WorkspaceClone) child;
                 if (opening)
-                    workspace.open ();
+                    workspace.open (this.gesture_animation_director, is_cancel_animation);
                 else
-                    workspace.close ();
+                    workspace.close (this.gesture_animation_director);
+            }
+
+            if (opening)
```
donadigo, line 657: Use opening and closing brackets, even for one-line statements.

src/Widgets/MultitaskingView.vala
```diff
@@ -626,95 +649,177 @@ namespace Gala {
...
+                    if (cancel_action) {
+                        toggle (true);
+                    }
+
+                    return false;
+                });
+            };
+
+            if (this.gesture_animation_director == null) {
+                on_animation_end(100, false);
```
donadigo, line 693: Spaces between braces.

src/Widgets/MultitaskingView.vala
```diff
@@ -626,95 +649,177 @@ namespace Gala {
...
+            } else {
+                this.gesture_animation_director.on_animation_end.connect((percentage, cancel_action) => {
+                    on_animation_end(percentage, cancel_action);
```
donadigo, line 696: Here as well.

plugins/touchegg/Client.vala
```diff
@@ -0,0 +1,175 @@
...
+        public int fingers;
+        public uint64 elapsed_time;
+        public DeviceType performed_on_device_type;
+    }
+
+    /**
+     * This class connects to the Touchégg daemon to receive touch events.
+     * See: https://github.com/JoseExposito/touchegg
+     */
+    public class Client {
```
donadigo, line 48: This class should inherit from `Object`.
```suggestion
    public class Client : Object {
```

plugins/touchegg/Client.vala
```diff
@@ -0,0 +1,175 @@
...
+
+        /**
+         * Start receiving gestures.
+         */
+        public void run () throws IOError {
+            new Thread<void*> (null, this.recive_events);
+        }
+
+        private void* recive_events () {
+            uint8[] event_buffer = new uint8[sizeof (GestureEvent)];
```
donadigo, line 82: I don't think we need to allocate this buffer using `new`, I think there are better ways to do this (not sure which will compile though):
1. `uint8[sizeof(GestureEvent)] event_buffer;`
2. `socket.receive ((uint8[])event);`
3. If none of this works, I'm pretty sure you can allocate this buffer in the constructor and reuse it here, simply zeroing it before calling receive.

donadigo (changes requested):
I haven't gotten into the depth of this PR at all, but there's some starting comments while you're working on it.
Here's some tips on code style:
1. Only use `this` if it's actually required. In most places it isn't and we don't use it in most of the codebase.
2. We use brackets for all one-line ifs.
3. We use spaces between braces like `method_here (some_arg)`.

plugins/touchegg/Main.vala
JoseExposito: This solution is temporary, as mentioned in the `TODO` I'll add configuration options for this and I'll move this 2 methods to a different class.

plugins/touchegg/Client.vala
JoseExposito: The buffer only gets allocated once because on the next line the client starts its event loop.
Is Vala really slow allocating memory on the heap? This is the first time I write Vala so I may not know the advantage of using memory in the stack vs the heap.

hanaral:
@JoseExposito Not sure if this is intentional, but 3 finger tapping an app in the overview seems to kill the workspace. If it isn't a bug please consider changing it, since its a nightmare on less accurate trackpads that don't even have real multitouch (trackpads that detect the _location_ of 2 fingers but can only detect the _existance_ of 3 or more)

dar5hak:
@JoseExposito For Vala code style issues, there's [vala-lint](https://github.com/vala-lang/vala-lint). It cannot yet format the code for you, but it will list the issues so that you can fix them manually.

[Uncrustify](https://github.com/uncrustify/uncrustify) can format your code, but it's a bit clunkier and needs to be configured for elementary OS code style. (I've been meaning to make such a config and share the file on GitHub, but I'm lazy.)

JoseExposito:
@hanaral 3 finger tap is equivalent to mouse wheel click. The default behaviour in the multitasking view is to close the app you middle clicked on.

@dar5hak cool!! I'll have a look to those tools. This kind of thinks improve the developer experience a lot.

JoseExposito:
@donadigo as suggested by @dar5hak I used vala-lint to fix some style issues https://github.com/elementary/gala/pull/972/commits/0ab5518b53f12a6bbafd2da86cc55077366cdf5b

I made the changes in my other PR, because it includes this code plus some extras. Just let me know if you want me to cherry pick them to this PR as well, or if reviewing the commit works for you.

plugins/touchegg/Client.vala
```diff
@@ -0,0 +1,175 @@
...
+
+                    if (this.event != null 
+                            && this.event.event_type != GestureEventType.UNKNOWN 
+                            && this.event.event_type != GestureEventType.END) {
+                        this.event.event_type = GestureEventType.END;
+                        this.emit_event (this.event);
+                    }
+
+                    if (this.socket != null) {
+                        this.socket.close ();
```
danrabbit, line 135: Getting a warning here about unhandled GLib.Error

plugins/touchegg/Main.vala
```diff
@@ -0,0 +1,103 @@
...
+        }));
+        this.client.on_gesture_update.connect ((gesture) => Idle.add (() => {
+            this.on_handle_gesture (gesture, "update");
+            return false;
+        }));
+        this.client.on_gesture_end.connect ((gesture) => Idle.add (() => {
+            this.on_handle_gesture (gesture, "end");
+            return false;
+        }));
+        this.client.run ();
```
danrabbit, line 44: Warning here about unhandled GLib.IOError

src/Widgets/WindowCloneContainer.vala
```diff
@@ -361,24 +361,26 @@ namespace Gala {
 
             // make sure our windows are where they belong in case they were moved
             // while were closed.
-            foreach (var window in get_children ())
-                ((WindowClone) window).transition_to_original_state (false);
-
-            reflow ();
+            if (!is_cancel_animation) {
+                foreach (var window in get_children ())
+                    ((WindowClone) window).transition_to_original_state (false);
```
danrabbit, line 366: elementary code style is to use brackets even for one-liners. It looks like this was already incorrect before you got here, but if we're touching this line anyways might as well fix it

```suggestion
                foreach (var window in get_children ()) {
                    ((WindowClone) window).transition_to_original_state (false);
                }
```

src/Widgets/WindowClone.vala
```diff
@@ -295,53 +295,149 @@ namespace Gala {
...
+
+                save_easing_state ();
+                set_easing_mode (MultitaskingView.ANIMATION_MODE);
+                set_easing_duration (animate ? MultitaskingView.ANIMATION_DURATION : 0);
+    
+                set_position (target_x, target_y);
+                set_size (outer_rect.width, outer_rect.height);
+    
+                if (should_fade ())
+                    opacity = 0;
```
danrabbit, line 339: Needs brackets:
```suggestion
                if (should_fade ()) {
                    opacity = 0;
                }
```

src/Widgets/WindowClone.vala
```diff
@@ -295,53 +295,149 @@ namespace Gala {
...
+
+                restore_easing_state ();
+    
+                if (animate)
+                    toggle_shadow (false);
```
danrabbit, line 344:
```suggestion
                if (animate) {
                    toggle_shadow (false);
                }
```

src/GestureAnimationDirector.vala
```diff
@@ -0,0 +1,53 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+ namespace Gala {
+    public class GestureAnimationDirector : Object {
```
danrabbit, line 20: It's not really necessary to do this style of namespacing. If we're creating a new class, might as well do:
```suggestion
public class Gala.GestureAnimationDirector : Object {
```

danrabbit (changes requested):
Left a few code style comments and there are a couple of unhandled errors :)

JoseExposito:
Thanks for the review Daniel, I'll have a look to the warnings tomorrow :D

JoseExposito:
I fixed the compilation warnings @danrabbit , I'll merge the changes in my other PRs so the review is easier.

src/GestureAnimationDirector.vala
```diff
@@ -0,0 +1,51 @@
...
+
+        switch (event) {
+            case "begin":
+                this.on_animation_begin (percentage);
+                break;
+            case "update":
+                this.on_animation_update (percentage);
+                break;
+            case "end":
+            default: {
```
donadigo, line 40: Not sure about this brace here, we don't use these in cases like this.

src/Widgets/WindowClone.vala
```diff
@@ -295,53 +295,151 @@ namespace Gala {
...
+                window_icon.set_position ((outer_rect.width - WINDOW_ICON_SIZE) / 2, outer_rect.height - (WINDOW_ICON_SIZE * scale_factor) * 0.75f);
+                window_icon.opacity = 0;
+                close_button.opacity = 0;
+            };
+
+            if (gesture_animation_director == null) {
+                on_animation_begin (0);
+                on_animation_end (100, false);
+            } else {
+                gesture_animation_director.on_animation_begin.connect ((percentage) => {
```
donadigo, line 357: I feel like these could be put on one line to keep it clear:
`gesture_animation_director.on_animation_begin.connect ((percentage) => on_animation_begin (percentage))`

or simply
`gesture_animation_director.on_animation_begin.connect (on_animation_begin);`
if that works.

src/Widgets/WindowClone.vala
```diff
@@ -295,53 +295,151 @@ namespace Gala {
...
+                    restore_easing_state ();
+                }
+            };
+
+            if (gesture_animation_director == null) {
+                on_animation_begin (0);
+                on_animation_end (100, false);
+            } else {
+                gesture_animation_director.on_animation_begin.connect ((percentage) => {
+                    on_animation_begin (percentage);
```
donadigo, line 435: Same here.

src/Widgets/WindowClone.vala
```diff
@@ -295,53 +295,151 @@ namespace Gala {
...
+            var target_y = outer_rect.y - offset_y;
 
-            if (should_fade ())
-                opacity = 0;
+            GestureAnimationDirector.OnBegin on_animation_begin = () => {
+                window_icon.set_easing_duration (0);
+            };
 
-            restore_easing_state ();
+            GestureAnimationDirector.OnUpdate on_animation_update = (percentage) => {
```
donadigo, line 310: There's a lot of code repetition in these lambdas. Perhaps there's a way to put these in a common function?

donadigo (changes requested):
Here are a couple of comments again.
My comments are more suggestive than hard commands, simply because the PR is quite large to review and I'd like to get these right before merging.

I feel like a lot of the code was changed simply to account for the fact that `GestureAnimationDirector` exists and a lot of methods now require it as an argument, which I perceive as a bad thing. 

How about making `GestureAnimationDirector` a singleton class (with a `get_default ()` method that can be used in each component's implementation)? This would revert some gimnastics in the code and be a little easier to read overall. 

The code style issues with using `this` still stand, as well as fixing // TODO comments.
As such I will need another review from someone else to make sure this doesn't become too complicated to maintain in the future.

plugins/touchegg/Client.vala
```diff
@@ -0,0 +1,184 @@
...
+        private GestureEvent *event = null;
+
+        /**
+         * Start receiving gestures.
+         */
+        public void run () throws IOError {
+            new Thread<void*> (null, recive_events);
+        }
+
+        private void* recive_events () {
```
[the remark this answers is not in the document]
JoseExposito: What error is it throwing? I don't see a warning in the console, but I may be missing it

plugins/touchegg/Client.vala
```diff
@@ -0,0 +1,184 @@
...
+
+            return null;
+        }
+
+        private void handle_disconnection () {
+            reconnection_attemps++;
+
+            if (event != null
+                    && event.event_type != GestureEventType.UNKNOWN
+                    && event.event_type != GestureEventType.END) {
```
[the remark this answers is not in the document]
JoseExposito: Do you mean...
(1)
```
if (event != null &&
```

or (2)
```
if (event != null
    && ....
```

It looks like (2) is used in the code.

src/GestureAnimationDirector.vala
JoseExposito: That'd not compile in C/C++, it seem like it is not required in Vala, removing it!

src/Widgets/WindowClone.vala
JoseExposito: I moved the animation value calculation to a common method, but I don't think I can move more code to a common place because each lambda needs to animate different properties in different objects. 
Could you provide an example, please?

src/Widgets/WindowClone.vala
JoseExposito: The second option prints a warning:
```
 warning: Connecting delegates to signals is experimental
```

And the first one will pass the 120 characters limit. I like the first option though, but I don't know how hard is the 120 characters limit....

danrabbit: @JoseExposito not that hard :) Clarity is what's important here

JoseExposito: Good to know. I mean, it makes a lot of sense, specially now that monitors are quiet big.

I'm going to close this PR, as the other one includes every change and the code is cleaner.

Btw, I changed this code to use a simpler method:
https://github.com/elementary/gala/pull/983/files#diff-82d4c23d6a3129070cf40cd4e43dbb42eb991ab0fe7d1bd508c7ee41a793fd60R702

JoseExposito:
Closing this pull request, changes are included in https://github.com/elementary/gala/pull/983

PR #959, opened 2020-11-08
pip: Use DragDropAction for move

davidmhewitt:
`Clutter.DragAction` is gone in newer versions of mutter so we need to stop using it. I've replaced one instance (used for moving the PiP window with dragging) with Gala's own `DragDropAction`.

There's another instance of `Clutter.DragAction` being used for resizing the PiP window that I'll follow up on in another PR. But this is a good start that brings about a nice code simplification.


tintou (approved):
I love that diff

PR #960, opened 2020-11-08
use accent color for tiling preview

andreasfelix:
I set the `tiling_preview` actor to `null` when the preview closes so that accent color is updated every time the preview is opened. Is there a more elegant way?


danrabbit (changes requested):
Instead of setting the preview to null, it seems to work fine if you set the background color outside of `if (tile_preview == null) {}` (at line `1077`/`1071`).

Otherwise, this works as expected. Thanks for fixing this!

andreasfelix:
@danrabbit Yes, this would also work, but I assumed that `InternalUtils.get_theme_accent_color ()` is a relatively "expensive" operation and should therefore only be called once per tiling action. Setting it outside of `if (tile_preview == null)` would trigger it hundreds of times (depending on the users pointer movement) for each tiling action.

danrabbit (approved):
This works as expected so we can always revisit if it turns out it's expensive :)

PR #962, opened 2020-11-09
Gettext: Checkout the repository using the elementaryBot token

tintou:


davidmhewitt:
I think something else is wrong here because the gettext action already sets the token passed via env var on the remote:
https://github.com/elementary/actions/blob/master/gettext-template/entrypoint.sh#L35

tintou:
I've applied this to calendar and it worked there https://github.com/elementary/calendar/commits/master
the GitHub checkoutv2 action is probably setting the credentials using a different method

davidmhewitt:
Had a look into that and opened an issue here:
https://github.com/elementary/actions/issues/55

I _think_ that should work, but I haven't tested it anywhere. But hopefully it's better to fix it there rather than having to go and make this change in every repo.

jeremypw:
@tintou Is this PR still potentially useful or will the issue be fixed upstream?

PR #963, opened 2020-11-10
Introduce a Tiling Mode

andreasfelix:
#### Behavior

1. Similar to other tiling window managers a *Tiling-Mode*  can be activated by pressing <kbd>Super</kbd><kbd>Return</kbd>: This will show a tile preview for the current window, which can then be moved using <kbd>↑</kbd><kbd>←</kbd><kbd>↓</kbd><kbd>→</kbd>, <kbd>w</kbd><kbd>a</kbd><kbd>s</kbd><kbd>d</kbd>, <kbd>h</kbd><kbd>j</kbd><kbd>k</kbd><kbd>l</kbd> or the mouse. Pressing <kbd>Return</kbd> or <kbd>space</kbd> will confirm the current preview and will tile the window accordingly. (Windows can also be moved between multiple monitors)
2. In the future this mode could be complement with a *Window-Selection-Mode*: Here windows can be selected using the arrow keys (similar to the Multitasking view).
3. The grid can be changed by pressing the number keys when in *Tiling-Mode*. For example pressing <kbd>3</kbd><kbd>2</kbd> will set a 3 by 2 grid.
4. Inspired by the [latest blog post](https://blog.elementary.io/dark-style-progress/) I also added optional gaps, which can be toggled by pressing <kbd>g</kbd> when in tiling mode.
5. The <kbd>Drag</kbd><kbd>Ctrl</kbd> tile action of #595 is still available. But I consider to remove it as it just as convenient to press the <kbd>Super</kbd><kbd>Return</kbd> shortcut.

### DIFF
--- a/data/gala.gschema.xml
+++ b/data/gala.gschema.xml
@@ -105,6 +105,18 @@
 			<summary>Automatically move fullscreened windows to a new workspace</summary>
 			<description></description>
 		</key>
+    <key type="b" name="experimental-tiling-mode">
+      <!-- TODO: set to false before merging -->
+			<default>true</default> 
+			<summary>Toggle experimental tiling mode.</summary>
+			<description></description>
+		</key>
+    <!-- TODO: remove before merging -->
+    <key type="b" name="experimental-tiling-mode-shrink">
+			<default>false</default> 
+			<summary>Whether or not the window shrunk when entering tiling mode.</summary>
+			<description></description>
+		</key>
   </schema>
 
   <schema path="/org/pantheon/desktop/gala/keybindings/" id="org.pantheon.desktop.gala.keybindings">
@@ -145,6 +157,12 @@
     <key type="as" name="expose-all-windows">
       <default><![CDATA[['<Super>a']]]></default>
       <summary>Shortcut to open the window overview for all windows</summary>
+      <description></description>
+    </key>
+    <key type="as" name="tiling-mode">
+      <!-- <default><![CDATA[['<Control><Super>Left', '<Control><Super>Down', '<Control><Super>Up', '<Control><Super>Right']]]></default> -->
+      <default><![CDATA[['<Super>Return']]]></default>
+      <summary>Show pie menu.</summary>
       <description></description>
     </key>
     <key type="as" name="cycle-workspaces-next">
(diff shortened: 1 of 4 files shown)


ToxicDragon:
Please also add the pictures if you were to delete #595

jeremypw:
As the UX team is concerned that this is kind of a hack which is not true tiling, because Mutter does support tiling natively I am going to convert this to draft until the future of tiling in Mutter is clarified as until that point there is little point in reviewing this code.

andreasfelix:
The "hacky" part of this PR (the `calculate_tile_rect` function) is only 11 lines of code. Once mutter exposes an API for window tiling it should be pretty straightforward to replace this. IMO the limitation of the mutter API shouldn't be the reason to reject this PR. Of course there could be other valid reasons like that a tiling-mode could incoherent with the existing ux of elementary or something else. But as far as I see it, the amount of technical debt this PR introduces is really minimal.

andreasfelix:
Another solution could be just implement "real tiling" in gala itself. A tiling API seems very high level to me, so might not even a good idea to implement something like this mutter. IMO core libraries like mutter should only do the heavy lifting and expose composable low level APIs. There are different way to implement window tiling and implementing it in mutter takes some flexibility from the window managers built on top of it.

jeremypw:
Well, you need to get the UX team on board with the approach as this obviously has a major impact on UX.

jeremypw:
Its probably not a good time in the cycle to introduce a major change.

andreasfelix:
Agree, I didn't meant to push this in before the OS 6 release. Just in general: IMO when there are different ways to do something it shouldn't be implemented in the underlying library but in the application itself.

ToxicDragon:
See also the Microsoft Powertoys FancyZones tool that makes it possible to configure a custom grid for tiling (also per monitor). Something similar were really great for your presented tiling mode solution aswell! 

https://docs.microsoft.com/en-us/windows/powertoys/fancyzones

ToxicDragon:
@andreasfelix is there a way to try your solution e.g. as an application on e.g. elemtsry OS 6? I'd really like to try it and give feedback. I have a 4K OLED 48" screen and therefore a custom grid layout would really be good for work productivity.

PR #965, opened 2020-11-12
Code Style: Initialize Objects with Properties

andreasfelix:
Some files do not follow the [recommend elementary code style for initializing objects with properties](https://docs.elementary.io/develop/writing-apps/code-style#initialize-objects-with-properites).

I wrote a small Python script which finds all these cases and replaces them with the recommended code style. It could also be used for other repos.

##### Usage

`python codestyle.py <path/to/file.vala>` or `python codestyle.py <path/to/repo>`

<details><summary>codestyle.py</summary>
<p>

#### 
```python
from operator import itemgetter
from pathlib import Path
import regex

pattern = regex.compile(
    r"(?P<space>[ \t]*)(?P<init>[\w,.]+ (?P<name>\w+) = new [\w,.]+ \([^;]+);\n"
    r"(?:[ \t]*(?P=name).(?P<props>\w+\s?=\s?[^;]+);\n)+"
)


def replace(match):
    capturesdict = match.capturesdict()
    (space,), (init,), props = itemgetter("space", "init", "props")(capturesdict)
    props = ",\n".join(f"    {space}{prop}" for prop in props)
    return f"{space}{init} {{\n{props}\n{space}}};\n"


def cli():
    import sys

    if (path := Path(sys.argv[-1])).is_file() and path.suffix == ".vala":
        paths = [path]
    elif path.is_dir():
        paths = path.rglob("*.vala")
    else:
        print(f"'{path}' is not a valid vala file!")
        exit(-1)

    for path in paths:
        path.write_text(pattern.sub(replace, path.read_text()))


if __name__ == "__main__":
    cli()
```

</p>
</details>

### DIFF
--- a/src/NotificationStack.vala
+++ b/src/NotificationStack.vala
@@ -82,8 +82,9 @@ public class Gala.NotificationStack : Object {
             flip_transition.set_key_frames ({ 0.6 });
             flip_transition.set_values ({ -10.0 });
 
-            var entry = new Clutter.TransitionGroup ();
-            entry.duration = 400;
+            var entry = new Clutter.TransitionGroup () {
+                duration = 400
+            };
             entry.add_transition (opacity_transition);
             entry.add_transition (flip_transition);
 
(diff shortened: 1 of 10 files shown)


meisenzahl:
I wonder if the logic could be ported to https://github.com/vala-lang/vala-lint and made available with an option 🤔

fabiozaramella (approved):
Neat 🚀
I agree the script will be very useful for other repos too, thanks for that.

PR #968, opened 2020-11-16
pip: Remove deprecated Clutter.DragAction

davidmhewitt:
`Clutter.DragAction` is gone in mutter-7 (the version above what we're using in 6.0). Let's future proof ourselves and stop using it now.

Only affects the resize button/handle of the PiP plugin, so testing will need to focus around that.


tintou (approved):
Let's land this and deal with regressions later

Issue #789, opened 2020-05-05
1:1 response when swipe to switch workspaces


danrabbit:
~A three-finger horizontal swipe should switch workspaces. Ideally we'd track the swipe 1:1 as in Hdy.Paginator~

Now that we're shipping touchegg, this feature exists, but it should still be a 1:1 response.

andirsun:
Could be possible a three-finger down to hide all wokspaces and show the homescreen ? and again three finger up to re-open ?

danrabbit:
@andirsun that is filed as https://github.com/elementary/gala/issues/790

andirsun:
> @andirsun that is filed as https://github.com/elementary/gala/issues/790

Ohh sorry my mistake thanks a lot .

danrabbit:
This could possibly be resolved by https://github.com/elementary/gala/pull/958 or a follow-up to that branch

JoseExposito:
I'm on it https://github.com/elementary/gala/pull/972

danrabbit:
Fixed in #983!

PR #972, opened 2020-11-23
Gesture and 1:1 animation to switch between workspaces

JoseExposito:
Fix https://github.com/elementary/gala/issues/789

https://github.com/elementary/gala/pull/958 needs to be merged before this one.

### What is included?

- [X] 1:1 animation to switch workspaces
- [X] Cancellation if the gestures doesn't reach the 20% threshold
- [X] 1:1 "nudge" animation as well


JoseExposito:
I'm investigating an issue with the desktop wallpaper window being displayed above the windows in the workspace after connecting a secondary display, changing the main display or reordering the displays.

I'm not able to reproduce it, but it happened to me twice, so I think it is important enough to spend some time debugging it.
I'm not sure if I introduced the bug or it is already there.

Have you ever got a similar error?

My guess is that the issue is in `end_switch_workspace`, as it removes and adds the wallpaper. Maybe I should rethink how the cancel animation works to avoid this in case I introduced it.

hanaral:
@JoseExposito I'm able to reproduce it without an external monitor, but I can't figure out what causes it. My laptop seems to get the gray background after a few workspace switches. It also seems that the workspace switching gesture is reversed for me. I'll report back when I do it again

JoseExposito:
I'm not able to find the steps to reproduce it, but I'm glad you're also able to see the bug.
I'll keep investigating....

hanaral:
@JoseExposito are you able to replecate the reversed direction bug that I have? It seems to only be me

JoseExposito:
That's a work in progress feature. In addition to allow to configure gestures, I'll allow to follow your natural scroll preferences.
But first I need to figure out a way to reproduce the bug I mentioned, hopefully this weekend I'll be able to spend some time looking into it

JoseExposito:
OK, I finally managed to reproduce it :tada:

Just swipe to a new workspace, end the gesture and while it is transitioning start a new swipe gesture.
You will see lots of Glib-CRITICAL in the console and I'll feel bad because I didn't notice this animation doesn't have a flag like the multitasking view one :laughing: 
It is easier to reproduce slowing down the animation a bit. 

Once again I went through the usual development cycle to find it:
Swearing -> Crying -> Thinking it was impossible to reproduce -> Finding it -> Feeling stupid -> Going for a coffee

JoseExposito:
Ok, the issue with the animation is fixed now.

hanaral:
Even with the settings branch of the workspace gestures I get the invisible windows bug very often and have to restart gala to get rid of it. I've found I can reproduce it by swiping around in quick succession, and I think that something this critical should not be left in.

JoseExposito:
Hi @hanaral, could you test this PR please? https://github.com/elementary/gala/pull/983
Just to make sure every change is included. I'm not able to reproduce the issue any more, even if I make the animation really slow. I use this version of Gala everyday, but we might be doing something different.

hanaral:
> Hi @hanaral, could you test this PR please? #983

@JoseExposito I'll test it today and see if I come across anything.

JoseExposito:
By the way @hanaral , what's your Touchégg configuration? It may intercede with Gala. 

To test it as it will be packaged, your config in `/usr/share/touchegg/touchegg.conf` should be like:
https://github.com/elementary/default-settings/pull/198/files#diff-1fa8270101efcba9624c8645e951cdeee3d87190d12de471638f6dc846a23780

And you should not have any config in `/home/<YOUR USER>/.config/touchegg`, you can even remove the folder.

hanaral:
So far so good with the refactor branch, I have not had any issues whatsoever.

> And you should not have any config in `/home/<YOUR USER>/.config/touchegg`, you can even remove the folder.

I nuked my config when I first started texting since I assumed it would.  I'll go ahead and wipe the folder completely if its going to have a gsetting instead.

JoseExposito:
Awesome! Remember to override your config in `/usr/share/touchegg/touchegg.conf` after deleting the folder.
By default Touchégg uses the config in `/usr/share/touchegg/touchegg.conf` if no config is present in your home directory.

Thanks a lot for testing it!

hanaral:
> Awesome! Remember to override your config in `/usr/share/touchegg/touchegg.conf` after deleting the folder.
I'm not quite sure I understand, it seems to work normally just by deleting by user's config.
Also I'm not sure if this is worth reporting but I got the gray background bug for a couple of hundred milliseconds - it didn't do anything else but I'm going to continue testing until I get it again

JoseExposito:
> I'm not quite sure I understand, it seems to work normally just by deleting by user's config.

Ideally your config in /usr/share should be similar to the config that is going to be packaged by default:

```xml
<touchégg>
  <settings>
    <property name="animation_delay">150</property>
    <property name="action_execute_threshold">20</property>
    <property name="color">auto</property>
    <property name="borderColor">auto</property>
  </settings>
  <application name="All">
  </application>
</touchégg>
```

Otherwise, some unwanted gestures might be configured.

> Also I'm not sure if this is worth reporting but I got the gray background bug for a couple of hundred milliseconds - it didn't do anything else but I'm going to continue testing until I get it again

I think that is an already existing bug

hanaral:
> I think that is already an existing bug 

I can consistently reproduce the background bug now. 

1. Lightly swipe back and forth on the same workspace without fully completing any animations

2. Continue swiping until the background plays an animation of it changing to a gray one (but the workspace stays the same)

JoseExposito:
I'm not able to reproduce it :S I know there are some existing problems with the grey background when multiple monitor are connected. Do you have multiple monitors?

hanaral:
> I'm not able to reproduce it :S I know there are some existing problems with the grey background when multiple monitor are connected. Do you have multiple monitors?

Nope, just a laptop (thinkpad x240 for referance)

JoseExposito:
Closing this pull request, changes are included in https://github.com/elementary/gala/pull/983

PR #976, opened 2020-12-01
Gestures configuration

JoseExposito:
Allow to configure the multiasking view and switch workspace gestures.

The UI work is happening here https://github.com/elementary/switchboard-plug-mouse-touchpad/pull/157

https://github.com/elementary/gala/pull/958 and https://github.com/elementary/gala/pull/972 should be merged before this PR.

As suggested by @danrabbit I'm using `/io/elementary/desktop/wm/gestures/` instead of one of the existing schema paths.

PR #983, opened 2020-12-07
Multi-touch support

JoseExposito:
As suggested on slack, moves the `GestureAnimationDirector` to the constructor.

Includes changes from https://github.com/elementary/gala/pull/958, https://github.com/elementary/gala/pull/972 and https://github.com/elementary/gala/pull/976 as doing it in 3 PR was too much work.


plugins/touchegg/Main.vala
```diff
@@ -0,0 +1,157 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+public class Gala.Plugins.Touchegg.Plugin : Gala.Plugin {
+    private Gala.WindowManager? wm = null;
+    private Client? client = null;
+    private GLib.Settings gala_settings = new GLib.Settings ("io.elementary.desktop.wm.gestures");
+    private GLib.Settings touchpad_settings = new GLib.Settings ("org.gnome.desktop.peripherals.touchpad");
```
donadigo, line 23: Is it possible to initialize these in `initialize`? I think this would be the more correct option here.

src/GestureAnimationDirector.vala
```diff
@@ -0,0 +1,83 @@
...
+
+    public signal void on_animation_begin (int percentage);
+    public signal void on_animation_update (int percentage);
+    public signal void on_animation_end (int percentage, bool cancel_action);
+
+    public delegate void OnBegin (int percentage);
+    public delegate void OnUpdate (int percentage);
+    public delegate void OnEnd (int percentage, bool cancel_action);
+
+    private Array<ulong> handlers;
```
donadigo, line 31: We tend to use Gee for lists / arrays and such. You could replace this type with e.g `Gee.ArrayList` and use all native features like `array[i]`, `array.clear ()` and `array.add` and `foreach (var id in array)`.

src/WindowManager.vala
```diff
@@ -95,7 +99,12 @@ namespace Gala {
         private GLib.Settings animations_settings;
         private GLib.Settings behavior_settings;
 
+        private bool animating_switch_workspace = false;
+        private GestureAnimationDirector gesture_animation_director;
```
donadigo, line 103: Could you clear as to why is there a reason to have two `GestureAnimationDirector`'s declared in WindowManager and in MultitaskingView? Is it because we want to track handlers for the two classes separately?

donadigo (changes requested):
I'm now much more confident about this code than before. Clearly a lot of improvement and it seems that passing the Director in constructors also is the way to go. I still have 3 comments.

src/WindowManager.vala
JoseExposito: No reason at all, the same director can be used and passed in the constructor

plugins/touchegg/Main.vala
```diff
@@ -0,0 +1,159 @@
...
+        bool enabled = gala_settings.get_boolean ("multitasking-gesture-enabled");
+        int fingers = gala_settings.get_int ("multitasking-gesture-fingers");
+
+        return enabled
+            && gesture.type == GestureType.SWIPE
+            && gesture.direction == GestureDirection.DOWN
+            && gesture.fingers == fingers;
+    }
+
+    private bool is_next_desktop_gesture (Gesture gesture) {
```
donadigo, line 121: I feel like `is_next_desktop_gesture` and `is_previous_desktop_gesture` could be combined into one method that returns a bool and the direction by an `out` parameter like this:
```
private bool is_switch_desktop_gesture (Gesture gesture, out Meta.MotionDirection dir) {
    ...

    if (enabled && gesture.type == GestureType.SWIPE && gesture.fingers == fingers) {
         if (gesture.direction == GestureDirection.LEFT) {
             dir = natural_scroll ? Meta.MotionDirection.RIGHT : Meta.MotionDirection.LEFT;
             return true;
         } else if (gesture.direction == GestureDirection.RIGHT) {
             dir = natural_scroll ? Meta.MotionDirection.LEFT : Meta.MotionDirection.RIGHT;
             return true;
         }
    }

    return false;
}
```

And then use:
```
Meta.MotionDirection dir;
if (is_switch_desktop_gesture (gesture, out dir)) {
    if (!wm.workspace_view.is_opened ()) {
        wm.switch_to_next_workspace (dir, hints);
    }
}
````

**Review by donadigo — CHANGES_REQUESTED** [review]
Some more comments. Mostly about typos but there's one about unifying the method to determine if the swipe is in either direction for desktop switch.

File: plugins/touchegg/Client.vala [hunk 5]
```diff
@@ -0,0 +1,195 @@
...
+
+        /**
+         * Socket used to connect to the daemon.
+         */
+        private Socket? socket = null;
+
+        /**
+         * Current number of reconnection attemps.
+         */
+        private int reconnection_attemps = 0;
```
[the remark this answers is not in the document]
JoseExposito: Good catch. I usually rely on a VSCode plugin for typos, but it wasn't enabled for vala files. Fixed a few of them. Thanks!

plugins/touchegg/Main.vala
JoseExposito: Because the methods are rather simple and this only avoids duplicating a couple of bool checks, I'll rather leave as it is. In my opinion is more readable and avoids returning 2 arguments from a function. But feel free to modify it if you really don't like it.

donadigo: Absolutely, this was just a suggestion and nothing that I feel strongly about. The separate versions are easier to reason about for sure.

danrabbit (commented):
Something I'm noticing is there appears to be no upper limit on animation speed, which kind of takes away from the smoothness if you gesture too quickly. Maybe @Exalm knows off the top of his head if Handy limits the animation speed and what that value is, but I would imagine we wouldn't want the animation to be shorter than like 200ms or something

JoseExposito:
I know he had some thoughts about using unaccelerated events from libinput. That alone could reduce a lot this effect. I'll try to test it this weekend and see how it feels.

But it I'll be defined good to know from him, especially now that GTK4  is around the corner.

plugins/touchegg/Main.vala
```diff
@@ -0,0 +1,159 @@
...
+            wm.workspace_view.close (hints);
+        }
+
+        if (is_next_desktop_gesture (gesture)) {
+            if (!wm.workspace_view.is_opened ()) {
+                wm.switch_to_next_workspace (Meta.MotionDirection.RIGHT, hints);
+            }
+        }
+
+        if (is_previous_desktop_gesture (gesture)) {
```
donadigo, line 81: I think we could make all the if statements an `else if` here, removing the need to check for other gestures when we already executed an action based on it.

donadigo (changes requested):
One more thing.

donadigo (approved):
Good job, code review approve. That said, I'd really like @tintou to do a review as well, but if he's not available, we could merge this after a UX review as well.

danrabbit:
I'm happy with the behavior. I think we can iterate a bit on the animation smoothness but what's here is a huge improvement so UX approve from me :)

tintou:
I'm not really a fan of passing structs via a socket, you can also use D-Bus over a socket instead which would make it look better for inter-process communication. That won't block my approval here but I'd love to see a D-Bus iteration of this :wink:

danrabbit:
Since it's been such a long work in progress, I'm inclined to merge and then I'll file two issues with the above comments so we can continue to iterate on it :)

PR #992, opened 2020-12-17
Updated build to package the binary as a library, so can reuse compon…

89jd:
Changes to have OSX functionality of showing windows that match current application as well as existing that shows all windows open.

I created as a plugin for now, so didn't need to recompile and install gala on my machine. But ultimately, would be best if I subclassed WindowOverview, and then just override the 1 bit I need to cut down on code changes.

Let me know if there is anything I can do.

I have tested, and works as expected for me

Cheers

Jack

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,4 @@ po/Rules-quot
 po/stamp-po
 src/gala
 build
+.vscode/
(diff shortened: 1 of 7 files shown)


89jd:
I am not sure why it is isn't building. Works for me. Again, happy to change if any ideas

jeremypw:
CLosing as not reviewed/actioned after 2+yrs and conflicts

Issue #926, opened 2020-10-06
Background flashes while switching workspaces with multiple displays


danrabbit:
While using multiple displays, the background flashes while switching workspaces using keyboard shortcuts.

My guess is that this has to do with a kind of hacky thing that was done to give the illusion of space between workspaces. Gala is drawing over the workspace instead of them actually being spread apart during the animation. So we can probably remove that completely

hanaral:
Can confirm this, it seems to shift the background image in the animation and then ping back to its 'true' position. It does this with vertical displays too even with single display layouts but I might need to open a seperate report for that.

JoseExposito:
It is completely replacing the wallpaper in the current desktop with a grey square. The animation is too fast to notice it, but if you slow it down it is easier to see.

This was driving me crazy!

PR #994, opened 2020-12-19
Background disappears while switching workspaces with multiple displays

JoseExposito:
Fix https://github.com/elementary/gala/issues/926


cassidyjames:
Ha, I'm actually seeing the opposite effect with this branch; when switching workspaces, the wallpaper disappears and then snaps back when the animation is done. In master, the animation seems pretty misaligned and this branch _does_ fix that, though.

JoseExposito:
Could you try again @cassidyjames ? My monitors have the same size and, turns out, that was an edge case.
I tested changing the resolution and it works, no matter the how you align them. But I appreciate if you test it with other hardware.

cassidyjames:
@JoseExposito yes, the latest commit fixes the issue for me. This looks great, now!

cassidyjames (approved):

PR #996, opened 2020-12-21
Fix #991: Add a gap between workspaces when swiping/navigating

### LINKED ISSUE #991: Add a gap between workspaces when swiping/navigating
## Problem

It wasn't as noticeable before, but now that you can just sit here in in-between states it doesn't look as clean that workspaces are just right up against each other

Screenshot from 2020-12-16 13 21 47@2x

## Proposal

There should be some space between workspaces. Usually we do 24px of space between discrete sections in the UI

JoseExposito:
Fix https://github.com/elementary/gala/issues/991

### DIFF
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -102,6 +102,8 @@ namespace Gala {
         private bool animating_switch_workspace = false;
         private GestureAnimationDirector gesture_animation_director;
 
+        private const int WORKSPACE_GAP = 24;
+
         construct {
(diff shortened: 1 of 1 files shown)


danrabbit (changes requested):
Don't forget to multiply by scaling factor :+1: 

Otherwise, looks good!

JoseExposito:
Here you go, I wish those were CSS pixels :P

danrabbit (approved):
Looks good! Thanks :) And yeah definitely wish it was automagically DPI aware

Issue #998, opened 2020-12-21
Allow apps to overlap workspace edges in multitasking view


danrabbit:
Seeing mockups for GNOME 40, I think there's more of a sense of depth and it looks nice that windows kind of come out over the edges of the workspaces:

![pan-windowpicker](https://user-images.githubusercontent.com/7277719/102807409-6fbeb000-4373-11eb-842a-350637f0ce17.png)

We could zoom those wallpapers back further since they're so unimportant here and make apps feel like they're more focused and popped out

JoseExposito:
It is easy to implement, however I kind of like it more in the mocks -with conveniently sized windows- that in real life.

One single window maximized is like the worst case.
Here are some examples:

![0](https://user-images.githubusercontent.com/1335948/102809659-ad96f500-43c2-11eb-84bf-4c0e9ab753e4.png)
![1](https://user-images.githubusercontent.com/1335948/102809683-b982b700-43c2-11eb-89e7-ad90eb1e554b.png)
![2](https://user-images.githubusercontent.com/1335948/102809688-bb4c7a80-43c2-11eb-828f-be489e5a4307.png)
![3](https://user-images.githubusercontent.com/1335948/102809693-bd163e00-43c2-11eb-9416-29e3b95f7e0d.png)
![4](https://user-images.githubusercontent.com/1335948/102809700-c0112e80-43c2-11eb-9e03-3fddf3a2f4e6.png)

I can create a WIP PR if you want to test with different values.

danrabbit:
Yeah there definitely should be much less of a difference between the edges of the window spread area and the wallpaper area. I played with that patch you sent and got something more like this:

![Screenshot from 2020-12-21 10 56 18@2x](https://user-images.githubusercontent.com/7277719/102823675-35afd700-4390-11eb-952a-f5ce0d21bf2e.png)

Obviously I missed it up and the switcher is below the screen, but you get the idea. Something else that I think is making the actual implementation look not very good is that windows are so close together in the spread. So in the GNOME 40 mockups you'll notice that the window previews are much smaller than what we have here

JoseExposito:
I just created a WIP PR so we can adjust some constants there https://github.com/elementary/gala/pull/999

Issue #997, opened 2020-12-21
Workspaces in multitasking view animate as if they are stacked


danrabbit:
## What Happened

Now that you can see the multitasking view at your own pace with gestures, it's obvious that the origin point for adjacent desktops is behind the current desktop. This is inconsistent with the spacial model you would assume from horizontal swipes.

![Screenshot from 2020-12-21 09 50 17@2x](https://user-images.githubusercontent.com/7277719/102806567-f96d7e00-4371-11eb-855e-531f33988ad4.png)

## Expected Behavior

Desktops should appear to always be laid out horizontally and its only our view of them that zooms out to reveal that fact.

## Steps to Reproduce

Set up a few workspaces and then slowly push your way into multitasking view with gestures. Notice adjacent workspaces come out from behind the focused workspace instead of already being adjacent and simply zooming out

PR #999, opened 2020-12-22
#998: Allow apps to overlap workspace edges in multitasking view

JoseExposito:
Here is the required code to implement https://github.com/elementary/gala/issues/998

It is marked as WIP because @danrabbit might want to adjust some values.

padding

### DIFF
--- a/src/InternalUtils.vala
+++ b/src/InternalUtils.vala
@@ -348,7 +348,7 @@ namespace Gala {
                                          area.y + (slot / columns) * slot_height,
                                          slot_width,
                                          slot_height};
-                target = rect_adjusted (target, 10, 10, -10, -10);
+                target = rect_adjusted (target, 20, 20, -20, -20);
 
                 float scale;
                 if (target.width / (double)rect.width < target.height / (double)rect.height) {
--- a/src/Widgets/WorkspaceClone.vala
+++ b/src/Widgets/WorkspaceClone.vala
@@ -123,6 +123,11 @@ namespace Gala {
         const int TOP_OFFSET = 20;
 
         /**
+         * Background wallpaper extra margin to apply.
+         */
+        const int BACKGROUND_OFFSET = 80;
+
+        /**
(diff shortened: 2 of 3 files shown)


cassidyjames:
@JoseExposito it might be related to/require fixing #997, but something I'm noticing here: the adjacent workspaces are now animated to way off screen which makes them not visible in the view (unless there's a maximized window). I think we want to keep them fairly visible, maybe somewhere in between master and that GNOME mockup.

Something else @elementary/ux needs to determine is if/how the bottom UI showing workspace previews should be adjusted, keeping in mind windows _can_ but won't _always_ go off the bottom of the workspace. But as-is it looks a little unbalanced to me when you don't have a window overlapping the bottom, and especially on larger displays.

Screenshot from 2020-12-22 11 12 36

JoseExposito:
You are right. In order to make it visible, I can think of 2 options:

 - Make the space used by the app grid a little bit smaller
 - Or just make the workspaces closer

Have in mind that, when only one maximized window is open (IMHO that looks terrible) the wallpaper will not be displayed. This might make a little bit less intuitive that windows can be drag&dropped to the next/previous workspace.

hanaral:
I'd like to chime in a little; I really am not seeing the benefits here, it seems to only bring in a few UX issues that need to be solved. I think that the current setup is fine, and has nice padding and even distances between the elements. 
_(as a seperate issue altogether there should be a smaller maximum window preview size, apps like dconf editor look like overinflated baloons)_

JoseExposito:
Fixing #997 was required to adjust the constants in a sane way, so I solved that in a different PR and merged it here as well.

However, I agree with @hanaral , this is creating some UX issues that we need to fix, or just discard this idea. This is the current state of this branch:

Screenshot from 2020-12-27 12 43 25

When dark style is selected, there is not enough contrast between the windows and the gray area behind the wallpaper.

danirabbit:
Closing since there's been no movement here for several years

PR #1002, opened 2020-12-27
Workspaces in multitasking view should not animate as if they are stacked

JoseExposito:
A fix for this animation issue https://github.com/elementary/gala/issues/997

### DIFF
--- a/src/Widgets/MultitaskingView.vala
+++ b/src/Widgets/MultitaskingView.vala
@@ -300,7 +300,6 @@ namespace Gala {
          *                positions immediately.
          */
         void update_positions (bool animate) {
-            var scale = InternalUtils.get_ui_scaling_factor ();
 #if HAS_MUTTER330
             unowned Meta.WorkspaceManager manager = display.get_workspace_manager ();
             var active_index = manager.get_active_workspace ().index ();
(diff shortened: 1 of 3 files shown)


danrabbit (changes requested):
This is so much better! Don't forget to also include that area we added between workspaces when swiping. I guess this should be set from the same constant so that it stays consistent

JoseExposito:
Those are 2 different constants. It is a little bit confusing, but, in the multitasking view, the workspace (clones) are presented in an actor and they are actually overlapping. In this way, when the workspaces are scaled down, the adjentend workspaces are visible.

However, when switching workspaces, WorkspaceClone is not used. A wallpaper clone is created and there is actually a gap:
https://github.com/elementary/gala/blob/master/src/WindowManager.vala#L105

So, in my opinion, it is better not to move the workspace gap constant to the WorkspaceClone class, as it is not involved in the switch workspace animation.

danrabbit:
Right, so you and I know that these are two different implementations, but to the user there should be a consistent sense of object permanence. You can see that it's a bit weird that when you swipe workspaces there's this gap between the workspaces present, but if you push into multitasking view you can see workspaces are right up against each other.

JoseExposito:
Aaaaahh ok, I didn't understand correctly your first comment. I thought you wanted to move the constant to the WorkspaceClone, that's why I pointed out that they're different stuff.
Now I understood what you meant 🤣

danrabbit (approved):
Looks great!

Issue #989, opened 2020-12-16
Multitouch gestures have no upper speed limit


danrabbit:
## Problem

When performing multi-touch swipes, if you swipe fast enough there is no perceivable animation or transition. This ends up feeling kind of janky and not smooth

## Proposal

There should be some upper speed limit on transitions so that we're always at least getting some kind of animation. Maybe like a 200-300ms minimum animation time or something

## Prior Art

We should look at what Handy does here so that it feels internally consistent between app gestures and WM gestures

danrabbit:
@Exalm if you could point to code in Handy or otherwise comment/explain what is done there that would be great :)

Exalm:
libhandy clamps to [100:400] ms range.

Exalm:
What you really want here is a spring animation and not to use velocity, but that's patented. 😒️

JoseExposito:
@Exalm could you point me to the implementation of this in libhandy, please? I'll need to read a lot of libhandy's code to make this feel consistent.
Ideally, I'll like to also make the final transition kinetic and that should also be consistent with libhandy.

Just thinking out loud, as this could take a ridiculous amount of work or just not being possible, but since you are here, I can not loose the opportunity to ask you :smile:
I'm not familiar with libhandy's code, but... Do you think that it'll be possible to isolate the code in charge of calculating stuff like minimal animation duration, kinetic physics, etc and move it to a shared place?
I'll be happy do help with that upstream, and I think it could be very useful to many many projects out there, not just Gala.

Exalm:
https://gitlab.gnome.org/GNOME/libhandy/blob/master/src/hdy-swipe-tracker.c#L325 it's literlaly just clamping. :p

Exalm:
> Do you think that it'll be possible to isolate the code in charge of calculating stuff like minimal animation duration, kinetic physics, etc and move it to a shared place?

There aren't much physics there atm really. There is a big rework in https://gitlab.gnome.org/GNOME/libhandy/-/merge_requests/543 though, that one is actually physics-based.

I think sharing would be overkill here tbh, there's not _that_ much code, and e.g. WebKit won't want to depend on that so you'd still have copypasta.

JoseExposito:
> https://gitlab.gnome.org/GNOME/libhandy/blob/master/src/hdy-swipe-tracker.c#L325 it's literlaly just clamping. :p

Thanks for the link!!
[...]

PR #1006, opened 2020-12-29
#989: Multi-touch gestures have no upper speed limit

JoseExposito:
Here is some initial testing with the switch workspace gesture to mimic libhandy/gnome-shell behaviour.

As suggested in https://github.com/elementary/gala/issues/989 by @Exalm I had a look to what libhandy and gnome-shell are doing under the hood and port it (or at least tried to port it) to Gala.

Main changes:

 - Update gesture phase: Compile and install the latest Touchégg so unaccelerated deltas are used. 
 - End gesture phase: The code changes included in this PR.
   Because, for the moment, this PR is just for testing, I didn't move the `duration` calculation to the `GestureAnimationDirector`, but it'll go there so it can be used in other gestures.

Let me know what do you think.


hanaral:
I think that it feels a little too rapid to feel natural. Is it possible to have a ramp up and down curve in the speed of the animation (I think this is how macOS does it)?

lib/Constants.vala
```diff
@@ -29,6 +29,7 @@ namespace Gala {
         // Duration of the snap animation as used by maximize/unmaximize
         SNAP = 250,
         // Duration of the workspace switch animation
-        WORKSPACE_SWITCH = 300,
+        WORKSPACE_SWITCH_MIN = 100,
```
danrabbit, line 32: This feels like 10000% better than master already, but I would be okay with going even longer since this is such a big animation. Even at `300` for minimum, it still feels pretty fast imo

JoseExposito:
I made some extra changes:

 - Update phase: I limited the maximum speed. When reached, `GestureAnimationDirector.MAX_VELOCITY` is used instead of the gesture velocity. Play a little bit with that constant if you want.
 - End phase: Minimum animation time on 300 instead of 100.
 - End phase: Previously, the action was cancelled if the animation didn't reached the 20%. I also added a speed based condition so small but fast swipes trigger the action.

> I think that it feels a little too rapid to feel natural. Is it possible to have a ramp up and down curve in the speed of the animation (I think this is how macOS does it)?

Definitely the animation performed when the gesture ends can be improved, specially with slow swipes. I believe macOS calculates proper acceleration/deceleration.
In my opinion,  this is the last thing that'd be nice to implement in order to have high quality animations. I'll appreciate some help implementing it.

-------------------

Unless you want to make some code changes or adjust constants, I'm quite happy with this PR.
Remember to test the multitasking view as well, as the change applies there as well.

As a side note, the multitasking view animation mode is `EASE_OUT_QUAD` instead of `EASE_OUT_CUBIC`. Maybe we should use the same mode in both animations.

danrabbit:
Entering multitasking view is now much much more consistent. Nice work smoothing things out! This feels soooo good :heart:

danrabbit:
I think the difference in animation timing is intentional. In one animation you have an element exiting the screen while another element enters the screen, but in the other animation you're only really pushing away the single element to a new position on screen

JoseExposito:
Cool! Feel free to merge the PR when you are done testing it. I don't need to make any more code changes.
I'll release in a few days a new version of Touchégg as well to make this feature fully available in the daily builds.

danrabbit:
I think it would be good to get a quick code review from @donadigo, but I am happy with the UX here :)

donadigo (approved):
Code looks good.

PR #1011, opened 2021-01-03
 Make Mutter 3.36 the minimum libmutter version

tintou:
We don't expect gala to work with something lower than 3.36 (elementary Odin)


danrabbit (approved):
Let's gooo

Issue #990, opened 2020-12-16
Communicating with Touchégg should not pass structs via a socket


danrabbit:
@tintou's comments here:
> I'm not really a fan of passing structs via a socket, you can also use D-Bus over a socket instead which would make it look better for inter-process communication.
> for an example on how to do it, it's what I've done for switchboard:
> https://github.com/elementary/applications-menu/blob/master/src/synapse-plugins/worker-link.vala
> https://github.com/elementary/applications-menu/blob/master/src/synapse-plugins/switchboard-plugin.vala
> https://github.com/elementary/applications-menu/blob/master/src/synapse-plugins/switchboard-plugin/plugin.vala

JoseExposito:
This could be an interesting addition, specially to make easier from higher level languages that don't have access to the big variety of ints C++ has or that might align memory in a different way, like JS.

I'll look into this, but with low priority.

JoseExposito:
I just finished the implementation upstream: https://github.com/JoseExposito/touchegg/issues/436

As soon as I figure out how D-Bus works in Vala, I'll do my best to port it.

meisenzahl:
@JoseExposito looking at [your implementation](https://github.com/JoseExposito/touchegg/blob/42bb9748e2c140095c0fd572f3c03f6a3fcc81c6/src/daemon/dbus.h) using only signals it should be pretty forward:

1. [Create an interface for the D-Bus service](https://github.com/elementary/switchboard-plug-about/blob/f8392a34ae3eff9a54219f0b718055cb6134004a/src/Interfaces/FwupdManager.vala#L23)
2. [Get a proxy for the service](https://github.com/elementary/switchboard-plug-about/blob/f8392a34ae3eff9a54219f0b718055cb6134004a/src/Interfaces/FwupdManager.vala#L285)
3. [Connect to the signal on the proxy](https://github.com/elementary/switchboard-plug-about/blob/f8392a34ae3eff9a54219f0b718055cb6134004a/src/Interfaces/FwupdManager.vala#L287)

If I can help you in any way, let me know :)

PR #1017, opened 2021-01-14
Multi-touch gestures aren't pixel-precise at low speeds + D-Bus

### LINKED ISSUE #1016: Multi-touch gestures aren't pixel-precise at low speeds
## What Happened
This is less of a bug and more of unwanted behavior. Touchegg gestures track fingers really well but the animations are choppy/jumpy. When you move your fingers slowly, you can see each animation "frame" jumps about 10px in size. Maybe there is a performance reason for this, but I hoping we could reduce the amount of pixels each "frame" moves.

## Expected Behavior

I took a look at my wife's macbook. The animations are a bit smoother with each "jump" being only a pixel or two.

### Logs

No logs to show but can be shown if requested. I imagine this is less of a "my machine" thing. 

## Platform Information

Manjaro Gnome with a pantheon session.

### Demo
The gif below will show what I am talking about. You can see when i move my fingers consistently, the animation looks pretty smooth. However, if I slow down, you can begin to see how large each jump is. 

Screen Capture_select-area_20210113103841-min

JoseExposito:
Fix #1016 and https://github.com/elementary/gala/issues/990

Requires the changes introduced in the daemon here: https://github.com/JoseExposito/touchegg/issues/434 & https://github.com/JoseExposito/touchegg/issues/436

Merging this will need some synchronization releasing the daemon, just ping me on Slack before merging this.


danrabbit (dismissed):
I can confirm this works as expected with the latest Touchégg master. So we just are blocked until the next upstream release :)

JoseExposito:
As I mentioned in Slack, I was working on D-Bus support. Since this branch is already blocked and the migration to D-Bus also uses a `double` for gesture percentage, it didn't make sense to create a different PR.

I followed @tintou's [implementation](https://github.com/elementary/applications-menu/blob/master/src/synapse-plugins/switchboard-plugin/plugin.vala), in case you want to follow that for the review. He also helped me with a bug I wasn't able to find, so thanks a lot! :smile: 

I tried to follow @meisenzahl's [advice](https://github.com/elementary/gala/issues/990#issuecomment-766308634) but I wasn't able to use an interface as he suggested over Unix socket. I don't know if it is possible, but I wasn't able to find a way.
I have 0 experience with D-Bus, so I appreciate every comment you can make to improve the code.

JoseExposito:
I'll release a new Touchégg version in 1 or 2 days. I think it is safe to do so as the package should be generated from os-patches and not from upstream, but I just wanted to make sure this is the case to avoid breaking Gala gestures :smile:

danrabbit:
@JoseExposito it won't break anything as I'm manually pulling in release snapshots :)

danrabbit (approved):
Extremely smooth. Nice work!

andrewc910:
Working great on my XPS 9560. Thank you @JoseExposito

PR #1020, opened 2021-01-16
ScreenShield: Use async versions of Bus.get_proxy

davidmhewitt:
This class' constructor is on the critical path of Gala starting up, so we should use the async versions of these methods. There are already checks in place that the interfaces are ready before using them in all the other methods.


tintou (approved):
Looks like a reasonable thing to do

PR #1021, opened 2021-01-16
AppCache: Rebuild in a thread

davidmhewitt:
This class is initialized in a static constructor so its in the critical path of Gala starting up. There isn't an async version of `GLib.AppInfo.get_all` and it's an I/O heavy operation that enumerates the desktop files on the system, so we should move this out of the main compositor thread.

### DIFF
--- a/lib/AppCache.vala
+++ b/lib/AppCache.vala
@@ -34,7 +34,7 @@ public class Gala.AppCache : GLib.Object {
         app_info_monitor = GLib.AppInfoMonitor.@get ();
         app_info_monitor.changed.connect (queue_cache_update);
 
-        rebuild_cache ();
+        rebuild_cache.begin ();
     }
 
     private void queue_cache_update () {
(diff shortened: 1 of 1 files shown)


lib/AppCache.vala
```diff
@@ -43,36 +43,48 @@ public class Gala.AppCache : GLib.Object {
...
+                id_to_app.clear ();
 
-        foreach (unowned GLib.AppInfo app in app_infos) {
-            var id = app.get_id ();
-            var startup_wm_class = ((GLib.DesktopAppInfo)app).get_startup_wm_class ();
+                var app_infos = GLib.AppInfo.get_all ();
 
-            id_to_app[id] = (GLib.DesktopAppInfo)app;
+                foreach (unowned GLib.AppInfo app in app_infos) {
+                    var id = app.get_id ();
```
tintou, line 68:
```suggestion
                    unowned string id = app.get_id ();
```

lib/AppCache.vala
```diff
@@ -43,36 +43,48 @@ public class Gala.AppCache : GLib.Object {
...
+                    var startup_wm_class = ((GLib.DesktopAppInfo)app).get_startup_wm_class ();
```
tintou, line 69:
```suggestion
                    unowned string startup_wm_class = ((GLib.DesktopAppInfo)app).get_startup_wm_class ();
```

tintou:
Two tiny bits and it should be great to finally take advantage of this GLib capability

tintou (approved):

PR #1023, opened 2021-01-20
Fix #1018: Align multi-touch animations to physical pixels

### LINKED ISSUE #1018: Align multi-touch animations to physical pixels
## Problem

As mentioned in a [different issue](https://github.com/elementary/gala/issues/1016#issuecomment-759790517) by @Exalm it'd be interesting to align multi-touch animations to physical pixels.

In this way, in progress multi-touch animations will never look blurry, even in HiDPI displays.

## Proposal

Investigate if Clutter can handle this in low and high DPI displays and, if it can't, implement it in Gala.

## Prior Art

Is libhandy already doing this?

JoseExposito:
Allow to align the switch workspace animation to physical pixels.

With the current master it is not easy to notice the difference, it is easier to test with https://github.com/elementary/gala/pull/1017

With LoDPI displays, the values returned by `GestureAnimationDirector.animation_value` are integers (0, 1, 2, 3...)
With HiDPI displays, let's say that the scale factor is 2, values will be 0, 0.5, 1, 1.5...

If I understood how scaling works in Gala those should be the right values on HiDPI. However, I don't have a HiDPI display to test this, so I'd appreciate your help testing it for me :smile: 

Fix https://github.com/elementary/gala/issues/1018

### DIFF
--- a/src/GestureAnimationDirector.vala
+++ b/src/GestureAnimationDirector.vala
@@ -96,8 +96,15 @@ public class Gala.GestureAnimationDirector : Object {
         handlers.clear ();
     }
 
-    public static float animation_value (float initial_value, float target_value, int percentage) {
-        return (((target_value - initial_value) * percentage) / 100) + initial_value;
+    public static float animation_value (float initial_value, float target_value, int percentage, bool rounded = false) {
+        float value = (((target_value - initial_value) * percentage) / 100) + initial_value;
+
+        if (rounded) {
+            var scale_factor = InternalUtils.get_ui_scaling_factor ();
+            value = (float) Math.round (value * scale_factor) / scale_factor;
+        }
+
+        return value;
     }
 
     public void update_animation (HashTable<string,Variant> hints) {
--- a/src/GestureAnimationDirector.vala
+++ b/src/GestureAnimationDirector.vala
@@ -96,6 +96,18 @@ public class Gala.GestureAnimationDirector : Object {
         handlers.clear ();
     }
 
+    /**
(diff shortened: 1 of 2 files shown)


src/GestureAnimationDirector.vala
```diff
@@ -96,8 +96,15 @@ public class Gala.GestureAnimationDirector : Object {
         handlers.clear ();
     }
 
-    public static float animation_value (float initial_value, float target_value, int percentage) {
-        return (((target_value - initial_value) * percentage) / 100) + initial_value;
+    public static float animation_value (float initial_value, float target_value, int percentage, bool rounded = false) {
```
Philip-Scott, line 99: Are there animations you don't want to round off?

JoseExposito: Animations that involve scaling, like the multitasking view desktop animation, use a value between 0.0 and 1.0. Rounding that value breaks the animation.
Also, when scaling a window down, the window may not have a size divisible by physical pixels. That's why the `rounded` flag is not set in those cases.

Philip-Scott: Awesome, thank you for explaining. Would you mind though just writing that as a comment on the function? :) That way a year from now when we or someone else looks at the code, we will know the reason behind this

JoseExposito: I just documented the method. Thanks for the review Felipe.

JoseExposito:
Done :+1:

danrabbit (approved):

Issue #886, opened 2020-07-28
Auto-move maximized windows with multi-monitor setups


andreasfelix:
See #885 
Related to #788, #643

Currently their are some issues with the *auto-move maximized* windows on multi-monitor setups:

* The workspace on the primary monitor changes, if one maximizes a window on a secondary monitor.
* The workspace on the primary monitor changes, if one un-maximizes a window on a secondary monitor, even if the workspace is non-empty.

One solution would be to check on if the maximized window is one the primary and only create a new workspace if this is the case. But for me, this wouldn't be an optimal solution as this would mean, that the maximize action would behave differently depending on if a window is on a primary on secondary monitor. A real solution (with consistent behavior) would require an implementation of workspaces for secondary monitors. This would also raise the question if multiple monitors should share workspaces or if workspaces should be specific to the individual monitor. I would be in favor for the second option.

cassidyjames:
>A real solution (with consistent behavior) would require an implementation of workspaces for secondary monitors.

For sure, see #26. Probably my biggest wishlist item for Gala, but afaik it's still not trivial due to it not existing in Mutter or X or something. But I do wonder if we could paper over/fake it.

lenemter:
Fixed in https://github.com/elementary/gala/pull/1033

PR #885, opened 2020-07-28
Auto-move maximized windows by default

danrabbit:

PR #1033, opened 2021-01-29
Don't move to next workspace when maximizing a window on secondary monitor

Philip-Scott:
If you are on a multi-monitor setup, maximizing a window on a secondary monitor causes the workspaces on the primary monitor to  move to the next one, even though the maximized window is not moving. I'm just adding a check on `move_window_to_next_ws ()` to ensure that the window is on the primary monitor before doing any other checks

### DIFF
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -1620,7 +1620,7 @@ namespace Gala {
             }
 
             // Do nothing if window is not on primary monitor
-            if (!window.is_on_primary_monitor()) {
+            if (!window.is_on_primary_monitor ()) {
                 return;
             }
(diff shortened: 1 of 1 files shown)


andreasfelix (approved):
LGTM and fixes the regressions #885
also partially fixes #886

src/WindowManager.vala
```diff
@@ -1619,6 +1619,11 @@ namespace Gala {
                 return;
             }
 
+            // Do nothing if window is not on primary monitor
```
andreasfelix, line 1622: For me, this comment seems a bit too obvious ;)

PR #1036, opened 2021-01-31
Fix #1032: Workspace doesn't bump hard enough when swiping at the end

### LINKED ISSUE #1032: Workspace doesn't bump hard enough when swiping at the end
## What Happened

If you swipe at the end of the workspace list, there is a bump animation to let you know you've reached the end and can't go any further. However, this bump is often much smaller than the bump you see when using keyboard shortcuts. Sometimes it's only a few pixels wide

## Expected Behavior

If we're bumping, we should make sure the animation is dramatic enough to be noticeable

JoseExposito:
Fix https://github.com/elementary/gala/issues/1032

### DIFF
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -521,13 +521,24 @@ namespace Gala {
             }
 
             animating_switch_workspace = true;
+            var nudge_gap = NUDGE_GAP * InternalUtils.get_ui_scaling_factor ();
 
-            var dest = (direction == Meta.MotionDirection.LEFT ? NUDGE_GAP : -NUDGE_GAP);
-            dest *= InternalUtils.get_ui_scaling_factor ();
+            float dest = 0;
+            if (!switch_workspace_with_gesture) {
+                dest = nudge_gap;
+            } else {
+                unowned Meta.Display display = get_display ();
+                var workspaces_geometry = InternalUtils.get_workspaces_geometry (display);
+                dest = workspaces_geometry.width;
+            }
+
+            if (direction == Meta.MotionDirection.RIGHT) {
+                dest *= -1;
+            }
 
             GestureTracker.OnUpdate on_animation_update = (percentage) => {
                 var x = GestureTracker.animation_value (0.0f, dest, percentage, true);
-                ui_group.x = x;
+                ui_group.x = x.clamp (-nudge_gap, nudge_gap);
             };
 
             GestureTracker.OnEnd on_animation_end = (percentage, cancel_action) => {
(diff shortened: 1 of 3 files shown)


src/WindowManager.vala
```diff
@@ -105,6 +105,13 @@ namespace Gala {
          */
         public const int WORKSPACE_GAP = 24;
 
+        /**
+         * When swiping at the begin/end of the workspace list, a nudge animation is played.
+         * If using a touchpad or touchscreen the animation is smaller that using the keyboard. This
```
jlnr, line 110: Typo?
```suggestion
         * If using a touchpad or touchscreen the animation is smaller than using the keyboard. This
```

JoseExposito: Yep

danrabbit (changes requested):
Hrm this seems really arbitrary and I don't think really solves the issue. This would make the nudge much large sometimes and still smaller sometimes. It wouldn't make the two nudges uniform.

When you start a workspace switch, the minimum distance the workspace is moved when the gesture is accepted is about the distance of the workspace gap, but when you start a nudge, the minimum distance is just a few pixels. So if you're swiping quickly, even with this branch you might only move 6px or so and the nudge animation is barely noticeable.

JoseExposito:
> When you start a workspace switch, the minimum distance the workspace is moved when the gesture is accepted is about the distance of the workspace gap, but when you start a nudge, the minimum distance is just a few pixels. So if you're swiping quickly, even with this branch you might only move 6px or so and the nudge animation is barely noticeable.

Aham, now I understand what solution you want.

So, when you switch workspaces, the 100% of the movement in your touchpad matched the size of your screen.
However, in the nudge animation that 100% of movement matches 32 pixels.
I just pushed some changes so both animation are identical with the exception than the nudge animation ends at 32px.

Even though this is the "coherent" way of handling this animation it feels a little bit rough...

hanaral:
I've already filed feature request for  - But I was wondering how difficult it would be to make the workspace bump a part of the animation before it i.e. swiping from workspace 2 to workspace 1 and continuing to pull would animate the stretch

JoseExposito:
> I've already filed feature request for it - But I was wondering how difficult it would be to make this workspace bump a part of the animation before it i.e. swiping from workspace 2 to workspace 1 and continuing to pull would animate the stretch

I'm afraid that at the moment that'll require few changes... We'll need to handle that extra swipe movement and also change how the animation is performed drastically as, under the hood, only the current desktop + windows and the target desktop + windows are drawn.
It's possible, but it'll require a lot of work. By the way, feel free to mention me in a issue if you want to ask a question, I don't check the issue tracker that often :smile:

danrabbit:
can confirm, this branch bumps :sunglasses:

JoseExposito:
@danrabbit I rebased the changes in 3 commits so the change is easier to follow

danrabbit (approved):

PR #1039, opened 2021-02-02
Calculate notification width dynamically

### LINKED ISSUE #821: Notifications render too far to the right on the primary monitor
## Prerequisites

## Describe the bug
I'm running Elementary OS 6.0-next on Ubuntu 20.04 as a base (I changed my /etc/apt/sources.list on Elementary OS 5.0 and changed PPAs to the daily PPAs to "upgrade). I noticed when notifications appear, they appear a little too far to the right, going onto the monitor right of my primary monitor.

Edit: Running Gala version 3.3.2

### To Reproduce
1. Run Elementary's Daily PPA on Ubuntu 20.04 base
2. Get a notification and have multiple monitors
3. Become slightly demoralized when the notification spans onto the monitor right of the primary monitor

### Expected behavior
Notifications render properly in the top right of the screen

### Screenshots or screen recordings
Screenshot from 2020-05-27 22 26 22

### Logs
N/A

## Platform Information
Screenshot from 2020-05-27 22 30 43
- [x] I'm using the daily PPA on Ubuntu 20.04
- [ ] I'm using the latest released stable version

## Additional context

I am aware the setup I am running is not "supported", but it seems this is not an issue caused by my setup; rather, a bug in how notifications are rendered. I figured this is an issue with window management, so I opened the issue here. I apologize in advance if this is the wrong place to open an issue for this bug.

davidmhewitt:
Fixes #821 

Instead of using a constant value for the notification width, lets calculate it based on the resolution of the monitor and the width of the notification.

### DIFF
--- a/src/NotificationStack.vala
+++ b/src/NotificationStack.vala
@@ -28,7 +28,6 @@ public class Gala.NotificationStack : Object {
 
     private const int WIDTH = 300;
 
-    private int stack_x;
     private int stack_y;
     private int stack_width;
 
@@ -84,7 +83,12 @@ public class Gala.NotificationStack : Object {
          */
         update_positions (animate, notification.height);
 
-        move_window (notification, stack_x, stack_y + TOP_OFFSET + ADDITIONAL_MARGIN * scale);
+        var primary = display.get_primary_monitor ();
+        var area = display.get_workspace_manager ().get_active_workspace ().get_work_area_for_monitor (primary);
+
+        int notification_x_pos = area.width - window.get_frame_rect ().width;
+
+        move_window (notification, notification_x_pos, stack_y + TOP_OFFSET + ADDITIONAL_MARGIN * scale);
         notifications.insert (0, notification);
     }
 
@@ -95,7 +99,6 @@ public class Gala.NotificationStack : Object {
         var scale = Utils.get_ui_scaling_factor ();
         stack_width = (WIDTH + MARGIN) * scale;
 
-        stack_x = area.x + area.width - stack_width;
         stack_y = area.y;
     }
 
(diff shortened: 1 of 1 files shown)


src/NotificationStack.vala
```diff
@@ -84,7 +83,12 @@ public class Gala.NotificationStack : Object {
          */
         update_positions (animate, notification.height);
 
-        move_window (notification, stack_x, stack_y + TOP_OFFSET + ADDITIONAL_MARGIN * scale);
+        var primary = display.get_primary_monitor ();
+        var area = display.get_workspace_manager ().get_active_workspace ().get_work_area_for_monitor (primary);
+
+        int notification_x_pos = area.width - window.get_frame_rect ().width;
```
JoseExposito, line 89: In my setup I have 2 monitors: |2|1|
At the moment the notification is being showed in my secondary monitor, this should fix it:

```suggestion
        int notification_x_pos = area.x + area.width - window.get_frame_rect ().width;
```

src/NotificationStack.vala
```diff
@@ -95,7 +99,6 @@ public class Gala.NotificationStack : Object {
         var scale = Utils.get_ui_scaling_factor ();
         stack_width = (WIDTH + MARGIN) * scale;
 
-        stack_x = area.x + area.width - stack_width;
```
JoseExposito, line 98: More of a question, can't the X coordinate of the notification be calculated here?

JoseExposito (changes requested):
Just a little bug, other than that pretty solid.

src/NotificationStack.vala
davidmhewitt: My thinking was that it's possible (but unlikely) for different notifications to be different widths, so I wanted to account for that. `update_stack_allocation` is only called when the size/layout of the monitors is changed.

JoseExposito (approved):
Looking great :+1:

PR #1041, opened 2021-02-03
Ignore ibus type source when setting keymap

jeremypw:
This is required to fix an issue with https://github.com/elementary/wingpanel-indicator-keyboard/pull/96 and https://github.com/elementary/switchboard-plug-keyboard/pull/341.

Once ibus methods are integrated into org.gnome.desktop.input-sources.sources Gala should ignore changes to org.gnome.desktop.input-source.current when current is an IBus method.  Otherwise IBus will not use the expected (last used) keyboard map.


davidmhewitt (approved):
The code here makes sense as everything that follows is xkb specific. I haven't done any testing, but happy for this to be merged if others have.

davidmhewitt:
Additionally, please can you propose the same changes against the Greeter compositor? It also uses the same class.

ryonakano (approved):
I tested this branch and confirmed my keyboard layout is detected as expected with ibus-mozc. The same issue happens in Greeter too so we need this fix also in Greeter as @davidmhewitt said.

PR #1048, opened 2021-02-12
Faster workspace switch animation when using keyboard shortcuts

JoseExposito:
Restore the default workspace switch animation time.

The animation was incremented to 400 ms and it should be 300 ms if we want to be consistent with stable.


danrabbit (approved):
Snappy!

PR #1051, opened 2021-02-13
Multitasking view 1:1 switch desktop animation with scroll events

### LINKED ISSUE #993: Touchégg workspace gestures stop working in Multitasking View
## What Happened

I three-finger swipe up to go into the Multitasking View, then try to three-finger swipe left or right to switch workspaces.

## Expected Behavior

I expected three-finger swipes left/right to work in the Multitasking View.

## Steps to Reproduce

1. Enter the Multitasking View (via three-finger swipe, keyboard shortcut, or .desktop file)
2. Three-finger swipe left or right to a new workspace

## Platform Information

elementary OS 6 daily build, updated as of today.

JoseExposito:
Partially solves  #993

When the multitasking view is opened, the switch workspace animation is not 1:1 at the moment.
As a user, I'd expect this view to respond to scroll events (like in libhandy) and also to my configured multi-touch gesture (3 or 4 fingers swipe).

At the moment we don't have a way to handle scroll events, so my goal here is to unify how scroll events and multi-touch events are handled in a single interface (`GestureTracker`) that will replace the current `GestureAnimationDirector`.
`GestureTracker` allows to configure different backends, at the moment scroll and in a follow up branch, Touchégg. In the future we can add touch and, when Wayland is ready, get rid of Touchégg.

The diff is already quite big, so I'm going to split this in 2 pull request:

 1 - This PR. Add support for smooth scrolling, already works on its own
 2 - In a follow up PR I'll unify how multi-touch events are handled

### DIFF
--- a/src/Gestures/Gesture.vala
+++ b/src/Gestures/Gesture.vala
@@ -17,13 +17,6 @@
  */
 
 namespace Gala {
-    public enum GestureType {
-        NOT_SUPPORTED = 0,
-        SWIPE = 1,
-        PINCH = 2,
-        SCROLL,
-    }
-
     public enum GestureDirection {
         UNKNOWN = 0,
 
@@ -38,16 +31,10 @@ namespace Gala {
         OUT = 6,
     }
 
-    public enum DeviceType {
-        UNKNOWN = 0,
-        TOUCHPAD = 1,
-        TOUCHSCREEN = 2,
-    }
-
     public class Gesture {
-        public GestureType type;
+        public Gdk.EventType type;
         public GestureDirection direction;
         public int fingers;
-        public DeviceType performed_on_device_type;
+        public Gdk.InputSource performed_on_device_type;
     }
 }
--- a/src/Gestures/Gesture.vala
+++ b/src/Gestures/Gesture.vala
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2020 elementary, Inc (https://elementary.io)
+ *           2020 José Expósito <jose.exposito89@gmail.com>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+namespace Gala {
+    public enum GestureType {
+        NOT_SUPPORTED = 0,
+        SWIPE = 1,
+        PINCH = 2,
+        SCROLL,
+    }
+
+    public enum GestureDirection {
+        UNKNOWN = 0,
+
+        // GestureType.SWIPE and GestureType.SCROLL
+        UP = 1,
+        DOWN = 2,
+        LEFT = 3,
+        RIGHT = 4,
+
+        // GestureType.PINCH
+        IN = 5,
+        OUT = 6,
+    }
+
+    public enum DeviceType {
+        UNKNOWN = 0,
+        TOUCHPAD = 1,
+        TOUCHSCREEN = 2,
+    }
+
+    public class Gesture {
+        public GestureType type;
+        public GestureDirection direction;
+        public int fingers;
+        public DeviceType performed_on_device_type;
+    }
+}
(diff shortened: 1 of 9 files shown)


src/Gestures/Gesture.vala
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright 2020 elementary, Inc (https://elementary.io)
+ *           2020 José Expósito <jose.exposito89@gmail.com>
```
meisenzahl, line 3:
```suggestion
 * Copyright 2021 elementary, Inc (https://elementary.io)
 *           2021 José Expósito <jose.exposito89@gmail.com>
```

Happy new year 😜

JoseExposito: Happy new year! 🤣 Still work in progress, I wish this was the only issue with this PR at the moment 🐛

src/WindowManager.vala
```diff
@@ -496,7 +501,7 @@ namespace Gala {
 
         private void play_nudge_animation (Meta.MotionDirection direction) {
             int duration = 360;
```
danrabbit, line 503: Looks like this is no longer used :)
```suggestion
```

src/Gestures/Gesture.vala
```diff
@@ -0,0 +1,53 @@
...
+        // GestureType.PINCH
+        IN = 5,
+        OUT = 6,
+    }
+
+    public enum DeviceType {
+        UNKNOWN = 0,
+        TOUCHPAD = 1,
+        TOUCHSCREEN = 2,
+    }
```
danrabbit, line 45: We could probably instead use https://valadoc.org/gdk-3.0/Gdk.InputSource.html right?

src/Gestures/Gesture.vala
```diff
@@ -0,0 +1,53 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+namespace Gala {
+    public enum GestureType {
+        NOT_SUPPORTED = 0,
+        SWIPE = 1,
+        PINCH = 2,
+        SCROLL,
+    }
```
danrabbit, line 25: Could we use https://valadoc.org/gdk-3.0/Gdk.EventType.html for this?

danrabbit (changes requested):
Not a thorough code review, but we can probably use some of the constants from Gdk instead of inventing our own here :)

src/Gestures/Gesture.vala
JoseExposito: True. I was keeping those for compatibility with Touchégg, but it makes more sense to transform into common types in the Touchégg backend.

src/WindowManager.vala
```diff
@@ -100,6 +100,11 @@ namespace Gala {
         private bool animating_switch_workspace = false;
         private GestureAnimationDirector gesture_animation_director;
 
+        /**
+         * Amount of pixels to move on the nudge animation.
+         */
+        public const float NUDGE_GAP = 32;
```
danrabbit, line 106: Does this need to be multiplied by scale?

danrabbit (commented):
From an interaction perspective this is :man_cook: :kiss:. The only thing I noticed is that the nudge is different when using the keyboard vs scrolling. It looks like the keyboard nudge actually moves the switcher at the bottom as well (which personally I don't think makes sense). Maybe there's an opportunity to DRY these a bit? If it would be too complicated, it can wait for another branch, just something I noticed

Would love some feedback from @Exalm on the gesture object

Great work!

src/WindowManager.vala
JoseExposito: I thought exactly the same, but it is not multiplied in the workspace switch animation (in `WindowManager.vala`), so I leave it the same way. However, I suspect it is a bug in the original implementation. Should I fix it in both places?

JoseExposito:
> The only thing I noticed is that the nudge is different when using the keyboard vs scrolling. It looks like the keyboard nudge actually moves the switcher at the bottom as well

That's actually a bug not related with functionality, and it is a little bit to complicated to fix it in this PR.
For that animation Gala is bumping the whole UI group instead of creating the required clones and animating them.
If you try the animation with multiple monitors connected, you will notice than the adjacent monitor will move into the main monitor.

> Would love some feedback from @Exalm on the gesture object

He will find `ScrollBackend`'s code pretty familiar :laughing:

src/WindowManager.vala
danrabbit: That would be nice if it's not too much trouble :)

Exalm:
FTR we've landed the new swipe model in both libhandy and gnome-shell by now, so we have this now:

https://www.youtube.com/watch?v=Y6WTOt4HLsI

When long swipes are enabled, it's possible to swipe many pages at once, when disabled, swipes can be responsive even when you do quick consecutive swipes - that was a problem in the past.

src/WindowManager.vala
JoseExposito: Done :D

JoseExposito:
Yes, I saw your video and your MR in GNOME Shell, that's pretty cool!! Specially for touchscreens and mobile :rocket: 

I think that Gala should match the behaviour of libhandy, so if long swipes are going to be enabled and activated on Odin, I'll have to implement them.

danrabbit (approved):

PR #1055, opened 2021-02-17
Plugins/MaskCorners: Code style cleanups

danrabbit:
View with hide whitespace changes

* No `using` namespace
* Explicit scope
* brackets


JoseExposito (approved):
It looks much better :rocket: There is a linter warning in line 128 in case you want to solve it... But it was already there, so this is an improvement.

Issue #955, opened 2020-11-02
Add Take Screenshot to the header context menu


danrabbit:
## Problem

It’s a long multi-step process to open the screenshot app, select the window capture mode, then ensure the correct window is focused when the screenshot occurs 

## Proposal

Add a menu item “Take Screenshot” (or “Capture” I’m not sure which language we typically use) to the header context menu

Cassidy pointed out that this is also a natural place to expose the keyboard shortcut for window captures

cassidyjames:
We use “Take Screenshot” as the action in Screenshot, and the term “grab” for the different modes in Screenshot—i.e. “Grab the current window”. For the app actions, it’s just “Screenshot of…”

I feel like “take” or “capture” work equally well here, but we should update the Screenshot app itself to be more consistent with what we decide (https://github.com/elementary/screenshot/issues/195).

Issue #137, opened 2017-12-08
Send a persistent notification after taking a screenshot


danirabbit:
This is a migration of https://github.com/elementary/screenshot-tool/issues/30

After taking a screenshot, Gala should send a persistent notification using the screenshot preview as the notification image. Clicking the notification should open the image in the default image viewer.

This would give:
1. A visual queue that the screenshot was completed, useful when sound is disabled or the user is not able to hear
2. A way to quickly view/edit/share the screenshot without having to know where screenshots are stored on the filesystem

peteruithoven:
I still also like the idea of being able of dragging and dropping that preview: https://github.com/elementary/screenshot-tool/issues/29 (which was also implemented in KDE Plasma 5.9).

GranPC:
Are there any plans to make this optional/configurable, even if it's just a gsettings key that's not exposed in the UI? I personally find the notification kinda annoying and redundant.

PR #1068, opened 2021-03-02
Add Take Screenshot to the header context menu

JoseExposito:
Fix https://github.com/elementary/gala/issues/955

There are some design decisions that need to be taken, so invoking @cassidyjames here:

 - We need to decide if we want to flash the window when the screenshot is taken or not
 - The screenshot sound is not being played, but the screenshot app doesn't play it for window screenshots :thinking: 
 - Maybe we should send a notification when the screenshot is taken indicating the path it is saved at. Extra points if there is a way to open Files when clicking on the notification
 - Menu order:
Screenshot from 2021-03-02 07 53 30

And one quick technical question: Is there a way to share the code that sets the file name with the screenshots app? I'd like to move the generation of the string `_("Screenshot from %s").printf (date_time);` to the D-Bus interface and share that code.

### DIFF
--- a/daemon/MenuDaemon.vala
+++ b/daemon/MenuDaemon.vala
@@ -38,6 +38,7 @@ namespace Gala {
         private Granite.AccelLabel move_right_accellabel;
         private Granite.AccelLabel on_visible_workspace_accellabel;
         private Granite.AccelLabel resize_accellabel;
+        private Granite.AccelLabel screenshot_accellabel;
         Gtk.Menu? window_menu = null;
         Gtk.MenuItem hide;
         Gtk.MenuItem maximize;
@@ -48,6 +49,7 @@ namespace Gala {
         Gtk.MenuItem move_left;
         Gtk.MenuItem move_right;
         Gtk.MenuItem close;
+        Gtk.MenuItem screenshot;
 
         // Desktop Menu
         Gtk.Menu? desktop_menu = null;
@@ -58,9 +60,11 @@ namespace Gala {
         ulong on_visible_workspace_sid = 0U;
 
         private static GLib.Settings keybind_settings;
+        private static GLib.Settings media_keys_settings;
 
         static construct {
             keybind_settings = new GLib.Settings ("org.gnome.desktop.wm.keybindings");
+            media_keys_settings = new GLib.Settings ("org.gnome.settings-daemon.plugins.media-keys");
         }
 
         [DBus (visible = false)]
@@ -169,6 +173,14 @@ namespace Gala {
                 perform_action (Gala.ActionType.MOVE_CURRENT_WORKSPACE_RIGHT);
             });
 
+            screenshot_accellabel = new Granite.AccelLabel (_("Take Screenshot"));
+
+            screenshot = new Gtk.MenuItem ();
+            screenshot.add (screenshot_accellabel);
+            screenshot.activate.connect (() => {
+                perform_action (Gala.ActionType.SCREENSHOT_CURRENT);
+            });
+
             close_accellabel = new Granite.AccelLabel (_("Close"));
 
             close = new Gtk.MenuItem ();
@@ -186,6 +198,7 @@ namespace Gala {
             window_menu.append (on_visible_workspace);
             window_menu.append (move_left);
             window_menu.append (move_right);
+            window_menu.append (screenshot);
             window_menu.append (close);
             window_menu.show_all ();
         }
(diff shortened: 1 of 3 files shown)


JoseExposito:
Those are a lot of extra points xD I don't even know if that is possible to implement

danrabbit:
It would probably be good to flash the screen and make sure we play the screenshot sound any time a screenshot is taken, but that probably doesn't need to be done in the scope of this branch.

The menu order is indeed an awful mess. I think that could be tackled in another branch too. We need some separators in there. It's ugly lol

I agree that we should send a notification after taking screenshots. There's another issue report here: https://github.com/elementary/gala/issues/137

I think that could wait for another PR, but it would definitely be a big improvement. Likewise, DnD for images in screenshots would be an issue for the notifications server and it is filed here: https://github.com/elementary/notifications/issues/83

If we did that, I think it would make sense to remove the custom save dialog from the screenshot app and always use the notification method to handle saved screenshots. That would implicitly de-duplicate that file name code as well ;)

danrabbit:
Ah actually RE: the file name, when using keyboard shortcuts I think it's actually GNOME Settings Daemon that handles this file name. So I'm not sure if there's a good way to de-duplicate it in all cases

danrabbit (changes requested):
It looks like taking screenshots with the keyboard shortcuts does send sounds, so my guess is that this is implemented in GNOME Settings Daemon, which means we probably need to implement it here too and we may not be able to always send sounds or we'll get double sounds with some clients

hanaral:
Just a heads up - the names the gnome-settings-daemon uses do not at all align with the Screenshot app's own naming scheme in different locales. I should probably open an issue for that.

JoseExposito:
> It would probably be good to flash the screen 

Well, at the moment it flashes only the window, because it is what `screenshot_window` allows to do, but it can be changed if you don't like it.

> make sure we play the screenshot sound any time a screenshot is taken, but that probably doesn't need to be done in the scope of this branch.

The sound is being played here:
https://github.com/elementary/screenshot/blob/master/src/ScreenshotBackend.vala#L158

Canberra is already a dependency of Gala, so, personally, I'd move that code to `org.gnome.Shell.Screenshot` as an extra method.
Adding it as an extra param  to the `screenshot_*` methods (`bool play_sound`) will break the compatibility with [GNOME's](https://github.com/vinzenz/gnome-shell/blob/master/data/org.gnome.Shell.Screenshot.xml) implementation, so I think it is safer to just add a new method.

> The menu order is indeed an awful mess. I think that could be tackled in another branch too. We need some separators in there. It's ugly lol

+1 :laughing: If you have ideas I implement it in this PR, it's simple.

> I agree that we should send a notification after taking screenshots. There's another issue report here: #137

Cool, I think that the same logic applies here. I'd add a method that wraps the notification sending so we don't end up with duplicated logic.

> It looks like taking screenshots with the keyboard shortcuts does send sounds, so my guess is that this is implemented in GNOME Settings Daemon, which means we probably need to implement it here too and we may not be able to always send sounds or we'll get double sounds with some clients

I've been testing a little bit and it seems like the shortcuts that are copying images to clipboard are not playing sounds. However, shortcuts that are saving to file are plying it.
Do you know who is handling those shortcuts? GNOME settings daemon?

I see some references to clipboard in [io.elementary.screenshot](https://github.com/elementary/screenshot/blob/master/src/Application.vala#L40), but I'm not sure about how that works.

What do you think? I don't know what's the scope of `io.elementary.screenshot`. If it is mean to always run in a Pantheon sessions, it is safe to move some code to Gala. However, if the goal of the project is to also work on a GNOME environment, then we'll need to duplicate some code...

danrabbit:
@JoseExposito Ah yes sorry, flash window I mean! I think that's working as intended :)

Yeah we're already breaking compatibility with the GNOME interface because of the `with cursor` and the `conceal_text`/`redact` options, so I don't think it would be that big of a deal to add a `play_sound` param.

I think it would probably be best not to let the scope of the PR get out of control. So it's probably fine the way it is already right now and we can come back in future PRs to do things like clean up the menu, add a sounds param, etc.

Yes that's correct, keyboard shortcut handling is done in GNOME Settings Daemon afaik

I think we are carrying maybe some code in Screenshot just to for example hide the conceal text option if it's not available in the server, but I don't think we really expect to cater to running elementary apps outside of elementary OS

JoseExposito:
> I think it would probably be best not to let the scope of the PR get out of control. So it's probably fine the way it is already right now and we can come back in future PRs to do things like clean up the menu, add a sounds param, etc.

Fair enough, we can always clean up a little bit duplicated code in the future. Feel free to merge this as it is then @danrabbit  😄 

I'll do some testing with the portal as well as part of the clean up.

danrabbit (approved):
Does exactly what it says it does! We can follow up with some more enhancements in future branches :)

PR #1072, opened 2021-03-02
Do not recalculate the multitasking view layout when a window not present in that layout is removed

### LINKED ISSUE #1067: Weird behavior in trackpad gestures
## What Happened

deepin-screen-recorder_Select area_20210302001406

## Expected Behavior

hide the dock

## Steps to Reproduce

1. Put the mouse on dock
2.3 finger swipe up
3.see

## Platform Information

6.0 Daily

JoseExposito:
Fix #1067

The multitasking view layout was recalculated every time a window was destroyed, even when the window wasn't displayed in the multitasking view. This caused animation glitches.


JoseExposito:
6 commits for one line :facepalm: Turns out, I can not `git commit --amend`

danrabbit (approved):
This makes sense. We don't need to reflow unless we actually call `remove_child`

Issue #798, opened 2020-05-12
Show window titles in Multitasking View


snowparrot:
## Prerequisites
- [x] I have searched open and closed issues for duplicates.

## Feature
**Is your feature request related to a problem? Please describe.**

Recently I had lots of pdf viewers open and they showed similar pdf files. I wanted to choose with SUPER + DOWN (window overview? ) but I couldnt because they look so similar.

**Describe the solution you'd like**

Show a tooltip with the window's title so I can distinguish them better. 
**Existing work**

**Describe alternatives you've considered**

On Windows they show the window's title below the picture of the window. 
**Additional context**

cassidyjames:
Yeah, I've wondered about exposing the window title somewhere as well. A tooltip could work to keep the UI looking the same, though it's less discoverable than just putting a label in the UI.

mcclurgm:
I think that this should absolutely be a feature. Without it, I have to regularly resort to guesswork switching windows (many times every day), which is not a great experience. Some arguments both ways on the labeling style for your consideration:
- I think that a plain label would be more helpful than a tooltip for scanning. That way I don't have to hover my mouse over every candidate window, although it would require some work to make these labels readable. The simple solution would be to give them a solid background, but I'm not sure that's something we want to do.
- The tooltips don't provide this information at a glance, but if we make them responsive they can be fairly fast. This would be less cluttered, since every window would have to have a label in the other version. If there are a lot of windows on the desktop, that would be a lot of text, and would force each preview to be even smaller. IMO if we implement this version, we should make sure that the tooltips appear instantly, so I'm not waiting for a fraction of a second over each window I want to check. It would also make the labels discoverable, since they'd appear every time someone uses the multitasking view.

FWIW, GNOME uses permanent labels and macOS uses tooltips.

Akryum:
An example when the multitasking view is difficult to use, happening to me very frequently:
![image](https://user-images.githubusercontent.com/2798204/107710949-59740700-6cc7-11eb-8d72-5946bace7219.png)

JoseExposito:
@Akryum that's me every day :rofl: 

I'm putting something together, it is still in a early stage, but I'll open a PR tomorrow so @cassidyjames can make his magic with the design...

![tooltips](https://user-images.githubusercontent.com/1335948/109853133-e02c5c00-7c55-11eb-8170-78e81ea494a4.png)

cassidyjames:
[...]

PR #1076, opened 2021-03-04
Show window titles in Multitasking View

JoseExposito:
Fix https://github.com/elementary/gala/issues/798

Add window titles in multitasking view in a tooltip like widget.

The design is not clear yet, or when the title must be always visible or only on mouse over, that's why this is still a draft PR.

Here are some ideas I tried:

icon-tooltip

tooltip-icon

center


Akryum:
I like the 2nd option

hanaral:
It may overcomplicate things a little, but I think a good compramise would be to have the name reveal itself from the icon of the window you're hovering over
A bit like the revealer sub-text that modelbuttons can have:

https://user-images.githubusercontent.com/58219504/110054371-e5230580-7d52-11eb-955c-4dddd8abc415.mp4

I dont mean any fancy effects - the text should just unhide and push the icon down a little (and in reverse when the focus changes)

JoseExposito:
Personally, I like option 1 the most because it is consistent with how the icons are displayed in the applications menu. That could be a cool animation to add if we decide to show the titles on over.

That's the key thing, deciding if titles should be always visible or not. On the one hand, some times you need to see the title of the window to differentiate it, for example having multiple IDE window open. On the other, my guess is that most users don't need this that often and showing it on over is just OK.

It's a UX think, so, let's the experts decide xD

hanaral:
Well hovering over the window is far more direct than hovering over an icon, which was what I was proposing. It would make it more consistent with fovering over icons in the dock - But yeah the UX team needs to take a look

Akryum:
I strongly vote for always showing the title for accessibility and ease of use.

martinstork:
I prefer the last option with the centered labels.

cassidyjames:
Whoops, I meant to comment on this earlier. I think to start, it might make sense to basically treat them as tooltips. Keep the existing icon location, but show the window title when hovering the window. As for the exact location, I'm not 100% sure but I think I would expect it either below the mouse cursor (like an actual tooltip), or centered beneath the icon if we want a static/consistent location.

JoseExposito:
I can not tell the design reasons behind it, but GNOME's grid has a bigger gap and, one of the reason, could be to display the window title.

Gala uses as much space as possible to display the window previews so showing the title like in GNOME will actually be a bigger change to the layout. We'd also have to make the window preview smaller and I don't think we wan that.
Here is an example of what I mean:

do-not-move-the-icons

As you can see the title overlaps others windows space and also the icon group at the bottom.


I gave a try to the animation Hanaral suggested, here is a demo:

https://user-images.githubusercontent.com/1335948/110207554-8b225d00-7e84-11eb-9889-d8a06b90b925.mp4

I feel like it is "cool" the first time you use it, but there is too much movement going on and you might get tired of it quickly. What do you think?
Let me know if you want me to push the changes to show the title on mouse over in case you want to test it.

martinstork:
Looks slick, but might be frustrating. Being able the quickly glance without having to hover individual windows takes less effort.

Have you tried placing the label on top of the windows (where a header bar would be)?

Another option could be placing app icons in the center and the labels at the bottom.

hanaral:
O.K, so how about just making the window paler and showing the overlay in the center (when hovered)? That way it looks a lot like normal overlay/accel(which one is correct?) and has good contrast every time thanks to the window being whiter.

cassidyjames:
@JoseExposito the animation is cute but maybe to start we center the tooltip above the icon; that would require the least amount of reworking things and the least disruption to the current design.

Also for everyone here, I want to be careful to avoid too much design-by-committee. I would prefer this to be a small iterative change, and then we can spend some design cycles actually thinking through this and other changes (like #998) in conjunction.

JoseExposito:
@cassidyjames title centered above the icon and only visible on mouse over, ready for code review and UX input.

JoseExposito:
By the way, I also fixed my personal pet peeve with this animation: The close icon was visible during the open/close animation in the window under the cursor.
Leaving that and adding the title to the equation, was too much for me :laughing:

cassidyjames:
@JoseExposito do you know why text would look blurry/non-pixel-aligned here?

Screenshot from 2021-03-12 12 17 01

hanaral:
@JoseExposito I'd like to add that a lot of the app icons in multitasking seem blurry to me seemingly at random - I can't reproduce it though

hanaral:
Thanks for the changes! It might be a bit much for this PR but I think it might be good to experiment with dimming the window previews a bit as well (to fix the issue of them looking the same as when you can interact with them)
I'll tag you in the original issue if it's out of scope for this PR

danrabbit (changes requested):
Don't forget to multiply values by scale for HiDPI :)

src/Widgets/Tooltip.vala
```diff
@@ -0,0 +1,119 @@
...
+ * Clutter actor to display text in a tooltip-like component.
+ *
+ * Style constants from:
+ * https://github.com/elementary/stylesheet/blob/master/src/widgets/_tooltips.scss
+ */
+public class Gala.Tooltip : Clutter.Actor {
+    private const double BACKGROUND_COLOR = (26 / 255); // #1a
+    private const double BACKGROUND_OPACITY = 0.9;
+    private const double BACKGROUND_BORDER_RADIUS = 3;
+    private const string COLOR = "#ffffff";
```
donadigo, line 29: This definitely is not needed. We should not store a string just to parse it everytime we redraw the tooltip. You can statically get an instance to a white color via `Clutter.Color.get_static (Clutter.StaticColor.WHITE)`.

src/Widgets/Tooltip.vala
```diff
@@ -0,0 +1,119 @@
...
+    private const int PADDING_RIGHT = 6;
+
+    public string text {
+        get {
+            return _text;
+        }
+        set {
+            _text = value;
+            draw ();
+        }
```
donadigo, line 42: I'm generally not a fan of properties that do a lot of work when getting / setting them, because it makes it harder to follow what the code is actually doing. You may only call `draw` here but `draw` in itself creates new objects and updates the Actor's children list so it is a lot for just setting a property. Instead, these should be standard properties and the setter should decide whether to call `draw ()` or not.

src/Widgets/Tooltip.vala
```diff
@@ -0,0 +1,119 @@
...
+    }
+    private string _text = "";
+
+    public float max_width {
+        get {
+            return _max_width;
+        }
+        set {
+            _max_width = value;
+            draw ();
```
donadigo, line 52: Same thing here.

src/Widgets/Tooltip.vala
```diff
@@ -0,0 +1,119 @@
...
+        // First set the text
+        remove_child (text_actor);
+
+        text_actor = new Clutter.Text () {
+            color = Clutter.Color.from_string (COLOR),
+            x = PADDING_LEFT,
+            y = PADDING_TOP,
+            ellipsize = Pango.EllipsizeMode.END,
+            use_markup = true
+        };
```
donadigo, line 89: Why is it necessary to entirely recreate the text actor? Ideally, the Clutter.Text actor should be only instantiated only once when `Tooltip` constructs and then you'd set the text accordingly here (or even in the setter of `text`).

src/Widgets/WindowClone.vala
```diff
@@ -510,7 +530,17 @@ namespace Gala {
                     close_button.x = -close_button.width * 0.5f;
                     break;
             }
+
+            bool show = has_pointer && !in_slot_animation;
+            close_button.opacity = show ? 255 : 0;
+            window_title.opacity = show ? 255 : 0;
+
+            window_title.text = window.get_title ();
+            window_title.max_width = dest_width - (TITLE_MAX_WIDTH_MARGIN * scale_factor);
```
donadigo, line 539: These two sets alone will create the text actor, remove it and add it to the tooltip twice causing unnecessary work we can avoid. I would simply declare the `draw` method public or declare an `update` method that takes the text and max width and have it call draw once.

donadigo (changes requested):
I have comments.

src/Widgets/Tooltip.vala
JoseExposito: I prefer to keep the string because, as this is copy and pasted from the tooltip stylesheet CSS, it makes it easier to change it in the future if a different kind of white is used. Moved to the `static construct` though:
https://github.com/elementary/gala/pull/1076/files#diff-ed4232dda49f45599a7a209514eb953072ee8cc2515630daa9bef8719be1607cR63

src/Widgets/Tooltip.vala
JoseExposito: I don't like making public the draw method because then it is easy to forget calling it.
Ideally I'd like to set properties in the child text actor and avoid recreating it, but I have some issues (I'll add a comment about that in the appropriate thread).

For the moment I changed the method to a set with an optional parameter to redraw or not the tooltip so we can scratch a couple of milliseconds if required:
https://github.com/elementary/gala/pull/1076/files#diff-ed4232dda49f45599a7a209514eb953072ee8cc2515630daa9bef8719be1607cR81

src/Widgets/Tooltip.vala
JoseExposito: Same here: https://github.com/elementary/gala/pull/1076/files#diff-ed4232dda49f45599a7a209514eb953072ee8cc2515630daa9bef8719be1607cR89

src/Widgets/WindowClone.vala
JoseExposito: This is optimized now.

JoseExposito:
@danrabbit yup, I always forget about it. As soon as we support Wayland, I'm going to buy a 4k display :laughing: 

@donadigo about why I'm recreating the text actor...

My first implementation was very similar to what you suggest in (1) (redrawing the background when required, but overall the code looked similar).

However, if you try to run the code you suggested, you'll notice that, once the ` text_actor.width` is set it is not possible to adjust it anymore. Let me explained with an example:

- The tooltip is created with a text longer than the window preview
- You set the max width so the tooltip fits into the preview
- Now the title of the window changes and it is short enough to fit the window
- If you try to get the width of the text to check `if (text_actor.width > max_width)` and adjust the tooltip but it returns the width you set in step 2

I didn't find an API to tell the text actor to ignore the width I manually set in step 2 and use it's preferred size, but ideally yes, I'd like to be able to implement option 1. Do you know how I can do that?

Option 2 could solve the problem, but then if the tooltip gets extra properties we'll have to include them there and the method signature will be a mess.

danrabbit (approved):
LGTM from a design perspective :)

JoseExposito:
@donadigo did you have a chance to look into my comment?

hanaral:
Really small nitpick, it's probably not really an issue as much as it is a consistency thing: the backdrop colour for the text labels is different from the elementary style of overlay labels in Gtk
2021-03-30 10 39 41 のスクリーンショット
Screenshot from 2021-03-30 10-38-55

danrabbit (approved):

src/Widgets/Tooltip.vala
```diff
@@ -55,16 +48,33 @@ public class Gala.Tooltip : Clutter.Actor {
     public float max_width;
 
     static construct {
-        var scale = InternalUtils.get_ui_scaling_factor ();
+        var dummy_label = new Gtk.Label ("") {
```
JoseExposito, line 51: Even though this is really convenient, it'll crash on Wayland, like it happens now with the tile outline preview.

danrabbit: I guess it would crash anywhere we're using the Gtk foreign drawing API then? Because we're using it in several places to get accent colors

src/Widgets/Tooltip.vala
```diff
@@ -0,0 +1,153 @@
...
+
+    static construct {
+        var dummy_label = new Gtk.Label ("") {
+            tooltip_text = "null"
+        };
+
+        unowned var label_style_context = dummy_label.get_style_context ();
+
+        var widget_path = label_style_context.get_path ().copy ();
+        widget_path.iter_set_object_name (-1, "tooltip");
```
donadigo, line 58: I know this has been done before like this in gala, but I'm pretty sure the path itself can be constructed without initializing a new label. This is fine though.

src/Widgets/Tooltip.vala
```diff
@@ -0,0 +1,153 @@
...
+        // Adjust the size of the tooltip to the text
+        width = text_actor.width + padding.left + padding.right;
+        height = text_actor.height + padding.top + padding.bottom;
+        background_canvas.set_size ((int) width, (int) height);
+
+        // And paint the background
+        background_canvas.invalidate ();
+    }
+
+    private bool draw_background (Cairo.Context cr, int width, int height) {
```
donadigo, line 141: This method can be static.

src/Widgets/WindowClone.vala
```diff
@@ -510,7 +530,17 @@ namespace Gala {
                     close_button.x = -close_button.width * 0.5f;
                     break;
             }
+
+            bool show = has_pointer && !in_slot_animation;
+            close_button.opacity = show ? 255 : 0;
+            window_title.opacity = show ? 255 : 0;
```
donadigo, line 536: Could simply do
```suggestion
            window_title.opacity = close_button.opacity;
```

donadigo (changes requested):
Some more comments.

src/Widgets/Tooltip.vala
danrabbit: Thanks for pushing me in that direction! fixed

donadigo (approved):
Looks good.

src/Widgets/Tooltip.vala
JoseExposito: Yes, using Gtk in the compositor causes the problem, not the foreign drawing API in particular. We have the same issue with the shadows, for example. I'd personally avoid it.

Issue #1031, opened 2021-01-27
Add "Tile" actions and shortcuts to window menu


cassidyjames:
Related to #1028

## Problem

It's not obvious from the window menu that you can tile windows left/right. We show it in the Keyboard Shortcuts overlay, but not contextually in the window menu, where it might be more obvious and connected to other window actions. 

## Proposal

Add a "Tile window left" and "Tile window right" actions to the window menu. This should probably also include a more general cleanup/reorganizing and grouping of window menu actions since there's a lot going on in there.

## Prior Art

macOS has a sort of context menu on their maximize button with tiling options:

![macos-big-sur-pages-split-screen](https://user-images.githubusercontent.com/611168/106038581-5e18a880-6095-11eb-9bcc-67d5a58e6f4e.jpg)

Chrome OS has a weird [click-and-hold on the maximize button](https://support.google.com/chromebook/answer/177891?hl=en) that then shows a left and right arrow for tiling.

It is [not clear if Windows has a context menu](https://support.microsoft.com/en-us/windows/snap-your-windows-885a9b1e-a983-a3b1-16cd-c531795e6241) for this, but they support dragging windows and keyboard shortcuts.

hanaral:
Should there be icons for it? macOS uses them despite not really using icons in its normal context menus.

PR #1080, opened 2021-03-13
Add "Tile" actions and shortcuts to window menu

meisenzahl:
Closes https://github.com/elementary/gala/issues/1031


JoseExposito (commented):
Just dropping a comment here to reflect the conversation we had on Slack.

At the moment it is not possible to tile a window programmatically because Mutter doesn't expose the tile API. I submitted a merge request [upstream](https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/1805) to make the required APIs public, so once it gets merged, we could do something like...

```
case ActionType.TILE_CURRENT_LEFT:
    current.tile (Meta.TileMode.LEFT);
    break;
```

danirabbit:

PR #1084, opened 2021-03-22
MenuDaemon: re-order window menuitems

danrabbit:
Tried to group these a little better. Open to suggestions

Also, thought it might be better to make the move items insensitive instead of visible to tie the relationship closer with that checkbox, but I can revert that change if it's weird

## BEFORE
Screenshot from 2021-03-22 12 59 16 Screenshot from 2021-03-22 12 59 37

## AFTER
Screenshot from 2021-03-22 12 58 05 Screenshot from 2021-03-22 12 58 14

### DIFF
--- a/daemon/MenuDaemon.vala
+++ b/daemon/MenuDaemon.vala
@@ -190,15 +190,18 @@ namespace Gala {
             });
 
             window_menu = new Gtk.Menu ();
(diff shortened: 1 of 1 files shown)


cassidyjames (approved):
I like it. Much cleaner and using sensitivity instead of visibility makes sense.

PR #1087, opened 2021-03-27
Fix #1074: Window menu says Unmaximize when tiled

### LINKED ISSUE #1074: Window menu says "Unmaximize" when tiled
## What Happened

If you tile a window, the window menu says "Unmaximize" as one of the menu items

## Expected Behavior

It should instead say "Untile"

## Steps to Reproduce

1. Tile a window
2. Secondary click the window handle area
3. See that the menu says "Unmaximize" instead of "Untile"

JoseExposito:
Fix https://github.com/elementary/gala/issues/1074:  Show "Untile" instead of "Unmaximize" in the window menu when a window is tiled.

Also, the current behaviour is to maximize the window when it is tiled. Fix it to restore the window when the "Untile" menu option is selected.

### DIFF
--- a/daemon/MenuDaemon.vala
+++ b/daemon/MenuDaemon.vala
@@ -218,7 +218,7 @@ namespace Gala {
 
             maximize.visible = Gala.WindowFlags.CAN_MAXIMIZE in flags;
             if (maximize.visible) {
-                string maximize_label;
+                unowned string maximize_label;
                 if (Gala.WindowFlags.IS_MAXIMIZED in flags) {
                     maximize_label = (Gala.WindowFlags.IS_TILED in flags) ? _("Untile") : _("Unmaximize");
                 } else {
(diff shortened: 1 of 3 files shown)


tintou (approved):

daemon/MenuDaemon.vala
```diff
@@ -218,7 +218,12 @@ namespace Gala {
 
             maximize.visible = Gala.WindowFlags.CAN_MAXIMIZE in flags;
             if (maximize.visible) {
-                var maximize_label = Gala.WindowFlags.IS_MAXIMIZED in flags ? _("Unmaximize") : _("Maximize");
+                string maximize_label;
```
tintou, line 221:
```suggestion
                unowned string maximize_label;
```

JoseExposito: Thanks Corentin :+1:

PR #1092, opened 2021-04-02
Fix #1029: Zoom using multi-touch gestures

### LINKED ISSUE #1029: Zoom using multitouch gestures
## Problem

Some people with low vision have reported that it would be nice to be able to access the zoom feature more quickly than with keyboard shortcuts.

## Proposal

It would probably be nice to be able to zoom smoothly using multitouch gestures such as a 3-finger pinch or a 4-finger push/pull

JoseExposito:
Fix #1029

Implement 3/4 finger pinch to zoom the entire desktop.

The changes are split in commits so it is easier to review:

- [Move the zoom plugin to the core code:](https://github.com/elementary/gala/commit/b814f987a7a96f8f2098d09aecdad158aa1fd3c4)
Just move the zoom plugin to `src`. This commit doesn't add new functionality.

- [Refactor the Zoom.zoom method to easily integrate gestures:](https://github.com/elementary/gala/commit/203f62bd1e5dee4b8e8f0fa2353bcd0c43b346f4)
A small refactor of the zoom class to easily integrate gestures in a future commit. This commit doesn't add new functionality.

- [Add GSettings to configure zoom gesture:](https://github.com/elementary/gala/pull/1092/commits/cf4236b437cdd63fc321ed841ae5e2d9ae33ca8d)
Add the required settings to be able to configure this from System Settings -> Mouse & Touchpad -> Gestures.
I'll add the required UI there once this gets merged.

- [Zoom using multi-touch gestures:](https://github.com/elementary/gala/commit/d18b1f55e0eb4bbbe64df35d64df6f4e10c4ebc5)
The actual implementation of the feature.


hanaral:
Some feedback:
Works really well with 3 fingers, but 4 fingers feel a little difficult to initiate with sometimes and feels like I need more spreads than 3 to zoom the same amount
_solution(?):_
might help if 4 finger pinching and spreading is more sensitive imo because the way you do 4 fingers takes up far more room than 3

Would it be possible it to take a little more force at the beginning than the rest and make it rubberband back unless you zoom more than X minimum? This way the feature isn't accidentally triggered and can't be unintentionally left semi-activated by people who may not be 'technically adept'.

JoseExposito:
Hi @hanaral 

Thanks for testing this :smile: 

Pinch is detected by libinput and, even though the version included on Odin works much better than the one included on Hera, there is still some room for improvement. The latest master works even better and [there is](https://gitlab.freedesktop.org/libinput/libinput/-/merge_requests/584) work in progress to make the gesture detection even better, so eventually, all of those features will hit elementary.

The 3 vs 4 fingers thing... I believe it depends a lot on the size of your touchpad. I have a Magic Trackpad 2, which is considerable big, and the gesture is comfortable to perform. However, on most touchpads, 3 fingers is the right number, that's why I set it as the default. I wouldn't mind to limit this feature to 3 fingers only honestly.

> Would it be possible it to take a little more force at the beginning than the rest and make it rubberband back unless you zoom more than X minimum? This way the feature isn't accidentally triggered and can't be unintentionally left semi-activated by people who may not be 'technically adept'.

That's a very good point from a usability perspective. Unlike other gestures, there is not an obvious way to revert the action in case it gets triggered accidentally. Maybe we should disable it by default?

cassidyjames:
I agree this should be disabled by default, and togglable in the Gesture settings.

src/WindowManager.vala
```diff
@@ -76,6 +76,11 @@ namespace Gala {
         Clutter.Actor? last_hotcorner;
         public ScreenSaverManager? screensaver { get; private set; }
 
+        /**
+         * Allow to zoom in/out the entire desktop.
+         */
+        Zoom? zoom = null;
```
danrabbit, line 82: Do we need to declare zoom here? It looks like it isn't ever called outside of `show_stage`

danrabbit (commented):
I only have one small comment, otherwise this works great! I think I agree we should go off-by-default though just because on some smaller trackpads it's a little hard to differentiate between 3 finger swipe and zoom.

danrabbit (approved):

olosegres:
> Hi @hanaral
> 
> Thanks for testing this 😄
> 
> Pinch is detected by libinput and, even though the version included on Odin works much better than the one included on Hera, there is still some room for improvement. The latest master works even better and [there is](https://gitlab.freedesktop.org/libinput/libinput/-/merge_requests/584) work in progress to make the gesture detection even better, so eventually, all of those features will hit elementary.
> 
> The 3 vs 4 fingers thing... I believe it depends a lot on the size of your touchpad. I have a Magic Trackpad 2, which is considerable big, and the gesture is comfortable to perform. However, on most touchpads, 3 fingers is the right number, that's why I set it as the default. I wouldn't mind to limit this feature to 3 fingers only honestly.
> 
> > Would it be possible it to take a little more force at the beginning than the rest and make it rubberband back unless you zoom more than X minimum? This way the feature isn't accidentally triggered and can't be unintentionally left semi-activated by people who may not be 'technically adept'.
> 
> That's a very good point from a usability perspective. Unlike other gestures, there is not an obvious way to revert the action in case it gets triggered accidentally. Maybe we should disable it by default?

What about to repeat how it implemented in macos? - not to trigger zoom if minimum distance of pinch not reached.

It's just that in our world (where touch devices is ubiquitous), this gesture is intuitive. If enabled by default, then new users can get a wow effect from this concern of developers.

PR #1108, opened 2021-04-11
Rewrite gesture schemas

### LINKED ISSUE #1093: Gsettings: use gestures as the key and actions as the value
## Problem

At the moment in gsettings we have 2 keys per action: a bool for enable/disable and an int for the number of fingers used. This is problematic in a few ways:

* Since we have a potentially infinite amount of actions (tile, maximize, multitasking, zoom, pip, lock, who knows), this means we'll keep adding more and more keys for each action
* There is no good way to allow actions to use multiple different types of gestures. For example, pinch to zoom is probably the most intuitive default, but a 4 finger push/pull also makes sense and might be more comfortable for some. Likewise, pinch might feel more natural for some to enter and exit multitasking. Swipe might make more sense for tiling, etc.

## Proposal

Instead we should structure the settings keys such that we have an enum of sensible actions for each gesture. For example:

- 3 finger swipe left/right
    - switch workspace
    - tile
- 3 finger pinch
    - zoom
    - multiasking view
    - maximize/restore
- 3 finger push/pull
    - zoom
    - multitasking view
    - maximize/restore  

This would also make it a lot easier to create a configuration UI that won't have multiple gestures assigned to a single action

danrabbit:
Fixes #1093 

Gsettings schemas now use gestures for the key and actions for the values. This should allow us to add many many more actions without adding many more keys. It also ensures that a given gesture cannot be assigned to more than one action. But an action can be assigned to several gestures.

We'll need to update Switchboard either in parallel or after this branch

### DIFF
--- a/data/gala.gschema.xml
+++ b/data/gala.gschema.xml
@@ -359,17 +359,5 @@
       <summary>Four-finger horizontal swipe gesture</summary>
       <description>The action that corresponds to performing a horizontal swipe gesture with four fingers</description>
     </key>
-
-    <key type="b" name="workspaces-gesture-enabled">
-      <default>true</default>
-      <summary>Switch workspace gesture</summary>
-      <description>If enabled, swipe left/right with the number of fingers set in io.elementary.desktop.wm.gestures.workspaces-gesture-fingers to switch between workspaces</description>
-    </key>
-    <key type="i" name="workspaces-gesture-fingers">
-      <default>3</default>
-      <range min="3" max="4"/>
-      <summary>Switch workspace gesture fingers</summary>
-      <description>Number of fingers used in the switch workspaces gesture</description>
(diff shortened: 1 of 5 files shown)


JoseExposito (changes requested):
Nitpick but I don't think we use `Gala.` anywhere else, ie, `Gala.GestureSettings`. Other than that it looks and works great :smile:

JoseExposito (approved):
I tested it and it work, also, the code looks great :+1: 

Feel free to merge it when you (or the switchboard plugin) are ready. Ping me if you need help with the switchboard plugin, I can make the changes there.

danrabbit:
@JoseExposito sure, if you'd like to work on that feel free! If not, I can do it later this week :)

PR #1115, opened 2021-04-25
Fix #1112: Adjust gap between notification bubbles

### LINKED ISSUE #1112: Giant size between notifications
When receiving several notifications together there is a giant space between the view 
I am on the latest version of ElementaryOS 6

Captura de pantalla de 2021-04-13 16-00-48 (copia)

JoseExposito:
The actor size was used to calculate the gap between the notificatification
bubbles. The actor was reporting a bigger height because the window
shadow is taken into account to calculate the height.

Use the window frame height instead of the actor height.


meisenzahl (approved):
Well done 🚀️

## Before

Bildschirmfoto von 2021-04-25 11-56-41

## After

Bildschirmfoto von 2021-04-25 11-57-21

PR #1119, opened 2021-04-30
Introduce a Launch-or-Focus Application Switcher

felix-andreas:
Supersedes #585

This PR introduces a new Application Switcher which let's the user bind any application to a <kbd>Launch-or-Focus</kbd> shortcut which ...

* ... launches the application, if it is not running.
* ... focuses the application, if it is running.

Pressing the <kbd>Launch-or-Focus</kbd> shortcut multiple times cycles through the different open windows of the application. In other words: an *application-specific* <kbd>Alt</kbd><kbd>Tab</kbd> shortcut.

Additionally  <kbd>Ctrl</kbd> + <kbd>Launch-or-Focus</kbd> always opens a new window of the application regardless if there is already another instance running. (similar to the dock, but this is not implemented yet)

Applications can be bound to a shortcut via the [system settings keyboard-plug](https://github.com/elementary/switchboard-plug-keyboard/pull/358). For demonstration purposes some default applications are bound to the following shortcuts:

| Application | Shortcut |
| --- | --- |
| Text Editor | <kbd>Super</kbd> <kbd>E</kbd> |
| File Manger | <kbd>Super</kbd> <kbd>M</kbd> |
| Web Browser | <kbd>Super</kbd> <kbd>B</kbd> |
| Terminal | <kbd>Super</kbd> <kbd>T</kbd> |

## Motivation

Most of the current possibilities of switching between different windows rely on some "state" which the user has to keep track of in order to build up muscle memory. Therefore switching windows, when working with many applications at the same time, can add a some cognitive workload to the user:

* For <kbd>Alt</kbd><kbd>Tab</kbd> the "state" would be the order of windows. It works great when toggling between two different application, but scales bad with increasing number of windows as the user has to keep track of the ordering in the window stack, which makes it hard to build up muscle memory.
*  For the **Mutlitasking View / Window Overview** the "state" would be the position of windows. As these change depending on which applications are running, it is also not possible to learn/build up muscle memory using a mouse or the <kbd>Super</kbd><kbd>Down</kbd> + <kbd>Arrow Keys</kbd> shortcuts.
* For **Workspaces** the "state" would be set of windows on the different workspaces. Learn/build up muscle memory for the <kbd>Super</kbd><kbd>Number</kbd> shortcuts, would require to setup applications so that they are always on the same workspace (see [Blockbuster](https://github.com/donadigo/blockbuster)). But his workflow might not be suitable for every user and the numbering also breaks down if one of the applications isn't running.
* The **dock** is in that sense "state-less", but has no keyboard shortcuts.

The goal of this PR is to introduce a <kbd>Launch-or-Focus</kbd> shortcut which is in that sense "state-less" that always leads to the same outcome: Regardless of whether the application is running or not, on which workspace its window is on and what position it has in the window stack, the <kbd>Launch-or-Focus</kbd> shortcut always brings the user to the application.

Due to the shortcut being "state-less", it possible for the user to learn/build up muscle memory for it. Switching between windows is a very fundamental operation when using a computer and I think that something like <kbd>Launch-or-Focus</kbd> shortcut could significantly increase the workflow for many people.

## Prior Art

* [jumpapp](https://github.com/mkropat/jumpapp) (does not work with Wayland)
* [Run-or-raise GNOME 3 extension](https://github.com/CZ-NIC/run-or-raise)

## ⚠️ Disclaimer ⚠️

Remove this plugin when switching back to the main branch: 

```rm /usr/lib/x86_64-linux-gnu/gala/plugins/libgala-launch-or-focus.so```

Otherwise gala crashes when logging in again!


jeremypw:
Sounds interesting but I am not sufficiently familiar with Gala coding to review this properly.

JoseExposito:
What do you think about this @elementary/ux ?

IMHO, this would be great to have, I'd use <kbd>⌘</kbd> <kbd>T</kbd> a lot :) However, I think that integrating this with [switchboard-plug-applications](https://github.com/elementary/switchboard-plug-applications) or at least [shortcut-overlay](https://github.com/elementary/shortcut-overlay) is mandatory to make shortcuts easy to discover.

jeremypw:
It might be easier to split this into two PRs - one dealing with standard apps and one dealing with custom shortcuts.  Then we can think about whether there is a way of not limiting the number of custom apps to 10 which seems rather ugly. E.g. is it not possible to use a setting that is a map of desktop ids to shortcuts rather than a separate setting for each custom app.

I'd like to get this merged at least for standard apps.

jeremypw:
Am I right in thinking that there is a separate shortcut for the launch-or-focus behaviour? This could be confusing. If this behaviour is desirable than maybe it should apply automatically to existing app shortcuts? Or have a toggle to turn it on/off globally?

andreasfelix:
> Am I right in thinking that there is a separate shortcut for the launch-or-focus behaviour? This could be confusing. If this behaviour is desirable than maybe it should apply automatically to existing app shortcuts? Or have a toggle to turn it on/off globally?

My intention was to make this the default behavior when opening apps using a shortcut. (See https://github.com/elementary/switchboard-plug-keyboard/pull/358)

andreasfelix:
> It might be easier to split this into two PRs - one dealing with standard apps and one dealing with custom shortcuts. 

OK! The last commit removed the ability to bind non-default applications.

> Then we can think about whether there is a way of not limiting the number of custom apps to 10 which seems rather ugly. E.g. is it not possible to use a setting that is a map of desktop ids to shortcuts rather than a separate setting for each custom app.

Yes, I must admit that my solution was not particularly elegant  :smile: . Using some kind of map which is not limit to 10 apps is probably possible. I have tried something like this in the commit https://github.com/elementary/gala/pull/585/commits/99b6eafe408aa58a0d012c9841d2f9c62fb2f819 of the old PR #585. I can't remember what was the problem, but it didn't quite work out. So I went with the ugly solution, which at least worked...

jeremypw:
The switchboard plug uses a relocatable schema to store info an arbitrary number of custom shortcuts (limited to 100 I assume for performance or key availability reasons).  Files uses a variant setting to store info on an arbitrary number of tabs to restore.

andreasfelix:
> The switchboard plug uses a relocatable schema to store info an arbitrary number of custom shortcuts (limited to 100 I assume for performance or key availability reasons). Files uses a variant setting to store info on an arbitrary number of tabs to restore.

Yes, but as we split up this PR this will be a concern of the next PR. Or?

jeremypw:
Yes, I think individual settings are acceptable for a fixed list of native apps.

PR #1135, opened 2021-05-12
Drag to launch

andreasfelix:
I just noticed that this branch got lost when I deleted my gala fork. 

Reopening this because its needed for https://github.com/elementary/dock/pull/62

The code is kinda messy and this could probably improved by rebasing this branch on https://github.com/elementary/gala/pull/963. But it works and should be good enough as a prototype.

I will leave it like this until a decision is made if this feature would be accepted.


danirabbit:

PR #1156, opened 2021-05-25
a11y: Delay dwell click timer animation

### LINKED ISSUE #1152: [Feature Request] Add delay to Long press Secondary Click cursor animation
## Describe the Issue
When Long-press secondary click is enabled, a very helpful circular wipe animation will appear to show how long we need to continue holding the click for. The long-press to secondary click will also be very helpful for those who are using eOS on a touch screen device.

However, this gets annoying really quickly when clicking, as the circular animation will start immediately upon the first click, regardless of whether the button is held. I think that this annoyance will especially affect those who are on devices that can support physical mice or touchpads (e.g. touch screen laptops, 2-in-1 laptops).

https://user-images.githubusercontent.com/25283178/119231932-c106e180-bb55-11eb-9259-9dbc86d126e8.mp4

## Suggestion
Add a delay (maybe 0.2s) before the animation starts, so that the animation will not play in a typical single click or double click.

e.g.
- Enable Long-press to secondary click
- Set delay to default 1.2s
- if (hold > 0.2s): startCircularWipeAnimation(1)    #total 1.2 seconds of "animation"

JoseExposito:
Add a delay to the dwell click timer animation so the circle is not displayed when the user click quickly.

libinput uses a timeout of 180ms when tapping is enabled. Use that value plus a safety margin so the animation is never displayed when tapping.

Fix #1152

### DIFF
--- a/src/Widgets/DwellClickTimer.vala
+++ b/src/Widgets/DwellClickTimer.vala
@@ -22,6 +22,13 @@ namespace Gala {
 
         private const double START_ANGLE = 3 * Math.PI / 2;
 
+        /**
+         * Delay, in milliseconds, before showing the animation.
+         * libinput uses a timeout of 180ms when tapping is enabled. Use that value plus a safety
+         * margin so the animation is never displayed when tapping.
+         */
+        private const double DELAY_TIMEOUT = 185;
+
         private int scaling_factor = 1;
         private int cursor_size = 24;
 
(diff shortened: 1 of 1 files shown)


danrabbit (changes requested):
Introducing the delay makes sense to me and it seems to work as expected. The only issue I'm noticing is that now the pie animation always starts partially completed. Perhaps we should subtract the delay amount from the timer so that the pie starts animating after the delay?

JoseExposito:
Good point, fixed :+1:

danrabbit (approved):
Great job! Much cleaner

PR #1161, opened 2021-06-04
Workspaces icons/folders make a weird pattern/shape when changing screen resolution

### LINKED ISSUE #1158: Workspaces icons/folders make a weird pattern/shape when changing screen resolution (EOS 6 beta)
## What Happened
I have a dell laptop with a 4k screen. I have changed the screen resolution to 1080p to get rid of the lag that I have been experiencing due to the lack of nvidia on the beta and the intel graphics (UHD 620) not being able to handle the UI animations quiet well on a 4k screen. So, today, I have decided to run "freetube" on one workspace and browse the internet on another. When I went into "overview" by swiping 3 fingers up, this happened:

Screenshot_20210527-144528_Gallery
A reboot did fix it, but I can easily recreate it by changing the resolution between 1080p and 4k and not reboot. Basically, every time I change the resolution, I'd have to reboot, otherwise, this will happen. And the weird shape changes to different shapes each time I recreate it, but I have only captured this one.
On a side note: I don't know if this intentional or not, but I can't take a screenshot while in "overview" mode. Just thought I'd bring this up here, too.

## Expected Behavior
Woekspaces icons/folders should look normal even after changing screen resolution

## Steps to Reproduce
If you have a 4k laptop (that is all I have, so I don't know if the same would happen on other laptops), change the resolution from 4k to 1080p (or vice versa and any other resolution), don't reboot after that, then swipe 3 fingers up to get into "overview" mode.

Dell laptop
Quad-Core Intel® Core™ i7-8550U CPU @ 1.80GHz
Intel® UHD Graphics 620 (Kabylake GT2)/hybrid with Nvidia geforce M130

JoseExposito:
Easier to review commit by commit.

I'm not specially happy with the last commit (IconGroupContainer: Force re-layout) but I wasn't able to find a way to force the layout to be recalculated. Suggestions to improve it are welcome.

Fix #1158

### DIFF
--- a/src/Widgets/IconGroup.vala
+++ b/src/Widgets/IconGroup.vala
@@ -30,6 +30,7 @@ namespace Gala {
         const int PLUS_SIZE = 8;
         const int PLUS_WIDTH = 24;
 
+        const int CLOSE_BUTTON_SIZE = 36;
         const int SHOW_CLOSE_BUTTON_DELAY = 200;
 
         /**
@@ -178,6 +179,7 @@ namespace Gala {
 
             if (show) {
                 show_close_button_timeout = Timeout.add (SHOW_CLOSE_BUTTON_DELAY, () => {
+                    place_close_button ();
                     close_button.visible = true;
                     close_button.opacity = 255;
                     show_close_button_timeout = 0;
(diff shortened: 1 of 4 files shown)


danrabbit (approved):
Works for me! Nice job :)

PR #1163, opened 2021-06-10
Multitasking view: fix scaling factor bugs

JoseExposito:
The multitasking view has a few issues when the scaling factor is changed.

This PR fixes them. Easier to review commit by commit.


cassidyjames (approved):
Nice work!

PR #980, opened 2020-12-03
Add initial support for Mutter 40

tintou:

PR #1181, opened 2021-07-14
[Follow up] Add initial support for Mutter 40

JoseExposito:
Follow up of @tintou's work to migrate Gala to Mutter 40.

He made all the hard work, I just rebased his branch on master, solved a bunch of conflicts and a couple of errors so it compiles with Mutter 40. Tested on Fedora 34.

When running it, `IconGroup.draw` crashes because `icon_container` is `null`:

```csharp
        bool draw (Cairo.Context cr) {
            var scale = InternalUtils.get_ui_scaling_factor ();

            cr.set_operator (Cairo.Operator.CLEAR);
            cr.paint ();
            cr.set_operator (Cairo.Operator.OVER);

            var n_windows = icon_container.get_n_children (); // <- Here
```

I'm not sure if it's supposed to run in a GNOME Shell session.

-----------------

Also, including @davidmhewitt  so he can review my commit "DBusAccelerator: Device ID not available on Mutter 40". I think he was involved in the creation of that interface and we might have to adapt some pieces now that the device ID is not available.


JoseExposito:
Changed the target branch to @tintou's branch, so we can keep the conversation in one place.

tintou:
@JoseExposito I've rebased and updated my branch using fedora:34 container, if you want to look at it

JoseExposito:
Thanks a lot, closing this one in favor of https://github.com/elementary/gala/pull/980

Issue #658, opened 2019-12-12
Make Picture-in-Picture "mask" the window instead of drawing a region


cassidyjames:
Right now PiP works by drawing a rectangular region on the window, which has some issues:

- You only get one shot; if you mess up, you have to close PiP and start over
- The original window sticks around, meaning you have two copies of the content
- It's unclear what you're supposed to do when you hit the shortcut

Instead, we should switch to a corner grabber design that masks/clips the window down to the chosen PiP area, with an explicit "confirm" button to complete the area selection. Here is a sketch I did a while back:

![MVIMG_20191211_220137](https://user-images.githubusercontent.com/611168/70750746-3aa56980-1cec-11ea-9226-ee76bfd0745b.jpg)

And [here is a rough motion mockup on YouTube](https://www.youtube.com/watch?v=SDJYnZlqdn4).

worldofpeace:
Would be fab for https://github.com/elementary/gala/issues/646.

hanaral:
Gtk4 seems to have good builtin media player support, maybe something could be done with that?

PR #1192, opened 2021-08-05
PiP: Mask window

JoseExposito:
## Description

Work in progress PR to fix https://github.com/elementary/gala/issues/658

The current implementation of PiP has some usability issues:

- You only get one shot; if you mess up, you have to close PiP and start over
- The original window sticks around, meaning you have two copies of the content
- It's unclear what you're supposed to do when you hit the shortcut
- Only works on the active window, if you select a different window, you end up with a transparent PiP
- It get's activate event if it is not possible to use it. For example, on the desktop

## Status

- [X] Do not trigger PiP if it is not possible to use it. For example, on the desktop
- [X] Draw the selection area as mask over the target window
- [X] Allow to resize the selection area within the limits of the target window
- [X] Allow to drag & drop the selection area within the limits of the target window
- [X] Exit PiP clicking outside the selection area
- [X] Capture PiP when the enter key is pressed
- [X] Capture PiP with a confirm button in the middle of the selection area
- [X] Change mouse pointer icons
- [X] Hide window areas outside of the mask
- [ ] Don't display the PiP and the original window at the same time
- [X] Remove unused code (the `MIN_SELECTION_SIZE` branch in `pip/Main.vala`)
- [ ] Design & icons (cc @cassidyjames)


JoseExposito:
Hey @cassidyjames, I finished the functional changes but there are some UI elements I hope you can help me with:

- The icon in `data/confirm.svg` is a random icon. I'm sure you'll be able to draw something good... I just don't have talent 
- In `SelectionArea.vala` you can play with the constants `MIN_SELECTION` and `CONFIRM_BUTTON_SIZE`
- Colours are set in `draw_area` in case you want to change them

Feel free to push the changes and I'll squash them into the right commit so we keep the git log useful.

I'll request a code review from desktop devs once UX have checked it :wink:

JoseExposito:
> another option could be to leave the window in those contexts, and automatically exit PiP if it is selected to be focused. That way it's still easy to switch back in the normal flow of things and you don't get into a weird state where it feels like it is two separate windows.

@cassidyjames we have 2 representations of the window. The [Meta.Window](https://gjs-docs.gnome.org/meta8~8_api/meta.window), that is the  actual window and the [Meta.WindowActor](https://gjs-docs.gnome.org/meta8~8_api/meta.windowactor), that is what we draw on screen.

If we hide the actor, the actual window is still on screen and you can interact with it, even though you can not see it.

As far as I know, the only way to really hide the window is minimizing it. I though https://github.com/elementary/gala/commit/65586b07ff3c1339422c9c3722f0f7f0f5d05140 should work, and it does with GTK apps, but not with Chrome. It seems like Chrome is optimized to stop repainting while minimizing, so it is not possible to watch YouTube videos in PiP.

I'm going to revert the patch... As far as I know, this is possible at a lower level, but not using Mutter's API :confused: 

Let's see if @tintou knows a way of doing it :crossed_fingers:

cassidyjames:
@JoseExposito ah, dang. :cry: I suppose another (probably even more hacky?) method could be to put the window on a hidden workspace, or something?

JoseExposito:
@cassidyjames I believe that even if we hide the workspace from the multitasking view, it'd be possible to navigate to it by using shortcuts... It's the "problem" with using an opinionated API like Mutter instead of an unopinionated one like wlroots. It's great, simplifies things a lot and we get a ton of features for free, but sometimes, we are forced to follow their rules. 

This case is especially painful, `gtk_window_hide` does exactly what we want, Mutter has an API to check if a window [is hidden](https://gjs-docs.gnome.org/meta8~8_api/meta.window#method-is_hidden) but, it won't allow us to hide a window :confused: 

`Compositor.hide_window` looked for a moment like the API to use, but look how pointer events still affect the window. Even tooltips are displayed!


https://user-images.githubusercontent.com/1335948/131459116-92a7acc6-9bd2-40e8-b84f-709891179654.mp4


I'll try one last hack: Adding an invisible actor that grabs every event on the top of the window. If that fails, I'll quit Gala development xD

cassidyjames:
@JoseExposito ha, I look forward to your continued experiments. This feels so close UX-wise! And in the end if it doesn't work out, we could put it on the back burner and consider revisiting it if the Mutter API improves.

JoseExposito:
@cassidyjames it's official, now I have to quit 🤣 

The actor approach didn't work. In case someone looks into this, window.shadow, like minimize, stops YouTube videos, so it doesn't work either.

The best we can do is sending the PiP window to the bottom of the window stack so at least looses the focus.

I still think that the mask is easier to use than the current approach. Let me know if you want to merge that part (after adjusting styles, of course). Otherwise I'll close the PR.

cassidyjames:
@JoseExposito so, I'm hesitant about switching to this design if we don't have a way to hide (and stop passing inputs to) the underlying window. If we're unable to do that for now, I think we need to:

1. Keep the existing design from master, or
2. Update this to _not_ actually crop the window when you're positioning the PiP area, as it feels weird that it goes away and then comes back.

Maybe it's just best to punt on this for now…

JoseExposito:
@cassidyjames as you prefer. Option 2 is as easy as dropping the "pip: Hide window areas outside of the mask" commit, in case you want to test it.

If you prefer to stick with the current selection mode, feel free to close this merge request. We can always reopen it in the future.

cassidyjames:
@JoseExposito I think I'd rather punt on this and close it for now—it feels really close but I fear a halfway-there implementation will feel worse than either our current implementation or an ideal one. Happy to get feedback from @elementary/ux if anyone feels differently, though.

JoseExposito:
No problem, let's close it for now then.

danrabbit:
Should this be "draft" instead of "closed"? Typically closed branches end up getting deleted

JoseExposito:
Up to you. As far as I know, there is no API to implement the task "Don't display the PiP and the original window at the same time" in Mutter (main branch) at the moment. If we don't want to implement the mask selection, this PR is going to be hanging on GitHub for a long time.

PR #1193, opened 2021-08-06
editorconfig: Add trim_trailing_whitespace

JoseExposito:
Avoid linter issues by automatically removing trailing white spaces.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,6 +7,7 @@ charset = utf-8
 end_of_line = lf
 indent_size = tab
 indent_style = space
+trim_trailing_whitespace = true
 insert_final_newline = true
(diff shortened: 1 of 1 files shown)


meisenzahl (approved):
**trim_trailing_whitespace**: set to true to remove any whitespace characters preceding newline characters and false to ensure it doesn't.

https://editorconfig.org/

JoseExposito:
Good point, added to the commit description :+1:

jeremypw:
I don't think Code automatically follows this at the moment (amongst several others).  The editorconfig plugin does not seem to do anything with this key.  There is a remove trailing spaces plugin that can be manually enabled.

PR #1218, opened 2021-08-19
Tooltip: use the same text color as in GTK

tintou:
Also simplify it a bit and avoid storing properties over and over

### DIFF
--- a/src/Widgets/Tooltip.vala
+++ b/src/Widgets/Tooltip.vala
@@ -20,7 +20,7 @@
  * Clutter actor to display text in a tooltip-like component.
  */
 public class Gala.Tooltip : Clutter.Actor {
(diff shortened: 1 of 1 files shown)


JoseExposito (changes requested):
The reason why the TextActor is created on draw is:

- Open a window with a long title that can't fit on the window preview
- Change the title to something short
- The tooltip is not able to resize:
image

Back in the day I tried with `allocate_available_size` and `natural_width` and many other workarounds, but they didn't work.

hanaral:
@tintou would it be too much of a stretch to also change the background colour to be the same as gtk tooltips? it seems to be black rather than a mix of gray/brown

danrabbit:
Closing since we merged #1295

PR #1228, opened 2021-08-26
Add initial support for Mutter 41

tintou:


JoseExposito (approved):
Tested with Mutter 41 Beta (latest main branch code) and it compiles.

I think we should wait until the final release is available  and merge as we did with 40.

decathorpe:
> I think we should wait until the final release is available and merge as we did with 40.

Well ... that will make it very hard for me to test this on Fedora. GNOME 41 will only be releases shortly before the Fedora 35 release. Fedora 35 and rawhide already ship GNOME 41 alpha and beta versions, and I'd rather not switch my gala build from building against mutter 3.38 to mutter 41 so late in the development cycle, if you decide to only merge this once GNOME 41 is official.

JoseExposito:
@decathorpe that's a fair point. Also, if Mutter changes its API before the releas we can always quickly solve it.

@tintou I had to modify a bit the meson.build file to build it against 41.alpha. It seems like it was picking 40, most likely because of the alpha suffix. 
If we want to merge it now, we need to fix it. Can you confirm if you have the same issue?

tintou:
@JoseExposito I only tried with 41.beta from fedora rawhide and had no issue detecting it there

JoseExposito:
Cool. As this doesn't affect other Mutter versions, merging so it can be tested. Thanks!!!

decathorpe:
Awesome! I'll set up my nightly gala builds to build against mutter 41, so I should notice pretty quickly if / when mutter adds more incompatible API changes.

decathorpe:
Seems to be working on Fedora 35 without major issues. Now only greeter and wingpanel need this treatment as well 😉

Issue #1232, opened 2021-08-28
Replace task switcher with Catts (Calmer Alt-Tab Task Switcher)


aral:
## Problem

![Screen recording of the default elementary OS task switcher: windows animate in and out while switching between apps and the dock is repurposed as the task switcher view and morphs from the dock icons to the active app icons during the task switching process.](https://raw.githubusercontent.com/small-tech/catts/master/demo/elementary-os-default-task-switcher.gif)

The current task switcher in elementary OS has excessive motion and gratuitous animation and overloads the dock. It also gets stuck sometimes.

  - __Overloads the dock__ (the dock is transformed to include icons of windows and the icons there used to indicate which window you’re switching to). This breaks the physicality of the dock and overloads its meaning. That said, due to the amount of other animation going on, willing myself to concentrate on the dock is the only way I can use it at all.

  - __Has excessive motion__ (animates windows backwards or forwards while dimming them in/out every time you press <kbd>alt</kbd> + <kbd>tab</kbd>). Imagine that happening with maximized or half-screen windows on a 24" monitor. I don’t normally have issues with motion and it makes me feel seasick after a few uses.

  - __Gets stuck.__ Sometimes it will just get stuck in a state where no window is selected. Pressing <kbd>alt</kbd> + <kbd>tab</kbd> again gets you out of it.

<kbd>alt</kbd> + <kbd>tab</kbd> is meant to be a hidden, shortcut gesture for quickly switching between the various windows you have open.

The current implementation fails to address this need.

## Proposal

![Screenshot of Catts in action. The switcher contains three icons: Tasks, AppCenter, and Calculator. AppCenter is selected. The switcher is in dark mode. The wallpaper behind the switcher is an illustration of cats on a roof. One (pink) looks out into the distance, a tiny blue one is about to paw its tail playfully, and a green one watches the scene.](https://raw.githubusercontent.com/small-tech/catts/master/catts-screenshot.jpg)

Replace it with [Catts](https://github.com/small-tech/catts) (Calmer Alt-Tab Task Switcher), an alternative task switcher for elementary OS 6 (Odin) that:

  - __Is calm.__ It does not animate my windows. I don’t want cognitive complexity when I’m fast switching between apps. I want to select the app I want to switch to and switch to it. That’s it.

  - __Uses icons.__ There is very little cognitive load to recognising an icon. There’s a reason we use icons of applications in menus, etc., instead of tiny thumbnails of them. The same principles apply here.

[...]

PR #1234, opened 2021-08-29
Bring "Catts" in-tree

davidmhewitt:
See #1232 

Currently, this is just the bare minimum to get this working in-tree instead of as a plugin. I'll spend the next little while cleaning up the code-style and looking into some of the issues with this. So just opened as a draft for now.


davidmhewitt:
cc @aral

Some of my commits here may be useful for you to port back into Catts while we're waiting for any feedback from the core team.

Currently the only relevant one is https://github.com/elementary/gala/pull/1234/commits/c1d58714a39d22287a23b9e81d5321a80d7755fa . You can drop the dependency on Wnck entirely as the window manager is perfectly capable of giving you title of the window in a non X11 specific way.

aral:
> Some of my commits here may be useful for you to port back into Catts

Thanks, David. These look great. Being new to the platform, I wish I had the namespace declarations when I was trying to port it to Odin; would have made it easier to figure what what was from where :)

Will look into incorporating the ones you recommend back into Catts and fingers crossed Catts won’t need to exist for much longer. Will be a good day when I can hit the archive button on the repository :)

aral:
@davidmhewitt Just released Catts 1.0.1 with your escape fix. Thanks again :)

aral:
@davidmhewitt Regarding your HiDPI support commit, is there a way I can get a reference to `InternalUtils` from the plugin to port these changes back to Catts? (Or is there a public API plugins can use for the same purpose?)

davidmhewitt:
@aral `Utils.get_ui_scaling_factor` should be the one

aral:
> @aral `Utils.get_ui_scaling_factor` should be the one

Thank you :)

cassidyjames:
Couple of quick notes:

- we'll need to figure out the exact styling, but the current pure white with no shadow doesn't work well e.g. with a white window behind it _cc @elementary/ux_


- Alt+Tabbing with a single (or no windows) should emit the GLib.beep thud sound like the current implementation

Otherwise I think this is a great start at a traditional window switcher.

ChildishGiant:
I think that the switcher should appear on whichever screen is active (has the cursor on). Currently I can be using my second monitor and then when I try to alt-tab I think it's not worked, only to remember that it is on the other screen.

aral:
>     * we'll need to figure out the exact styling, but the current pure white with no shadow doesn't work well e.g. with a white window behind it _cc @elementary/ux_

Also, please see my commits linked to from https://github.com/elementary/gala/pull/1234#commitcomment-55657508 as the hidpi commit here changes the appearance of the switcher and those commits both make the icon size consistent with Wingpanel (apart from @1x, where I implemented an optical adjustment to make the icons larger for better legibility/usability – the latter adjustment is in a separate commit to make it easier to cherry pick.)

davidmhewitt:
@aral I've updated the calculations for positioning and sizing based on your changes, but have left the icons sizes and padding as is for now.

64px feels a little small for both HiDPI and LoDPI for me, and we generally don't make any allowances anywhere for things in HiDPI to be a different size to its LoDPI counterpart. Any differences you see there will be purely down to the differing DPI of your two displays, and we don't make allowances for that, since everyone's display is different, we simply scale by the scale factor.

However, it's a shame that 96px feels like the "right" size, as we don't ship 96px icons, so displaying them at this size means they aren't aligned to the pixel grid and will look blurry relative to displaying them at 64px or 128px.

I've added the beep sound when switching isn't possible, will wait from further feedback from @elementary/ux regarding design/sizing.

cassidyjames:
Honestly my only major gripe at this point is the lack of a shadow behind the window. If we can resolve that (or alternatively dim everything behind the switcher, thoughts @danrabbit?), I think I'd be happy to get this merged in. I'm already getting used to it and annoyed with the current switcher when I switch devices/branches. :smile:

aral:
> Honestly my only major gripe at this point is the lack of a shadow behind the window. If we can resolve that (or alternatively dim everything behind the switcher, thoughts @danrabbit?), I think I'd be happy to get this merged in. I'm already getting used to it and annoyed with the current switcher when I switch devices/branches. smile

👍 I’m fine with a shadow if that’s the blocker. (Although we should think of this as more of an overlay than a window – it should really feel like part of the operating system rather than a separate app.) 

👎 Please don’t dim the rest of the interface. When quickly switching between apps, this will create a strobe-style effect and end up being, in effect, a full-screen transition again.

Looking forward to seeing it drop and happy to hear you’re enjoying it Cassidy :)

src/Widgets/WindowSwitcher.vala
```diff
@@ -15,405 +19,276 @@
 //  along with this program.  If not, see <http://www.gnu.org/licenses/>.
 //
```
tintou, line 20:
```suggestion
/*
 * Copyright 2021 Aral Balkan <mail@ar.al>
 * Copyright 2020 Mark Story <mark@mark-story.com>
 * Copyright 2017 Popye <sailor3101@gmail.com>
 * Copyright 2014 Tom Beckmann
 * SPDX-License-Identifier: GPL-3.0-or-later
 */
```
Please use this kind of format for new headers

src/Widgets/RoundedActor.vala
```diff
@@ -0,0 +1,71 @@
...
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//
```
tintou, line 15:
```suggestion
/*
 * Copyright 2017 Popye <sailor3101@gmail.com>
 * SPDX-License-Identifier: GPL-3.0-or-later
 */
```

aral:
> 64px feels a little small for both HiDPI and LoDPI for me, and we generally don't make any allowances anywhere for things in HiDPI to be a different size to its LoDPI counterpart. Any differences you see there will be purely down to the differing DPI of your two displays, and we don't make allowances for that, since everyone's display is different, we simply scale by the scale factor.
>
> However, it's a shame that 96px feels like the "right" size, as we don't ship 96px icons, so displaying them at this size means they aren't aligned to the pixel grid and will look blurry relative to displaying them at 64px or 128px.

In which case, we should probably go with 64px to make the icon sizes consistent with those in the Applications Menu in Wingpanel and to keep to system icon sizes. Thoughts? @cassidyjames @danrabbit

(The calculations in the commit I linked to should hold for that, we’d just need to not include the @1x special case.)

aral:
Blocker: needs accessibility

(Not sure if there’s a label for this that someone with admin access can set.)

I just tried it out with the screen reader and Catts is entirely invisible to the accessibility system. I’ve also just spent about half an hour trying to find a getting started guide or some information on how to implement accessibility in GTK+3 (with code examples) to no avail.

At a minimum:

- The task switcher should be announced: “Task switcher”
- The title of the selected app should be announced “X … selected”

aral:
@davidmhewitt I’m on holiday so apologies for the brief reply. I just spend a little time looking into this and it appears at ATK is also implemented at the Clutter-level.

I’ve managed to hack my way through to implementing basic accessibility for the component. Please see the last three commits at https://github.com/small-tech/catts/commits/accessibility

This leads to the switcher announcing itself as the task switcher on launch as well as announcing the number of items it holds. It also announces the title of the selected application as the selection changes.

My commits don’t currently use localisable strings for this. This will have to be implemented.

Would hugely appreciate it if you could port these into Gala and if we could launch with an accessible task switcher :)

davidmhewitt:
> I’ve managed to hack my way through to implementing basic accessibility for the component using whatever scant documentation and clues I could find online. Please see the last three four commits at https://github.com/small-tech/catts/commits/accessibility

Well, every day is a school day! Thanks for figuring that out, I'll try and take a look this weekend and test it out and port it across.

aral:
@davidmhewitt Thanks, David. Really appreciate it.

cassidyjames:
@aral

>:+1: I’m fine with a shadow if that’s the blocker. (Although we should think of this as more of an overlay than a window – it should really feel like part of the operating system rather than a separate app.)

We currently don't have that sort of visual distinction anywhere in elementary OS; the system shut down dialog looks like a regular dialog, as do agent/portal dialogs, the Shortcut Overlay, etc. If we were to make a distinction (which may be worth pursuing!), we'd want to do it across all similar elements at the same time. But since those all have standard window/dialog styling at the moment, I'd like to match it.

>:-1: Please don’t dim the rest of the interface. When quickly switching between apps, this will create a strobe-style effect and end up being, in effect, a full-screen transition again.

Ah yeah good call! :smile:

aral:
Hey all, just got back from three weeks away, seeing the parents and was wondering how things are going on this. Is there anything you need from me to move this forward? Hope you’re all well.

<3

cassidyjames:
@aral I have been using and enjoying it. I think we just still need to figure out a shadow, mark it as ready for review, and get it reviewed. :smile:

davidmhewitt:
And I had a poke at the accessibility stuff as promised but found it wasn't particularly consistent or reliable, plus it didn't behave like any other widget with the screen reader.

It seems like there should be a better way of doing it like adding the accessibility title to each "icon" and getting the keyboard focus to jump to each icon as you tab through, but this didn't seem to be working well either.

I guess I can find the "least bad" accessibility support and push that in here. I can also have a look at the shadow, but things got really busy with my day job recently.

danrabbit (approved):
It seems like possibly the same issue we're running into with menu shadows, so I don't think we should hold this back over it. I'd like to get this merged since it's a long-standing issue and unblocks Wayland and we can iterate more in future branches if need be. Better to get it in at the beginning of the month where there's lots of time to test

PR #1237, opened 2021-08-31
Hot corners: Migrate to barriers

### LINKED ISSUE #719: Allow for "cursor catching" in hot corners similar to how GNOME Shell does it
## Prerequisites
- [X] I have searched open and closed issues for duplicates (although I’m not 100% certain this is the right repo for it).

## Feature
**Is your feature request related to a problem? Please describe.**

On a multi-monitor setup, hot corners at the seam between two displays are hard to trigger because the mouse just slips right through them. GNOME shell fixes this for the activities overview by allowing the cursor to be briefly "caught" in the hot corner, allowing for a reliable way to trigger it with a quick flick of the mouse.

**Describe the solution you'd like**

I think elementary would benefit from a similar, perhaps optional, feature, especially after reading elsewhere that hot corners should be constrained to the main display (which makes complete sense in my opinion, because stretching them across monitors would just make them practically useless on large multi-monitor setups).

**Existing work**

As I’ve mentioned before, this is being done by GNOME Shell for the Activities Overview hot corner, unless that hot corner is disabled like on recent Ubuntu releases.

### LINKED ISSUE #1054: Multitasking View Hotcorner: Touchpad scrolling doesn't work on the selected corner location.
## What Happened

Choosing Multitasking View as a Hotcorner makes the touchpad scrolingl function not work on the selected corner location. Moving the cursor makes scrolling work again.

## Expected Behavior

Users shouldn't have to move the mouse cursor from the hotcorner in order to use the scrolling function.

## Steps to Reproduce

1. Go to System Settings/Desktop/Hotcorners and select "Multitasking View" as the bottom left Hotcorner.
2. Move the mouse cursor to the bottom left of the screen to open Multitasking View.
3. Try to use use the touchpad scrolling to change workspaces.

## Platform Information

elementary OS 5.1.7 Hera

### LINKED ISSUE #1078: Hotcorners: Pressure corner to activate again without having to move the mouse.
## Problem

Sometimes users just want to take a quick glance at the Multitasking View or at the Window Overview and in order to exit moving the mouse is needed.

Another case is #1054 where an user have to move the mouse to hit the corner again after using the mouse scroll.

## Proposal

It would be nice if the corner could be pressured to activate the corner again without having to move the mouse.

### LINKED ISSUE #1100: Add pressure reveal to hot corners
## Problem
Hot corners are activated whenever the mouse pointer is hovered over. This increases the chances of accidental activations of hot corners. 

## Proposal
Add pressure reveal (like in the dock) and maybe even enable it by default.

JoseExposito:
### Description

Our current implementation of hot corners places 4 Clutter Actors on screen to check mouse events.

While functional, it has some issues that can not be solved without migrating to barriers:

- Fix #719
- Fix #1054
- Fix #1078
- Fix #1100

### Code review

Changes are split as much as possible so the implementation is easy to follow.


cassidyjames (commented):
UX- and QA-wise I am happy with this across single and multiple displays! I think I’d still like a code review, but nice improvements @JoseExposito.

src/HotCorners/HotCornerManager.vala
```diff
@@ -0,0 +1,103 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+public class Gala.HotCornerManager : Object {
+    public signal void on_configured ();
+
+    public WindowManager wm { get; construct; }
+    public GLib.Settings behavior_settings { get; construct; }
+
+    private GLib.Array<HotCorner> hot_corners;
```
tintou, line 25:
```suggestion
    private GLib.GenericArray<HotCorner> hot_corners;
```
It's better to hold GObjects and you can use it as a loop too

JoseExposito:
Thanks a lot for the code review @tintou ! Fix applied in the right patch.

tintou (approved):

tintou:
Thanks @JoseExposito that looks great!

PR #1241, opened 2021-08-31
App: Implement Applications Management

tintou:
This allows us to completely replace libbamf in some specific cases (like system shell elements).

This PR is only opened for showing the intention, it's far from even starting to work.

### DIFF
--- a/lib/App.vala
+++ b/lib/App.vala
@@ -1,0 +1,179 @@
+/*
+ * Copyright 2021 elementary, Inc. <https://elementary.io>
+ * Copyright 2021 Corentin Noël <tintou@noel.tf>
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+public enum Gala.AppState {
+    STOPPED,
+    STARTING,
+    RUNNING
+}
+
+public class Gala.App : GLib.Object {
+    public string id {
+        get {
+            if (app_info != null) {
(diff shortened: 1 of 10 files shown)


davidmhewitt:
As a quick note here, I wonder if there are any benefits of making the DBus API compatible with `org.gnome.shell.Introspect`: https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/386d25e6f8ce11549526ea3776eb34138fcb3774/data/dbus-interfaces/org.gnome.Shell.Introspect.xml

I believe this is used for automated UI testing and some accessibility stuff, so perhaps useful?

tintou:
> As a quick note here, I wonder if there are any benefits of making the DBus API compatible with `org.gnome.shell.Introspect`: https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/386d25e6f8ce11549526ea3776eb34138fcb3774/data/dbus-interfaces/org.gnome.Shell.Introspect.xml
> 
> I believe this is used for automated UI testing and some accessibility stuff, so perhaps useful?

This is a private API that filters its users so I don't think that we should provide it

tintou:
Removing the draft status as I'd like it to land with experimental status

danirabbit:
Since this is only for use with other shell components I don’t see the harm in merging :)

danirabbit (approved):

PR #1245, opened 2021-09-06
Do not use color properties directly

### LINKED ISSUE #1243: Clutter.Color broken since Mutter 41 compatibility patch
Since #1228, `Clutter.Color.from_string` does not work _for all versions of Mutter_. So, even if we compile Gala against our current version of Mutter on Odin.

This breaks things like the ScreenShield as it can't set a black color for itself, and probably anywhere else where we use `Clutter.Color`.

I think removing `color_from_* skip` from `Clutter-9.metadata` and letting `vapigen` generate those methods (and of course reverting the changes to Gala) should work, but I don't have enough of an environment set up to test these changes on all versions of Mutter.

However, this is a release blocker, **do not release another version of Gala until this issue is resolved**

tintou:
They are being copied when retrieving as they are simple structs.
Closes #1243

### DIFF
--- a/src/Background/Background.vala
+++ b/src/Background/Background.vala
@@ -94,19 +94,17 @@ namespace Gala {
             var settings = background_source.settings;
 
             color_string = settings.get_string ("primary-color");
-            var color = Clutter.Color ();
-            color.from_string (color_string);
-
-            color_string = settings.get_string ("secondary-color");
-            var second_color = Clutter.Color ();
-            second_color.from_string (color_string);
+            var color = Clutter.Color.from_string (color_string);
 
             var shading_type = settings.get_enum ("color-shading-type");
 
-            if (shading_type == GDesktop.BackgroundShading.SOLID)
+            if (shading_type == GDesktop.BackgroundShading.SOLID) {
                 background.set_color (color);
-            else
+            } else {
+                color_string = settings.get_string ("secondary-color");
+                var second_color = Clutter.Color.from_string (color_string);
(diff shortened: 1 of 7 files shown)


JoseExposito (dismissed):
Is the same change required in `Background.load_pattern ()`?

By the way, tested with latest Mutter :+1:

davidmhewitt:
To be honest, this binding felt a lot better when `Clutter.Color.from_string` just returned a `Clutter.Color`, even if it did hide the boolean (that we don't use anyway). It felt like a "constructor", like all the other `from_x` methods that exist across GLib libraries.

What I don't understand is how these bindings are being generated because of this:
https://github.com/elementary/gala/blob/837fb56e70dab88aac56521103c9d58d671d883b/vapi/Clutter-9.metadata#L119

If those symbols are skipped and I can't find anywhere they're manually, re-added, why are they in the bindings at all?

tintou:
@davidmhewitt I've changed the .metadata and -custom.vapi to reflect the API you wanted here, makes things easier

JoseExposito (approved):
I was experimenting crashes when changing the wallpaper and I though it was because of this change, but no, this code is fine.

After bisecting it looks like the problem was introduced in https://github.com/elementary/gala/commit/4aec79ae0d4e2c53a397f7fac3d8e77243d38692

Merging this branch, as it works as expected both with elementary's Mutter and Fedora's Mutter, and I'll try to fix the crash.

PR #1246, opened 2021-09-07
WindowSwitcher: Use accent and palette colors

cassidyjames:
- Use elementary Palette colors by default
- Remove transparency (at least for now)
- Use the user's selected accent color for the icon highlight
- Simplify creation of variables and only override for dark preference

There may be a better way to get and set the accent color, and if possible, I'd like to get the palette colors from the system stylesheet as well. I'd also still like to get a shadow on this, but I think this is far more legible than the current branch.

## Before

Screenshot from 2021-09-07 11 38 35

## After

Screenshot from 2021-09-07 12 11 54

Screenshot from 2021-09-07 12 13 15

### DIFF
--- a/src/Widgets/WindowSwitcher.vala
+++ b/src/Widgets/WindowSwitcher.vala
@@ -83,20 +83,21 @@ namespace Gala {
             }
 
             // Set the colours based on the person’s light/dark scheme preference.
-            var wrapper_background_color = "red";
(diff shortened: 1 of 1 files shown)


JoseExposito (approved):
Code looks good, I like a lot the new background color, it makes easier to read the text :+1:

PR #1247, opened 2021-09-08
WindowSwitcher: draw from Gtk.StyleContext

danrabbit:
Draw the background of the switcher from a Gtk.StyleContext instead of hardcoding background/foreground/radius information

While here:
* Change the icon size to supported size
* Make sure we're drawing at integer values

### DIFF
--- a/src/Widgets/WindowSwitcher.vala
+++ b/src/Widgets/WindowSwitcher.vala
@@ -22,7 +22,7 @@
 namespace Gala {
     public class WindowSwitcher : Clutter.Actor {
         public const int ICON_SIZE = 64;
-        public const int WRAPPER_BORDER_RADIUS = 12;
+        public const int WRAPPER_BORDER_RADIUS = 3;
         public const int WRAPPER_PADDING = 12;
         public const string CAPTION_FONT_NAME = "Inter";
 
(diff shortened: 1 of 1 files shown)


src/Widgets/WindowSwitcher.vala
```diff
@@ -95,27 +133,18 @@ namespace Gala {
...
             indicator.margin_left = indicator.margin_top =
                 indicator.margin_right = indicator.margin_bottom = 0;
             indicator.set_pivot_point (0.5f, 0.5f);
 
+            var caption_color = "#2e2e31";
+
+            if (granite_settings.prefers_color_scheme == Granite.Settings.ColorScheme.DARK) {
+                caption_color = "#fafafa";
+            }
+
```
cassidyjames, line 147: @danrabbit was there any way to get this from the style context, too?

cassidyjames (approved):
One comment, but regardless this is an improvement. Thanks!

src/Widgets/WindowSwitcher.vala
danrabbit: I keep getting a segfault trying to grab it, so I figured it was better to leave for another branch

PR #1250, opened 2021-09-10
Fix crash changing wallpaper

JoseExposito:
The lambda function wasn't capturing the value of "i". Instead, "OPTIONS.length" was always used.

### DIFF
--- a/src/Background/BackgroundSource.vala
+++ b/src/Background/BackgroundSource.vala
@@ -52,14 +52,20 @@ namespace Gala {
             // new actors all the time, which lead to some problems in other areas of the code
             for (int i = 0; i < OPTIONS.length; i++) {
                 hash_cache[i] = settings.get_value (OPTIONS[i]).hash ();
-                settings.changed[OPTIONS[i]].connect (() => {
-                    uint new_hash = settings.get_value (OPTIONS[i]).hash ();
-                    if (new_hash != hash_cache[i]) {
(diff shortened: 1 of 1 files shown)


tintou:
You actually get the key as first argument:
```
settings.changed[OPTIONS[i]].connect ((key) => {
// use `key` instead of `OPTIONS[i]` inside the closure
}
```

JoseExposito:
@tintou but I'd need to search for the index of `OPTIONS[key]` to use it in `hash_cache[index]` and `hash_cache[key] = new_hash;`.
I didn't find a method to easily do it, so I leave it as simple as possible. Do you know if there is a method similar to [Array.indexOf](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf)?

Said that, I'm renaming "key" to "index". It's a more accurate name.

tintou:
Maybe replace that whole block with:
```vala
            settings.changed.connect ((key) => {
                for (int i = 0; i < OPTIONS.length; i++) {
                    if (key == OPTIONS[i]) {
                        uint new_hash = settings.get_value (key).hash ();
                        hash_cache[i] = new_hash;
                        changed ();
                    }
                }
            }
```

JoseExposito:
Done, it increases a bit the complexity, but it's easier to understand.

tintou (approved):

PR #1255, opened 2021-09-18
autostarts: set hidden under systemd

davidmhewitt:
This is a prerequisite for https://github.com/elementary/session-settings/issues/17 but it has no effect on the current way we're doing session management.

See: https://manpages.debian.org/experimental/gnome-session-bin/gnome-session.1.en.html

This just prevents `gnome-session` from autostarting gala if systemd is launching the session instead, which it isn't, currently.

All of the gsd autostarts already have this property and they still function.


tomaswarynyca (approved):

JoseExposito (approved):
Looks great. It shouldn't affect us at this point, correct? I think we are good to merge it.

PR #1256, opened 2021-09-19
Add systemd units

davidmhewitt:
Adds support for launching gala via the new systemd support in `gnome-session`. This has no effect on the current way of doing things as we're still using the `--builtin` flag on our session in session settings, but it prepares us for the future.

I've made this an option in meson so that distributions without systemd can use the old XDG autostarts way of launching gala in the future if they wish.

This can be tested in conjunction with https://github.com/elementary/gala/pull/1255 and https://github.com/elementary/session-settings/pull/55. If you remove the `--builtin` flag from `/usr/share/xsessions/pantheon.desktop`, `gnome-session` will launch Pantheon using the systemd units instead of the XDG autostarts.

### DIFF
--- a/data/gala-x11.service.in
+++ b/data/gala-x11.service.in
@@ -1,0 +1,28 @@
+[Unit]
+Description=Gala on X11
+# On X11, try to show the fail whale screen
+OnFailure=gnome-session-failed.target
+OnFailureJobMode=replace
+CollectMode=inactive-or-failed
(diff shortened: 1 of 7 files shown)


tomaswarynyca (approved):

JoseExposito (approved):
Once again, it looks good to me, but my experience with systemd is limited (and traumatic :P)

I'd appreciate another review.

PR #1266, opened 2021-10-05
Multitasking view: Don't use smooth scroll events to handle mouse wheel

JoseExposito:
Instead of relaying on a magic value to know detect mouse wheel "clicks" use the discrete events sent by Mutter.

This will avoid breaking the multitasking view scroll once `xf86-input-libinput` 1.2.0 lands in Ubuntu with high-resolution scroll wheel support.

### DIFF
--- a/src/Widgets/MultitaskingView.vala
+++ b/src/Widgets/MultitaskingView.vala
@@ -178,32 +178,34 @@ namespace Gala {
                 return true;
             }
 
-            if (scroll_event.direction != ScrollDirection.SMOOTH) {
+            if (scroll_event.direction == ScrollDirection.SMOOTH ||
+                scroll_event.scroll_source == ScrollSource.FINGER ||
+                scroll_event.get_source_device ().get_device_type () == Clutter.InputDeviceType.TOUCHPAD_DEVICE) {
                 return false;
             }
 
-            double dx, dy;
-#if VALA_0_32
-            scroll_event.get_scroll_delta (out dx, out dy);
-#else
-            var event = (Event*)(&scroll_event);
-            event->get_scroll_delta (out dx, out dy);
-#endif
-
-            // concept from maya to detect mouse wheel
-            if (Math.fabs (dy) == 1.0) {
-                var direction = dy > 0 ? MotionDirection.RIGHT : MotionDirection.LEFT;
-
-                unowned Meta.WorkspaceManager manager = display.get_workspace_manager ();
-                var active_workspace = manager.get_active_workspace ();
-                var new_workspace = active_workspace.get_neighbor (direction);
-
-                if (active_workspace != new_workspace) {
-                    new_workspace.activate (display.get_current_time ());
-                }
+            Meta.MotionDirection direction;
+            switch (scroll_event.direction) {
+                case ScrollDirection.UP:
+                case ScrollDirection.LEFT:
+                    direction = MotionDirection.LEFT;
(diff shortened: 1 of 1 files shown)


src/Widgets/MultitaskingView.vala
```diff
@@ -178,32 +178,34 @@ namespace Gala {
...
 
-                if (active_workspace != new_workspace) {
-                    new_workspace.activate (display.get_current_time ());
-                }
+            if (active_workspace != new_workspace) {
+                new_workspace.activate (display.get_current_time ());
             }
 
-            return false;
+            return true;
```
cassidyjames, line 208: As long as we're changing this line, can we use the Gdk macro so the intent is more clear?

```suggestion
            return EVENT_STOP;
```

davidmhewitt: Eh, since this isn't a GDK event handler, I'd be tempted not to. Yes, it's almost an impossibility that will ever be anything other than `true` but since the two things aren't related, it doesn't seem right to me.

We also need to be using less GTK/GDK in Gala, because it causes issues with Wayland. Again, yes, this is just a constant so it wouldn't be part of the problem. But if we can remove GTK altogether, we can drop the dependency.

Exalm: There's `CLUTTER_EVENT_STOP`, though doesn't seem like it's in the vapi.

JoseExposito: @Exalm is right, GNOME Shell uses `Clutter.EVENT_PROPAGATE/EVENT_STOP`. I tried to do the same, but as he mentioned, they are not in the vapi.

I guess it's something we can address in a different PR.

src/Widgets/MultitaskingView.vala
```diff
@@ -178,32 +178,34 @@ namespace Gala {
                 return true;
             }
 
-            if (scroll_event.direction != ScrollDirection.SMOOTH) {
+            if (scroll_event.direction == ScrollDirection.SMOOTH ||
```
davidmhewitt, line 181: Was this right before?

This has changed from `!= SMOOTH` to `== SMOOTH`.

So if it is a smooth scroll event, we return here and let `GestureTracker` handle it, right? But does that mean this was wrong before? Or is there something I don't understand.

JoseExposito: The key here is that, no matter what kind of device produced the scroll, Mutter sends both a smooth and a non-smooth scroll events.

For example, 1 low-resolution ("normal" mouse) mouse wheel click will generate 1 scroll event with direction UP/DOWN *and* 1 scroll event with direction SMOOTH.

A touchpad or a high-resolution scroll mouse, scrolling a bit will generate X number of SMOOTH scroll events and, after a certain threshold is reached, 1 scroll event with direction != SMOOTH.

This design is very clever, because it allows to listen to discrete scroll events without worrying about the underlying hardware.

TL;DR

Answering your question, yes, before this change, this code worked. When the SMOOTH scroll events weren't handled by the `GestureTracker` (ie, non-touchpad events), this code was triggered and, by using a magic number, we changed the workspace.

However, once `xf86-input-libinput` 1.2.0 lands in elementary OS, we will receive SMOOTH scroll events from mice, and this won't work. Also, it is a good idea to avoid magic numbers and rely on Mutter's discrete scroll emulation.

Issue #927, opened 2020-10-06
Workspace bump animation on non-primary display


danirabbit:
Even though we only have workspaces on the primary display, we get the workspace end bump animation on all displays

PR #1275, opened 2021-10-22
Multitasking View: Show nudge animation only on the primary display

JoseExposito:
Requires https://github.com/elementary/gala/pull/1270

----------------------

When the multitasking view is opened and a the keyboard shortcuts are used to switch between workspaces, avoid showing the nudge/bump animation on the non-primary display.

Partially address #927

### DIFF
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -472,7 +472,11 @@ namespace Gala {
                 neighbor.activate (display.get_current_time ());
             } else {
                 // if we didnt switch, show a nudge-over animation if one is not already in progress
-                play_nudge_animation (direction);
+                if (workspace_view.is_opened () && workspace_view is MultitaskingView) {
+                    ((MultitaskingView) workspace_view).play_nudge_animation (direction);
+                } else {
+                    play_nudge_animation (direction);
+                }
             }
         }
 


src/WindowManager.vala
```diff
@@ -472,7 +472,11 @@ namespace Gala {
                 neighbor.activate (display.get_current_time ());
             } else {
                 // if we didnt switch, show a nudge-over animation if one is not already in progress
-                play_nudge_animation (direction);
+                if (workspace_view.is_opened () && workspace_view is MultitaskingView) {
+                    (workspace_view as MultitaskingView).play_nudge_animation (direction);
```
jeremypw, line 476: Causes warning 
```
../src/WindowManager.vala:476.21-476.77: warning: Access to possible `null'. Perform a check or use an unsafe cast.
```

The elementary recommended way of casting to avoid the warning is `((MultitaskingView) workspace_view).play_nudge_animation (direction)`.

jeremypw (changes requested):
Just a small comment about code style.

src/WindowManager.vala
JoseExposito: Fixed, thanks!!

jeremypw (approved):

PR #1280, opened 2021-10-29
Handle signal to close Gala gracefully on Mutter 3.41

### LINKED ISSUE #1258: Add signal handlers for latest mutter
We need a commit similar to what has been done here as it is not done by mutter directly anymore https://gitlab.gnome.org/GNOME/mutter/-/commit/fca9010

tintou:
Closes #1258

### DIFF
--- a/src/Main.vala
+++ b/src/Main.vala
@@ -42,7 +42,27 @@ namespace Gala {
             return Posix.EXIT_FAILURE;
         }
 
+        /* Intercept signals */
         ctx.set_plugin_gtype (typeof (WindowManagerGala));
+        Posix.sigset_t empty_mask;
+        Posix.sigemptyset (out empty_mask);
+        Posix.sigaction_t act = {};
+        act.sa_handler = Posix.SIG_IGN;
+        act.sa_mask = empty_mask;
+        act.sa_flags = 0;
+
+        if (Posix.sigaction (Posix.SIGPIPE, act, null) < 0) {
+            warning ("Failed to register SIGPIPE handler: %s", GLib.strerror (GLib.errno));
+        }
+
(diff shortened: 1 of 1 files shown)


src/Main.vala
```diff
@@ -42,7 +42,27 @@ namespace Gala {
...
+        Posix.sigaction_t act = {};
+        act.sa_handler = Posix.SIG_IGN;
+        act.sa_mask = empty_mask;
+        act.sa_flags = 0;
+
+        if (Posix.sigaction (Posix.SIGPIPE, act, null) < 0) {
+            warning ("Failed to register SIGPIPE handler: %s", GLib.strerror (GLib.errno));
+        }
+
+        if (Posix.sigaction (Posix.SIGXFSZ, act, null) < 0) {
```
JoseExposito, line 58: I see both in [Mutter](https://gitlab.gnome.org/GNOME/mutter/-/blob/main/src/core/mutter.c#L83) and [GNOME Shell](https://github.com/GNOME/gnome-shell/blob/f579e9dd8e033666e97db042a89b360b5e27e3bd/src/main.c#L465) that SIGXFSZ is used inside an #ifdef.

Should we do the same?

JoseExposito (changes requested):

src/Main.vala
tintou: It is for non-linux users, we could if the non-linux users actually care for Gala so I'll be waiting for them to actually add it if it is the case

JoseExposito (approved):

src/Main.vala
JoseExposito: Fair enough, merging

PR #1285, opened 2021-11-02
WindowManager: dim parents of modals

### LINKED ISSUE #576: Dim parent windows
There is a vestigial gsetting for this that appears to not do anything. But it would be a nice touch to dim parent windows of modal dialogs. This has become a pretty common design pattern across platforms.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/76324225-dim-parent-windows?utm_campaign=plugin&utm_content=tracker%2F62872462&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F62872462&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #576


cassidyjames (changes requested):
This seems like a good start, but I am seeing some strange behavior when moving windows: if you open a modal window and then try to move the parent, the child ends up shooting off the right side of the display.

danrabbit:
@cassidyjames that also happens in master

cassidyjames:
@danrabbit lmao oh that's nasty. Alright, confirmed and filed as #1286

src/WindowManager.vala
```diff
@@ -721,17 +721,6 @@ namespace Gala {
                 out x, out y);
         }
 
-        public void dim_window (Meta.Window window, bool dim) {
```
JoseExposito, line 724: I think it'd be a good idea to keep this method around and move the logic to dim/"un-dim" an actor here.

JoseExposito (changes requested):
Just a little comment about the code.

The effect looks nice, but I feel like it's too sudden. Maybe we could animate it for a short period of time.

GNOME Shell has a good (JavaScript) example to follow.

danrabbit:
@JoseExposito pushed the method, but I don't think I'm familiar enough with clutter to be able to implement the animation. Would love help :sweat_smile:

JoseExposito (approved):
Code looks much nicer.

@danrabbit here GNOME's implementation:
https://github.com/GNOME/gnome-shell/blob/483d1b3177a8b35599b5825b4ac51a105b360c08/js/ui/windowManager.js#L125

On the surface, it looks simple, but things will get complicated to migrate their `ease_property` method:
https://github.com/GNOME/gnome-shell/blob/cc5cc0d653ccd307475921bf069773d44b34aa5e/js/ui/environment.js#L223

In my opinion we can merge this as it is and add the animation in the future. Feel free to press the merge button :)

PR #1288, opened 2021-11-04
Enable "Alt" + "~" to switch between windows of the same application #888

### LINKED ISSUE #536: Enable "Alt" + "~" to switch between windows of the same application

andreasfelix:
Supersedes #888
Fixes #536

Switching between different instances of the same applications via <kbd>Alt</kbd><kbd>Above_tab</kbd> is already implemented in mutter, but is - at least to my understanding - disabled by the following lines:

https://github.com/elementary/gala/blob/517748bcac9484df6a1f03c84f77c9a066ed8793/src/WindowManager.vala#L296-L297

Is there a reason behind this? 

This PR just enables the switch group shortcuts from mutter by removing these empty custom handlers.

### DIFF
--- a/data/gala.appdata.xml.in
+++ b/data/gala.appdata.xml.in
@@ -16,6 +16,7 @@
         <ul>
           <li>Support for high-resolution scroll events</li>
           <li>Redesigned Alt + Tab switcher</li>
+          <li>Enable switch group shortcuts from mutter (Alt + AboveTab)</li>
           <li>Updated translations</li>
         </ul>
       </description>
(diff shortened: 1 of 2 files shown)


jeremypw (changes requested):
For me, this PR stops the usual app switcher`Alt + Tab` from working and does not switch between windows of the same app when either `Alt + AboveTab` or `Alt + ~` is pressed..  This is after installing and running `gala --replace`.

For me, pressing `~` involves holding the `Shift` key down and there are known problems with `Alt + Shift` conflicting with the keyboard layout switching shortcut. 

Also `Alt + ~` is capture by the terminal ad prints a list of groups.

JoseExposito (changes requested):
In my case (Spanish Spain ISO keyboard) Alt+Tab keeps working, but Alt+Above Tab doesn't trigger any action.

PR #1291, opened 2021-11-05
IconGroupContainer: add null check to get_next_sibling

danrabbit:
Gets read of a terminal error I was getting


JoseExposito (approved):
Thanks

PR #1292, opened 2021-11-05
MultitaskingView: don't set custom easing mode

danrabbit:
This fixes a terminal error about setting a non-existant custom easing mode. Presumably the original author intended to have "no" easing mode, which should mean `LINEAR`

Also removed the constant `ANIMATION_MODE` while we're here since that just obfuscates what the actual easing we're setting here is


JoseExposito (approved):
Yes, I think that was the original's author intension 🙄

PR #1295, opened 2021-11-06
Tooltip: get text color from style context

danrabbit:
Just the color parts from #1218


src/Widgets/Tooltip.vala
```diff
@@ -20,7 +20,7 @@
  * Clutter actor to display text in a tooltip-like component.
  */
 public class Gala.Tooltip : Clutter.Actor {
-    private static Clutter.Color text_color;
+    private static Gdk.RGBA text_color;
```
JoseExposito, line 23: Instead of storing the `Gdk.RGBA` colour, we could store the `Clutter.Color` and avoid doing the multiplications in the `draw` method.

JoseExposito (changes requested):
Just one little change, other than that looking great.

src/Widgets/Tooltip.vala
danrabbit: Good call, cached the math!

JoseExposito (approved):

PR #1296, opened 2021-11-06
WindowClone: Include spread in bounding box

danrabbit:
this seems to solve an issue where the shadows of SSD windows (like Gtk.FileChooser) were being clipped

### DIFF
--- a/src/Widgets/WindowClone.vala
+++ b/src/Widgets/WindowClone.vala
@@ -28,7 +28,7 @@ namespace Gala {
 
         public override ActorBox get_bounding_box () {
             var scale_factor = InternalUtils.get_ui_scaling_factor ();
-            var size = shadow_size * scale_factor;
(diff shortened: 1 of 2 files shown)


src/Widgets/WindowClone.vala
```diff
@@ -28,7 +28,7 @@ namespace Gala {
 
         public override ActorBox get_bounding_box () {
             var scale_factor = InternalUtils.get_ui_scaling_factor ();
-            var size = shadow_size * scale_factor;
+            var size = shadow_size + shadow_spread * scale_factor;
```
cassidyjames, line 31: Haven't tested this, but wouldn't this make more sense as

```suggestion
            var size = (shadow_size + shadow_spread) * scale_factor;
```

danrabbit: Yeah you're probably right!

tintou (approved):

PR #1297, opened 2021-11-06
WindowManager: dialogs fall down

danrabbit:
dialogs fall down on top of windows instead of shooting out of them

https://user-images.githubusercontent.com/7277719/140624875-111a582f-d703-4383-b6cf-1a5a2abe3dc6.mp4


JoseExposito (approved):
Code and the new animation look good :)

PR #1298, opened 2021-11-06
WindowManager: explicitly dim and undim

danrabbit:
I've got into a situation a couple times where no dim effect is applied and then when the window is closed the dim is added. Oops

### DIFF
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -727,7 +727,7 @@ namespace Gala {
                 var ancestor = window.find_root_ancestor ();
                 if (ancestor != null && ancestor != window) {
                     var win = (Meta.WindowActor) ancestor.get_compositor_private ();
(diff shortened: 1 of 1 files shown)


davidmhewitt (approved):

src/WindowManager.vala
```diff
@@ -722,18 +722,19 @@ namespace Gala {
...
-                        win.clear_effects ();
-                    } else {
+                    // Can't rely on win.has_effects since other effects could be applied
+                    if (dim) {
                         var dark_effect = new Clutter.BrightnessContrastEffect ();
                         dark_effect.set_brightness (-0.4f);
 
                         win.add_effect (dark_effect);
+                    } else {
+                        win.clear_effects ();
```
vjr, line 737: A comment mentioned above says other effects could be applied so wouldn't this clear not just the intended dim effect but also any other effects not intended to be removed?

PR #1304, opened 2021-11-09
WindowManager: don't type check when undim

danrabbit:
I'm still having issues where sometimes windows don't undim, but it looks like maybe this really fixes it this time for real, final final.psd

### DIFF
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -734,7 +734,7 @@ namespace Gala {
 
                         win.add_effect_with_name ("dim-parent", dark_effect);
                     }
-                } else if (win.get_effect ("dim-parent") != null){
+                } else if (win.get_effect ("dim-parent") != null) {
                     win.remove_effect_by_name ("dim-parent");
                 }
             }
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -723,9 +723,9 @@ namespace Gala {
         }
 
         private void dim_parent_window (Meta.Window window, bool dim) {
-            var ancestor = window.find_root_ancestor ();
+            unowned var ancestor = window.find_root_ancestor ();
             if (ancestor != null && ancestor != window) {
-                var win = (Meta.WindowActor) ancestor.get_compositor_private ();
+                unowned var win = (Meta.WindowActor) ancestor.get_compositor_private ();
                 // Can't rely on win.has_effects since other effects could be applied
                 if (dim) {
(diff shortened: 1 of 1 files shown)


src/WindowManager.vala
```diff
@@ -723,19 +723,19 @@ namespace Gala {
         }
 
         private void dim_parent_window (Meta.Window window, bool dim) {
-            if (window.window_type == Meta.WindowType.MODAL_DIALOG) {
-                var ancestor = window.find_root_ancestor ();
-                if (ancestor != null && ancestor != window) {
-                    var win = (Meta.WindowActor) ancestor.get_compositor_private ();
-                    // Can't rely on win.has_effects since other effects could be applied
-                    if (dim) {
+            var ancestor = window.find_root_ancestor ();
```
tintou, line 726:
```suggestion
            unowned var ancestor = window.find_root_ancestor ();
```

src/WindowManager.vala
```diff
@@ -723,19 +723,19 @@ namespace Gala {
...
+            if (ancestor != null && ancestor != window) {
+                var win = (Meta.WindowActor) ancestor.get_compositor_private ();
```
tintou, line 728:
```suggestion
                unowned var win = (Meta.WindowActor) ancestor.get_compositor_private ();
```

src/WindowManager.vala
```diff
@@ -723,19 +723,19 @@ namespace Gala {
...
+                    if (window.window_type == Meta.WindowType.MODAL_DIALOG) {
                         var dark_effect = new Clutter.BrightnessContrastEffect ();
                         dark_effect.set_brightness (-0.4f);
 
                         win.add_effect (dark_effect);
-                    } else {
-                        win.clear_effects ();
                     }
+                } else {
+                    win.clear_effects ();
```
jeremypw, line 738: Since this function only applies the `dark_effect` to `MODAL_DIALOG` windows if `dim` is `true`,  should it not disapply only that effect to only the same window type when `dim` is `false`?   As you say, other effects could be applied to other windows.

danrabbit: @jeremypw possibly? I just know that with master I can reliably reproduce an issue where the "Unsaved changes" dialog in Code will add a darken effect, but when the window closes the effect is not removed. And this branch fixes it for me

tintou (approved):
Looks good, unsure that you actually need to check `if (win.get_effect ("dim-parent") != null` before removing

PR #1310, opened 2021-11-19
Fix various typos

luzpaz:
Found via `codespell -q 3 -S *.po`


jeremypw (approved):
Sounds like a useful tool!  Thanks.

PR #1315, opened 2021-11-24
WindowSwitcher: Simplify the window indicator by handling directly its content

tintou:
Remove the RoundedActor class.


decathorpe:
Fixes crashes when alt-tab switching on Fedora 35.

davidmhewitt (approved):
Nice simplification! I can't see any regressions on elementary OS 6 with this and can confirm the crash fix on Fedora 35.

The behaviour of the indicator is not correct on Fedora 35 compared to elementary Odin, but that wasn't introduced by this branch and that issue exists in master too.

PR #1319, opened 2021-11-28
WindowSwitcher: Follow accent color

### LINKED ISSUE #1316: Window switcher doesn't update to follow accent color
### What Happened?

The new <kbd>Alt</kbd><kbd>Tab</kbd> window switcher doesn't appear to update its accent color when the system accent color changes.

### Steps to Reproduce

1. Have an accent color selected (e.g. Blueberry) and look at the Alt+Tab window switcher
2. Change the accent color to something else, e.g. Orange
3. Look at the window switcher again

### Expected Behavior

It should follow the system accent color without requiring a restart

### OS Version

6.x (Odin)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

JoseExposito:
Repaint the window switcher indicator when the accent color changes.

Fix #1316

---------------------------------

cc @danrabbit 

Because I think she is working on the GTK 4 style sheet. I'd be nice to know if this approach will work on a GTK 4 world or we'd need to add a new API to Granite similar to `prefers-color-scheme` but for accent colours.

### DIFF
--- a/src/Widgets/WindowSwitcher.vala
+++ b/src/Widgets/WindowSwitcher.vala
@@ -43,6 +43,7 @@ namespace Gala {
         }
 
         construct {
+            var gtk_settings = Gtk.Settings.get_default ();
             granite_settings = Granite.Settings.get_default ();
 
             scaling_factor = InternalUtils.get_ui_scaling_factor ();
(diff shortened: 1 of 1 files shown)


cassidyjames (approved):

cassidyjames:
Approving because this works and fixes an issue today. I _think_ we'll be able to do the same approach for GTK4 but since that's pretty early days I wouldn't want to block this from being fixed in the meantime.

PR #1327, opened 2021-12-01
Dialogs: add AccessDialog class

### LINKED ISSUE #1306: “Does the display look OK?” dialogue box layout and colour scheme issue
### What Happened?

The “Does the display look OK?” dialogue box:

  1. Doesn’t use the system colour scheme.
  2. Wraps its label oddly.

Screenshot from 2021-11-12 10-07-31@2x

### Steps to Reproduce

1. Change a display setting.

### Expected Behavior

The dialogue should use the system colour scheme.

### OS Version

6.x (Odin)

### Software Version

### Log Output

```shell
n/a
```

### Hardware Info

n/a

Marukesu:
Use our implementation of  `org.freedesktop.impl.Portal.Access` to show the display change dialog. there's a issue when switchboard-plug-display is open that makes all dialogs  spawn at (0,0). i don't know if it is fixable here or in the plug.

Screenshot from 2021-11-30 19 00 20

we can make this a sort of "frontend" by getting the backend bus name the same way xdg-desktop-portal gets it, if wanted.

fixes #1306

### DIFF
--- a/data/gala.appdata.xml.in
+++ b/data/gala.appdata.xml.in
@@ -10,6 +10,15 @@
     <p>A window &amp; compositing manager based on libmutter and designed by elementary for use with Pantheon.</p>
   </description>
   <releases>
+    <release version="6.3.1" date="2021-12-01" urgency="medium">
+      <description>
+        <p>Improvements:</p>
+        <ul>
+          <li>Use access portal for display settings confirmation dialog</li>
+          <li>Updated translations</li>
+        </ul>
+      </description>
+    </release>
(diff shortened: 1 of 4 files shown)


danrabbit (changes requested):
Hey very cool! This looks way better. I can't confirm that this actually works to keep the new configuration however. It looks like both buttons revert changes.

I tested this both with the existing released of portals and your access rewrite branch for portals

danrabbit (approved):
Nice, works as expected!

PR #1336, opened 2021-12-06
1323: Always subtract from nat_width the scaled padding over children count

### LINKED ISSUE #1323: Alt + Tab task switcher: fix spacing around selection highlight
### What Happened?

The spacing around the selection highlight is irregular.

task-switcher-spacing-1

task-switcher-spacing-2

### Steps to Reproduce

See screenshots.

### Expected Behavior

The spacing should be regular.

### OS Version

6.x (Odin)

### Software Version

### Log Output

```shell
n/a
```

### Hardware Info

n/a

### LINKED ISSUE #1332: Alt-Tab switch app switch enabled for all workspaces instead of one
### Problem

For this idea I dont know it is necessary to write an explanation. 

### Proposal

Just including all application from different workspaces during Alt-Tab app switching.

Also if the terminal is another workspace and someone launches terminal via keyboard shortcut (super+T) , it should automatically go to workspace where the terminal is and keep to focus on the terminal.

### Prior Art (Optional)

_No response_

jpgleeson:
Resolves #1323 . This fixes an off-by-one adjacent error that left the indicator off by an amount based on the scale factor times the children.

The fix can be verified by logging the values for the container.width - (indicator.x + indicator.width) for the final item in the window, and ensuring it is the same as the indicator.x value for the first item in the window.

We will never divide by zero due to the check at the top of the function exiting if the children count is zero.

### DIFF
--- a/src/Widgets/WindowSwitcher.vala
+++ b/src/Widgets/WindowSwitcher.vala
@@ -278,9 +278,7 @@ namespace Gala {
             float nat_width, nat_height;
             container.get_preferred_size (null, null, out nat_width, null);
 
-            if (container.get_n_children () == 1) {
-                nat_width -= WRAPPER_PADDING * scaling_factor;
-            }
+            nat_width -= (WRAPPER_PADDING * scaling_factor) / container.get_n_children ();
 
             container.get_preferred_size (null, null, null, out nat_height);
 


src/Widgets/WindowSwitcher.vala
```diff
@@ -278,9 +278,7 @@ namespace Gala {
             float nat_width, nat_height;
             container.get_preferred_size (null, null, out nat_width, null);
 
-            if (container.get_n_children () == 1) {
-                nat_width -= WRAPPER_PADDING * scaling_factor;
-            }
+            nat_width -= (WRAPPER_PADDING * scaling_factor) / container.get_n_children ();
```
vjr, line 281: Is a division by zero possible with the return of get n children here?

jpgleeson: It is not. Line [243](https://github.com/elementary/gala/blob/349c7b6843c54b6e077d2a730b1628637589e806/src/Widgets/WindowSwitcher.vala#L243) just above checks for a zero return on get_n_children and returns if that’s the count before we reach this statement.

JoseExposito (approved):
Thanks for looking into this. Now it's pixel perfect :rocket:

aral:
Yay, thank you :two_hearts:

Issue #1322, opened 2021-11-30
Can't click after alt+tab


fleury08:
### What Happened?

This happened after update to new gala `Version: 6.3.0~r1571+pkg68~ubuntu6.1`.
When I `alt+tab`, I can't do any click action on any window below the previously shown `alt+tab` window. Appended example video.

[gif of an issue](https://imgur.com/a/EzZUgq7)

### Steps to Reproduce

1. Alt+Tab
2. Try to do mouse click action on the area of previously shown Alt+Tab window

### Expected Behavior

WIndow below should consume the action.

### OS Version

6.x (Odin)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell
lis 30 12:34:19 jaroslav-work gala.desktop[7320]: Window manager warning: Trying to re-add keybinding "switch-to-workspace-last".
lis 30 12:34:19 jaroslav-work gala.desktop[7320]: Window manager warning: Trying to re-add keybinding "move-to-workspace-last".
lis 30 12:34:19 jaroslav-work gala.desktop[7320]: Window manager warning: Overwriting existing binding of keysym 2b with keysym 31 (keycode a).
lis 30 12:34:20 jaroslav-work gala[7320]: STACK_OP_RAISE_ABOVE: sibling window 0x2e00007 (io.element) not in stack
lis 30 12:34:20 jaroslav-work gala[7320]: STACK_OP_RAISE_ABOVE: sibling window 0x2e00007 not in stack
lis 30 12:34:20 jaroslav-work gala[7320]: invalid (NULL) pointer instance
lis 30 12:34:20 jaroslav-work gala[7320]: g_signal_connect_data: assertion 'G_TYPE_CHECK_INSTANCE (instance)' failed
lis 30 12:34:21 jaroslav-work gala[7320]: invalid (NULL) pointer instance
lis 30 12:34:21 jaroslav-work gala[7320]: g_signal_connect_data: assertion 'G_TYPE_CHECK_INSTANCE (instance)' failed
lis 30 12:34:23 jaroslav-work gala[7320]: invalid (NULL) pointer instance
lis 30 12:34:23 jaroslav-work gala[7320]: g_signal_connect_data: assertion 'G_TYPE_CHECK_INSTANCE (instance)' failed
lis 30 12:34:23 jaroslav-work gala[7320]: invalid (NULL) pointer instance
[...]

PR #1341, opened 2021-12-09
switcher: fix container staying reactive in scene

tom95:
I noticed that I was not able to click some windows underneath the new alt-tab switcher.

Here is my hypothesis why this was only rarely noticed (one other report here: https://github.com/elementary/gala/issues/1322 but seemingly with a different cause?):

The container of the switcher stays reactive, however it is pushed underneath the **active** window, meaning no harm is done. However, **inactive** windows are not raised above the switcher, meaning clicks on inactive windows get absorbed by the invisible switcher.

As part of this PR, I extracted all logic that concerns showing/hiding into a toggle_display method, mostly for my own sanity that all "show things" are also properly undone when hiding. The minimal change to fix this behavior would have been to set `container.reactive = true/false` at each `opened = true/false`.

### DIFF
--- a/src/Widgets/WindowSwitcher.vala
+++ b/src/Widgets/WindowSwitcher.vala
@@ -332,7 +332,7 @@ namespace Gala {
             container.reactive = show;
         }
 
-        void push_modal() {
+        void push_modal () {
             modal_proxy = wm.push_modal ();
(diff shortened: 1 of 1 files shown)


JoseExposito (approved):
Thanks for the fix Tom.

Issue #484, opened 2019-03-01
Multitasking view wallpaper DPI switch issue


peteruithoven:
The wallpaper that's shown in the background of workspaces in the Multitasking view doesn't seem to update properly after a DPI switch. 

![screenshot from 2019-03-01 23 53 24 2x](https://user-images.githubusercontent.com/523210/53671190-959a3280-3c7d-11e9-93b1-f0cf8bea55e4.png)

![screenshot from 2019-03-01 23 54 19](https://user-images.githubusercontent.com/523210/53671204-a6e33f00-3c7d-11e9-8b62-05919df211b9.png)

For reference, this is how it should look:
![screenshot from 2019-03-02 00 10 49 2x](https://user-images.githubusercontent.com/523210/53671658-b499c400-3c7f-11e9-904b-222da0004ef4.png)

lenemter:
Fixed in #1344.

PR #1344, opened 2021-12-15
fix workspace clone not resizing on monitor change

tom95:
Workspace clones in the multitasking view haven't been resizing when monitors change (as for example reported in  #484). This should fix it.

Note: the PR contains two unrelated fixes:
* there was a segfault on monitor changes when gala was started from a context where the portal could not be acquired (e.g. when started from a TTY for debugging purposes)
* windows can apparently have NULL titles, e.g. spotify creates a somewhat broken window. Adding the empty string as fallback removes a lot of noise from the output for failed assertions.

### DIFF
--- a/src/Background/BackgroundManager.vala
+++ b/src/Background/BackgroundManager.vala
@@ -56,7 +56,7 @@ namespace Gala {
             }
         }
 
-        void swap_background_actor () {
+        void swap_background_actor (bool animate) {
             return_if_fail (new_background_actor != null);
 
             var old_background_actor = background_actor;
@@ -66,22 +66,27 @@ namespace Gala {
             if (old_background_actor == null)
                 return;
 
-            var transition = new Clutter.PropertyTransition ("opacity");
-            transition.set_from_value (255);
-            transition.set_to_value (0);
-            transition.duration = FADE_ANIMATION_TIME;
-            transition.progress_mode = Clutter.AnimationMode.EASE_OUT_QUAD;
-            transition.remove_on_complete = true;
-            transition.completed.connect (() => {
+            if (animate) {
+                var transition = new Clutter.PropertyTransition ("opacity");
+                transition.set_from_value (255);
+                transition.set_to_value (0);
+                transition.duration = FADE_ANIMATION_TIME;
+                transition.progress_mode = Clutter.AnimationMode.EASE_OUT_QUAD;
+                transition.remove_on_complete = true;
+                transition.completed.connect (() => {
+                    old_background_actor.destroy ();
+
+                    changed ();
+                });
+
+                old_background_actor.add_transition ("fade-out", transition);
+            } else {
                 old_background_actor.destroy ();
-
                 changed ();
-            });
-
-            old_background_actor.add_transition ("fade-out", transition);
+            }
         }
 
-        void update_background_actor () {
+        void update_background_actor (bool animate = true) {
             if (new_background_actor != null) {
                 // Skip displaying existing background queued for load
                 new_background_actor.destroy ();
@@ -108,7 +113,7 @@ namespace Gala {
 #endif
 
             if (background.is_loaded) {
-                swap_background_actor ();
+                swap_background_actor (animate);
                 return;
             }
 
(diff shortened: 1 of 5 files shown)


src/Background/BackgroundManager.vala
```diff
@@ -123,7 +123,9 @@ namespace Gala {
         }
 
         public void set_size (float width, float height) {
-            background_actor.set_size (width, height);
+            if (width != background_actor.width || height != background_actor.height) {
+                update_background_actor ();
```
JoseExposito, line 127: Wouldn't this trigger the fade animation between the old incorrectly sized background actor and the new one?

We'd probably want to skip the animation in this case.

JoseExposito (changes requested):
Looking good, just a little comment.

Also, if you could split the unrelated fixes in different commits it'd help with the code review and changes easier to revert in case of regressions.

src/Background/BackgroundManager.vala
tom95: Good point, I considered it not to be important but you're right, it may look odd in the case that a user manages to stumble upon it. Addressed now :)

tom95:
Thanks for the review! I addressed the two points now :)

src/Widgets/WindowClone.vala
```diff
@@ -409,10 +409,16 @@ namespace Gala {
...
+                if (transition != null) {
+                    transition.completed.connect (() => {
+                        in_slot_animation = false;
+                        place_widgets (outer_rect.width, outer_rect.height);
+                    });
+                } else {
                     in_slot_animation = false;
                     place_widgets (outer_rect.width, outer_rect.height);
-                });
+                }
```
tom95, line 421: NOTE: I allowed myself to push another tiny fix that now finally let's gala start assertion-failure free on my end. On WM startup, the icons already have 0-opacity, as such there is no transition to be had.

src/Background/BackgroundManager.vala
JoseExposito: Missing param?
```diff
-update_background_actor ();
+update_background_actor (false);
 ```

File: src/Widgets/WindowClone.vala (hunk 2 above)
Comment (JoseExposito) [reply src/Widgets/WindowClone.vala:421]: This patch looks good to me, thanks!

**Review by JoseExposito — CHANGES_REQUESTED** [review]
Thanks for rebasing this PR, now it looks much better. I think there is a missing param in the call to `update_background_actor` but other than that this is ready to go.

**Comment by tom95** [comment]
I had `animate` default to true to not change the existing semantics -- did you only miss that or is it something that you'd prefer to have explicit? :)

**Comment by JoseExposito** [comment]
Mmmm no, it's fine to have it as default. But I don't see where the `animate` param is set to `false` to skip the animation on screen scaling change. I'm probably missing it :sweat_smile:

File: src/Background/BackgroundManager.vala (hunk 1 above)
Comment (tom95) [reply src/Background/BackgroundManager.vala]: oh man of course. sorry, good thing you read this carefully. When I tested it I didn't manage to get the animation to show up, so I guess I must have figured that it's disabled.

File: src/Widgets/WorkspaceClone.vala [hunk 3]
```diff
@@ -266,7 +266,7 @@ namespace Gala {
                 remove_window (window);
         }
 
-        void update_size (Meta.Rectangle monitor_geometry) {
+        public void update_size (Meta.Rectangle monitor_geometry) {
             if (window_container.width != monitor_geometry.width || window_container.height != monitor_geometry.height) {
                 window_container.set_size (monitor_geometry.width, monitor_geometry.height);
                 background.set_size (window_container.width, window_container.height);
```
JoseExposito, line 272: It might be a good idea to expose the `animate` parameter in `BackgroundManager.set_size` and default it to `true`.

Right now we only call it here (and we happen to want to avoid the animation) but in the future this might change.

JoseExposito (approved):
Works as intended, thanks for making the suggested changes.

I left a comment about the an improvements we could make in the `BackgroundManager.set_size` API, but it's optional as it is trivial to change in the future.

You might want to `git rebase master` this PR and squash "fix missing param during invocation of update_background_actor" into "fix workspace clone not resizing on monitor geometry changes" for a cleaner git log... Or just press the merge button as it is :)

src/Widgets/WorkspaceClone.vala
tom95: I did actually have this parameter there for a moment as well but ultimately decided against this, thinking that my expectation of an "update_size" function would be that things happen right away. Of course, having the `animate` parameter there, defaulting to false, won't hurt and makes this explicit. I'd say we see when the use case comes up and only then extend the API :)

tom95:
It is now squashed and looking all proper, however I don't have rights to push, so you can go ahead anytime. Thanks for the comments again :)

JoseExposito:
Thanks, merged!

PR #1364, opened 2022-01-14
WindowClone: Remove Granite.CloseButtonPosition

JoseExposito:
The API Granite.CloseButtonPosition has been deprecated [1].

Remove it from Gala and always place the close button on the left of
the window in the multitasking view.

[1] https://github.com/elementary/granite/pull/568

### DIFF
--- a/src/Widgets/WindowClone.vala
+++ b/src/Widgets/WindowClone.vala
@@ -266,7 +266,7 @@ namespace Gala {
          * itself at the location of the original window. Also adds the shadow
          * effect and makes sure the shadow is updated on size changes.
          *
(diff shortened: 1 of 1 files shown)


tintou:
That feature could be re-implemented by using:
```vala
public static bool close_on_left () {
    var layout = Mesa.Pref.get_button_layout ();
    foreach (var button_function in layout.left_buttons) {
        if (button_function == Meta.ButtonFunction.CLOSE) {
            return true;
        }
    }
    
    return false;
}
```

JoseExposito:
Thanks for the tip @tintou, updated to include your function

tintou (approved):

PR #1376, opened 2022-02-03
Play login sound

meisenzahl:
Related to https://github.com/elementary/triage/issues/98

### DIFF
--- a/data/gala.gschema.xml
+++ b/data/gala.gschema.xml
@@ -288,6 +288,14 @@
     </key>
   </schema>
 
+  <schema path="/org/pantheon/desktop/gala/sounds/" id="org.pantheon.desktop.gala.sounds">
+    <key type="b" name="play-login-sound">
+      <default>true</default>
+      <summary>Play login sound</summary>
+      <description>Whether a sound should be played on login.</description>
+    </key>
+  </schema>
+
   <schema path="/org/pantheon/desktop/gala/mask-corners/" id="org.pantheon.desktop.gala.mask-corners">
     <key type="b" name="enable">
       <default>true</default>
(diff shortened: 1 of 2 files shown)


JoseExposito (changes requested):
Thanks for your PR! One question: since the sound is being played on the workspace manager, wouldn't it be played if Gala crashes and is relaunched?

meisenzahl:
@JoseExposito yes if Gala would crash, the sound would be played again.

Do you have any idea how to prevent this?

JoseExposito:
Maybe you could use [org.gnome.SessionManager](https://people.gnome.org/~mccann/gnome-session/docs/gnome-session.html) to handle this case? I'm not sure if there is a signal/method you can use there. Gala already has a `SessionManager` class, so part of the job should be already done.

cassidyjames:
Yeah, I don't think a login sound is desirable by default. I could see the argument for a "ready" sound from an accessibility standpoint, but then it would need to be short and subtle enough to not be annoying every time you turn on your computer. We would also likely want a similar sound on the login screen to signal that's available for typing a password. Regardless, I _think_ that would be off by default?

meisenzahl:
@elementary/ux is there any interest in adding optional system sounds?

SubhadeepJasu:
Regarding the exact sounds, just posting them from here: 
https://github.com/elementary/sound-theme/pull/16
https://github.com/elementary/triage/issues/98
Let me know if these are okay.

vjr:
My opinion is to play this in (perhaps) the greeter and not gala - so its a startup sound and not login - what say?

SubhadeepJasu:
> My opinion is to play this in (perhaps) the greeter and not gala - so its a startup sound and not login - what say?

I like that. But I also made a sound for shutdown. Can that be integrated in?

PR #1384, opened 2022-03-07
Fix compile warnings

jeremypw:
Silences warnings when compiling.

Unclear whether these warnings might be having any practical effect on the operation of Gala or causing any bugs.

Running with this PR has not caused any regressions so far.

### DIFF
--- a/lib/Plugin.vala
+++ b/lib/Plugin.vala
@@ -86,23 +86,23 @@ namespace Gala {
          * It is calculated by the system whenever update_region is called.
          * You can influce it with the custom_region and the track_actor function.
          */
-        public Meta.Rectangle[] region { get; private set; }
-
+        private Meta.Rectangle[] region;
+        public Meta.Rectangle[] get_region () {
+            return region;
+        }
         /**
          * This list will be merged with the region property. See region for
          * more details. Changing this property will cause update_region to be
          * called. Default to null.
          */
-        protected Meta.Rectangle[]? custom_region {
-            get {
-                return _custom_region;
-            }
-            protected set {
-                _custom_region = value;
-                update_region ();
-            }
+        private Meta.Rectangle[]? _custom_region = null;
+        protected Meta.Rectangle[]? get_custom_region () {
+            return _custom_region;
         }
-
+        protected void set_custom_region (Meta.Rectangle[]? custom_region) {
+            _custom_region = custom_region;
+            update_region ();
+        }
         /**
          * Set this property to true while animating an actor if you have tracked
          * actors to prevent constant recalculations of the regions during an
@@ -121,7 +121,6 @@ namespace Gala {
         }
 
         private bool _freeze_track = false;
-        private Meta.Rectangle[]? _custom_region = null;
         private List<Clutter.Actor> tracked_actors = new List<Clutter.Actor> ();
 
         /**
@@ -174,6 +173,7 @@ namespace Gala {
          * wm aware of the new position of the actor in question.
          */
         public void update_region () {
+            var custom_region = get_custom_region ();
             var has_custom = custom_region != null;
             var len = tracked_actors.length () + (has_custom ? custom_region.length : 0);
 
--- a/lib/WindowManager.vala
+++ b/lib/WindowManager.vala
@@ -66,7 +66,14 @@ namespace Gala {
          * Defaults to blocking all.
          * @see KeybindingFilter
          */
-        public KeybindingFilter? keybinding_filter { get; owned set; default = () => true; }
+        private KeybindingFilter? _keybinding_filter = () => true;
+        public KeybindingFilter? get_keybinding_filter () {
+            return _keybinding_filter;
+        }
+
+        public void set_keybinding_filter (KeybindingFilter? filter) {
+            _keybinding_filter = filter;
+        }
 
         public ModalProxy () {
         }
@@ -75,7 +82,7 @@ namespace Gala {
          * Small utility to allow all keybindings
          */
         public void allow_all_keybindings () {
-            keybinding_filter = null;
+            _keybinding_filter = null;
         }
     }
 
--- a/src/PluginManager.vala
+++ b/src/PluginManager.vala
@@ -31,7 +31,10 @@ namespace Gala {
 
         public bool initialized { get; private set; default = false; }
 
-        public X.Xrectangle[] regions { get; private set; }
+        private X.Xrectangle[] _regions = {};
+        public X.Xrectangle[] get_regions () {
+            return _regions;
+        }
 
         public string? window_switcher_provider { get; private set; default = null; }
         public string? desktop_provider { get; private set; default = null; }
@@ -201,7 +204,7 @@ namespace Gala {
             X.Xrectangle[] regions = {};
 
             plugins.@foreach ((name, plugin) => {
-                foreach (var region in plugin.region) {
+                foreach (var region in plugin.get_region ()) {
                     X.Xrectangle rect = {
                         (short) region.x,
                         (short) region.y,
@@ -213,7 +216,7 @@ namespace Gala {
                 }
             });
 
-            this.regions = regions;
+            this._regions = regions;
             regions_changed ();
         }
     }
--- a/src/Widgets/MultitaskingView.vala
+++ b/src/Widgets/MultitaskingView.vala
@@ -590,7 +590,7 @@ namespace Gala {
 
             if (opening) {
                 modal_proxy = wm.push_modal ();
-                modal_proxy.keybinding_filter = keybinding_filter;
+                modal_proxy.set_keybinding_filter (keybinding_filter);
 
                 wm.background_group.hide ();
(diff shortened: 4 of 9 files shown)


src/Widgets/MultitaskingView.vala
```diff
@@ -321,7 +321,10 @@ namespace Gala {
             GestureTracker.OnEnd on_animation_end = (percentage, cancel_action, calculated_duration) => {
                 workspace_gesture_tracker.enabled = false;
 
-                var duration = is_nudge_animation ? (AnimationDuration.NUDGE / 2) : calculated_duration;
+                var duration = is_nudge_animation ?
+                               (uint)(AnimationDuration.NUDGE / 2) :
+                               (uint)(calculated_duration.clamp (0, AnimationDuration.WORKSPACE_SWITCH));
```
JoseExposito, line 326: Couldn't this change the animation duration if the calculated values is greater than `AnimationDuration.WORKSPACE_SWITCH`?

src/WindowManager.vala
```diff
@@ -2042,10 +2042,14 @@ namespace Gala {
                 return false;
 
             var modal_proxy = modal_stack.peek_head ();
+            if (modal_proxy != null) {
```
JoseExposito, line 2045: Nitpick, so feel free to ignore it, but I think this could be easier to follow returning earlier:

```
var modal_proxy = modal_stack.peek_head ();
if (!modal_proxy) {
    return false;
}

var filter = modal_proxy.get_keybinding_filter ();
if (!filter) {
    return false;
}

return filter (binding);
```

JoseExposito (changes requested):
Is the `property { get; private set; }` being deprecated? I liked that :(

Just a couple of minor comments, but looking good. Thanks!

jeremypw:
Afaik the `property {get; private set; }` is not deprecated - it just does not work with some types (or, at least, it generates a warning).

src/Widgets/MultitaskingView.vala
jeremypw: This is done to ensure that the `duration` is within the range of `AnimationDuration`.  Otherwise it could (in theory) be arbitrarily long.  But I can remove this change as it is not strictly in scope.

jeremypw: It was also to fix the warning `Unsafe conversion from `int' to `Gala.AnimationDuration'` but I can try to do that without limiting calculated duration.

src/WindowManager.vala
jeremypw: I dislike having more than 2 `return` s in a function if it can be helped but I can see less nesting could be clearer.

src/Widgets/MultitaskingView.vala
JoseExposito: > This is done to ensure that the duration is within the range of AnimationDuration. Otherwise it could (in theory) be arbitrarily long

That's intended, the gesture handler takes care of the animation duration and should guarantee a reasonable animation time.

> It was also to fix the warning Unsafe conversion from 'int' to 'Gala.AnimationDuration'

If this warning is present in all uses of the signal, we could emit an unsigned int instead of an int to fix all warnings.

src/WindowManager.vala
JoseExposito: Fair enough, both approaches are fine. Resolving this conversation 👍

src/PluginManager.vala
```diff
@@ -31,7 +31,10 @@ namespace Gala {
 
         public bool initialized { get; private set; default = false; }
 
-        public X.Xrectangle[] regions { get; private set; }
+        private X.Xrectangle[] _regions = {};
+        public X.Xrectangle[] get_regions () {
```
Marukesu, line 35:
```suggestion
        public unowned X.Xrectangle[] get_regions () {
```

lib/WindowManager.vala
```diff
@@ -66,7 +66,14 @@ namespace Gala {
...
          * @see KeybindingFilter
          */
-        public KeybindingFilter? keybinding_filter { get; owned set; default = () => true; }
+        private KeybindingFilter? _keybinding_filter = () => true;
+        public KeybindingFilter? get_keybinding_filter () {
+            return _keybinding_filter;
+        }
+
+        public void set_keybinding_filter (KeybindingFilter? filter) {
+            _keybinding_filter = filter;
```
Marukesu, line 75:
```suggestion
            _keybinding_filter =  (owned) filter;
```

src/Widgets/MultitaskingView.vala
```diff
@@ -321,7 +321,10 @@ namespace Gala {
             GestureTracker.OnEnd on_animation_end = (percentage, cancel_action, calculated_duration) => {
                 workspace_gesture_tracker.enabled = false;
 
-                var duration = is_nudge_animation ? (AnimationDuration.NUDGE / 2) : calculated_duration;
+                var duration = is_nudge_animation ?
+                               (uint)(AnimationDuration.NUDGE / 2) :
+                               (uint)calculated_duration;
+
```
Marukesu, line 327: Code style.
```suggestion
                               (uint) (AnimationDuration.NUDGE / 2) :
                               (uint) calculated_duration;

```

lib/Plugin.vala
```diff
@@ -86,23 +86,23 @@ namespace Gala {
          * It is calculated by the system whenever update_region is called.
          * You can influce it with the custom_region and the track_actor function.
          */
-        public Meta.Rectangle[] region { get; private set; }
-
+        private Meta.Rectangle[] region;
+        public Meta.Rectangle[] get_region () {
```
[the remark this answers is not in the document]
jeremypw: Yes. you are right.

lib/WindowManager.vala
jeremypw: This change gives a compile error:
`../lib/WindowManager.vala:75.35-75.48: error: No reference to be transferred`

jeremypw:
Turns out you cannot (or do not need to) transfer ownership of the keybinding filter in the `public void set_keybinding_filter (KeybindingFilter? filter)` function.

jeremypw:
This is the generated C code of that function as is:
```
gala_modal_proxy_set_keybinding_filter (GalaModalProxy* self,
                                        GalaKeybindingFilter filter,
                                        gpointer filter_target)
{
#line 74 "../lib/WindowManager.vala"
	g_return_if_fail (self != NULL);
#line 75 "../lib/WindowManager.vala"
	(self->priv->_keybinding_filter_target_destroy_notify == NULL) ? NULL : (self->priv->_keybinding_filter_target_destroy_notify (self->priv->_keybinding_filter_target), NULL);
#line 75 "../lib/WindowManager.vala"
	self->priv->_keybinding_filter = NULL;
#line 75 "../lib/WindowManager.vala"
	self->priv->_keybinding_filter_target = NULL;
#line 75 "../lib/WindowManager.vala"
	self->priv->_keybinding_filter_target_destroy_notify = NULL;
#line 75 "../lib/WindowManager.vala"
	self->priv->_keybinding_filter = filter;
#line 75 "../lib/WindowManager.vala"
	self->priv->_keybinding_filter_target = filter_target;
#line 75 "../lib/WindowManager.vala"
	self->priv->_keybinding_filter_target_destroy_notify = NULL;
#line 326 "WindowManager.c"
}
```

`KeybindingFilter` is a delegate function so is not reference counted.

JoseExposito (approved):
LGTM! Let's see if @Marukesu wants to add something else

Marukesu:
> Turns out you cannot (or do not need to) transfer ownership of the keybinding filter in the `public void set_keybinding_filter (KeybindingFilter? filter)` function.

@jeremypw sorry for the incomplete suggestion on `set_keybinding_filter`, the parameter should be `owned` too:
```vala
public void set_keybinding_filter (owned KeybindingFilter? filter) {
    _keybinding_filter = (owned) filter;
}
```
> This is the generated C code of that function as is:

when compiling the method outside gala, valac show this warning:
```
warning: copying delegates is not supported
```

I don't known why it don't show the warning when compiling gala.

src/WindowManager.vala
```diff
@@ -2046,10 +2046,16 @@ namespace Gala {
                 return false;
 
             var modal_proxy = modal_stack.peek_head ();
+            if (modal_proxy == null) {
+                return false;
+            }
+
+            var filter = modal_proxy.get_keybinding_filter ();
```
Marukesu, line 2053:
```suggestion
           unowned var filter = modal_proxy.get_keybinding_filter ();
```

lib/Plugin.vala
```diff
@@ -174,6 +173,7 @@ namespace Gala {
          * wm aware of the new position of the actor in question.
          */
         public void update_region () {
+            var custom_region = get_custom_region ();
```
Marukesu, line 176:
```suggestion
            unowned var custom_region = get_custom_region ();
```

Marukesu (commented):
two extra places that should be unowned, otherwise it's all good.

PR #1386, opened 2022-03-07
Remove redundant test

jeremypw:
While studying Gala code it was noticed that an if clause contained an `OR` statement where one condition could never be `true` since it already been dealt with earlier in a switch clause.

In order to keep the behaviour the same as at present, the redundant if statement is just removed. Plugin areas are still not added in the case of InputArea.FULLSCREEN.

### DIFF
--- a/src/InternalUtils.vala
+++ b/src/InternalUtils.vala
@@ -123,11 +123,10 @@ namespace Gala {
                     rects = {topleft, topright, bottomleft, bottomright};
 
                     // add plugin's requested areas
-                    if (area == InputArea.FULLSCREEN || area == InputArea.DEFAULT) {
-                        foreach (var rect in PluginManager.get_default ().regions) {
-                            rects += rect;
-                        }
+                    foreach (var rect in PluginManager.get_default ().regions) {
+                        rects += rect;
                     }
+
                     break;
                 case InputArea.NONE:
                 default:


src/InternalUtils.vala
```diff
@@ -122,19 +122,18 @@ namespace Gala {
...
                     break;
                 case InputArea.NONE:
                 default:
                     unowned Meta.X11Display x11display = display.get_x11_display ();
                     x11display.clear_stage_input_region ();
                     return;
             }
+            // Must be InputArea.FULLSCREEN or InputArea.DEFAULT if reach here
+            // add plugin's requested areas
+            foreach (var rect in PluginManager.get_default ().regions) {
```
JoseExposito, line 134: While it's true that `if (area == InputArea.FULLSCREEN || area == InputArea.DEFAULT)` was redundant, this `foreach` was called only when `area == InputArea.DEFAULT`, wasn't it?

I think that calling it here would change the previous behaviour where this wasn't performed when `area == InputArea.FULLSCREEN`.

Having said that, I don't know it that could cause bugs.

JoseExposito (changes requested):

src/InternalUtils.vala
jeremypw: Yes, I am not totally sure what the intended behaviour actually was - what the if clause implies or what the switch clause implies. :disappointed: I may be safer not to add plugin areas to FULLSCREEN WINDOWS I guess.

JoseExposito: Yes, it is difficult to tell who is lying, the if or the switch xD But just in case, I'd try to keep the behaviour unchanged, just in case.

jeremypw: Done

JoseExposito (approved):
Thanks!!

PR #1400, opened 2022-04-12
[OS 7] Tooltip: Remove text size workaround for Mutter 10

JoseExposito:
Hi! This is a elementary OS 7 bug. I'm not sure whether we are already focusing on v7 on the `master` branch or if I should add a `#ifdef`. 

---------------------------

Previous to Mutter 10 it was necessary to set the text size to large in order to avoid displaying a tiny text in a Clutter.Text.

With Mutter 10, the default text size matches the system font size and the workaround is no longer necessary.

Before:

tooltip-before

After:

tooltip-after

### DIFF
--- a/src/Widgets/Tooltip.vala
+++ b/src/Widgets/Tooltip.vala
@@ -119,10 +119,14 @@ public class Gala.Tooltip : Clutter.Actor {
             color = text_color,
             x = padding.left,
             y = padding.top,
-            ellipsize = Pango.EllipsizeMode.MIDDLE,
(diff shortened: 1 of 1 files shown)


danrabbit:
Hey thanks for that! We probably should have an ifdef here yes until we drop support for older mutter from the codebase. We might still want to release Gala updates to OS 6 as well

JoseExposito:
Thanks Danielle, I just added and tested the required `#ifdefs` :)

tintou (approved):

PR #1407, opened 2022-04-21
Use Clutter.Grab with Mutter 42

tintou:
This is the new way of grabbing the keyboard input.


tintou:
This is the first step to use the new grab methods as per https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/2045

danrabbit (approved):
Can't confirm this fixes any particular issue, but it doesn't break OS 6 :+1:

PR #1408, opened 2022-04-22
Fix window switcher not focussing selected window upon release key.

### LINKED ISSUE #1396: Alt+Tab application switcher doesn't select the highlighted application
### What Happened?

When using the <kbd>Alt</kbd>+<kbd>Tab</kbd> keyboard shortcut, the application chooser doesn't disappear when the <kbd>Alt</kbd> key is released.
In that stage, clicking on the application or "outside" the chooser makes it disappear, but until then is not possible to "confirm" the highlighted application from the chooser.
Press <kbd>Enter</kbd> or <kbd>Space</kbd> doesn't do anything as the focus seems to be back on whatever application is currently visible.

### Steps to Reproduce

1. Install elementary OS 7
2. Open multiple applications
3. Press <kbd>Alt</kbd>+<kbd>Tab</kbd> to cycle through opened apps
4. Release the <kbd>Alt</kbd> key

### Expected Behavior

The highlighted application should be selected when the <kbd>Alt</kbd> key is released, and the switcher should disappear.

### OS Version

7.x (Early Access)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

vjr:
Fixes #1396 

Thanks @tintou for your recent merges, made this fix easy, assuming it's reviewed and looks good!

### DIFF
--- a/src/Widgets/WindowSwitcher.vala
+++ b/src/Widgets/WindowSwitcher.vala
@@ -344,7 +344,7 @@ namespace Gala {
                     || name == "switch-windows" || name == "switch-windows-backward");
             });
 
-            grab_key_focus ();
+            get_stage ().grab (this);
         }
 
         void close_switcher (uint32 time, bool cancel = false) {
--- a/src/Widgets/WindowSwitcher.vala
+++ b/src/Widgets/WindowSwitcher.vala
@@ -308,10 +308,6 @@ namespace Gala {
             if ((get_current_modifiers () & modifier_mask) == 0) {
                 close_switcher (wm.get_display ().get_current_time ());
             }
-
-#if !HAS_MUTTER42
-            grab_key_focus ();
-#endif
         }
 
         void toggle_display (bool show) {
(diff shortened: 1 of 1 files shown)


tintou:
Maybe it was just about removing the `grab_key_focus`
As the `get_stage ().grab` should already be done in https://github.com/elementary/gala/blob/master/src/WindowManager.vala#L683

vjr:
Converting to draft while I see why the build checks are failing on github.

vjr:
Thanks again @tintou yes it seems that extra grab was not needed. Appears to be fixed now!

tintou:
This breaks mutter<42 so I believe that it would be better to do:
```
#if !HAS_MUTTER42
            grab_key_focus ();
#endif
```

vjr:
> This breaks mutter<42 so I believe that it would be better to do:
> 
> ```vala
> #if !HAS_MUTTER42
>             grab_key_focus ();
> #endif
> ```

Done!

tintou (approved):

PR #1409, opened 2022-04-22
Move panel-main-menu schema in house

### LINKED ISSUE #1395: Super + Space doesn't open the applications menu
### What Happened?

Apologies if this is a duplicate issue, or if this is the wrong place to report it. I tried searching around but couldn't find any similar reference.

The <kbd>⌘</kbd>+<kbd>space</kbd> shortcut doesn't trigger the opening of the application menu (top left widget).
The shortcuts is correctly listed in the shortcuts list in the Keyboard switchboard settings.

Setting the <kbd>⌘</kbd> for the Application menu works, but it doesn't if set for the Shortcuts Overlay.

### Steps to Reproduce

1. Install elementary OS 7
2. Press <kbd>⌘</kbd>+<kbd>space</kbd>

### Expected Behavior

The application menu should be opened.

### OS Version

7.x (Early Access)

### Software Version

### Log Output

_No response_

### Hardware Info

Quad-Core Intel® Core™ i7-7700HQ CPU @ 2.80GHz
Intel® HD Graphics 630

davidmhewitt:
Fixes #1395 

The GNOME GSettings key we were piggybacking for this was deprecated a long time ago and mutter has finally stopped exposing those events. So, lets set up our own key.

This will require some work in the keyboard plug to change the setting on this key instead.

### DIFF
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -242,13 +242,10 @@ namespace Gala {
             display.add_keybinding ("move-to-workspace-last", keybinding_settings, 0, (Meta.KeyHandlerFunc) handle_move_to_workspace_end);
             display.add_keybinding ("cycle-workspaces-next", keybinding_settings, 0, (Meta.KeyHandlerFunc) handle_cycle_workspaces);
             display.add_keybinding ("cycle-workspaces-previous", keybinding_settings, 0, (Meta.KeyHandlerFunc) handle_cycle_workspaces);
+            display.add_keybinding ("panel-main-menu", keybinding_settings, 0, (Meta.KeyHandlerFunc) handle_applications_menu);
 
             display.overlay_key.connect (() => {
                 launch_action ("overlay-action");
-            });
-
-            Meta.KeyBinding.set_custom_handler ("panel-main-menu", () => {
-                launch_action ("panel-main-menu-action");
             });
 
             Meta.KeyBinding.set_custom_handler ("toggle-recording", () => {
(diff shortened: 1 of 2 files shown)


tintou:
Can you use an `#ifdef` for Mutter<42 ?

davidmhewitt:
> Can you use an `#ifdef` for Mutter<42 ?

This seems to be a Mutter 41 change (https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/1886) so I've added that in.

src/WindowManager.vala
```diff
@@ -242,15 +242,18 @@ namespace Gala {
             display.add_keybinding ("move-to-workspace-last", keybinding_settings, 0, (Meta.KeyHandlerFunc) handle_move_to_workspace_end);
             display.add_keybinding ("cycle-workspaces-next", keybinding_settings, 0, (Meta.KeyHandlerFunc) handle_cycle_workspaces);
             display.add_keybinding ("cycle-workspaces-previous", keybinding_settings, 0, (Meta.KeyHandlerFunc) handle_cycle_workspaces);
+#if HAS_MUTTER41
+            display.add_keybinding ("panel-main-menu", keybinding_settings, 0, (Meta.KeyHandlerFunc) handle_applications_menu);
+#else
+            Meta.KeyBinding.set_custom_handler ("panel-main-menu", () => {
+                launch_action ("panel-main-menu-action");
+            });
+#endif
```
tintou, line 251: It looks like they can even both share the same codepath
```suggestion
#if HAS_MUTTER41
            display.add_keybinding ("panel-main-menu", keybinding_settings, 0, (Meta.KeyHandlerFunc) handle_applications_menu);
#else
            Meta.KeyBinding.set_custom_handler ("panel-main-menu", (Meta.KeyHandlerFunc) handle_applications_menu);
#endif
```

tintou (approved):

vjr:
> Can you use an `#ifdef` for Mutter<42 ?

How do you deal with versioning of the change in the schema xml?

davidmhewitt:
> > Can you use an `#ifdef` for Mutter<42 ?
> 
> How do you deal with versioning of the change in the schema xml?

Doesn't matter. New Gala will use the new schema, old gala won't.

PR #1417, opened 2022-05-12
Fix initial alt-tab switcher indicator visibility.

### LINKED ISSUE #1412: Application not initially focused/selected in the ALT+TAB switcher
### What Happened?

When using the ALT+TAB application switcher, the list of applications doesn't have any indication of which one is currently focused and therefore will be brought to the foreground.

### Steps to Reproduce

1. Have more than 1 application opened.
2. Press ALT+TAB only once.
3. Notice how the name of the application that will be brought to the foreground is showed, but the icons is hot highlighted.

### Expected Behavior

The icon should be focused/selected, to match the name of the application that will be brought to to the foreground.

### OS Version

7.x (Early Access)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

vjr:
Fixes #1412 

The `get_origin ()` method of the `ActorBox` from the `allocation` property appears to be returning `inf` (infinity?) when I printed the X and Y float values, so making this change seems to resolve the issue.

### DIFF
--- a/src/Widgets/WindowSwitcher.vala
+++ b/src/Widgets/WindowSwitcher.vala
@@ -418,7 +418,7 @@ namespace Gala {
             //       are here too early, in which case all the children are at
             //       (0|0), so we can easily check for that and come back later
             if (container.get_n_children () > 1
-                && container.get_child_at_index (1).allocation.x1 < 1) {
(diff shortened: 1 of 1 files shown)


danrabbit (approved):
Fixes the issue for me, and afaict works on OS 6.1 as well

PR #1419, opened 2022-05-12
WorkspaceClone: draw stroke with Cairo

### LINKED ISSUE #1356: Wallpaper is dimmed in multitasking view
### What Happened?

When entering multitasking view, the wallpaper becomes dimmed

### Steps to Reproduce

1. Open Multitasking View
2. Notice the dim wallpaper

### Expected Behavior

No dimming :)

### OS Version

7.x (Early Access)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

danrabbit:
Fixes #1356 

AFAICT, Cogl.Framebuffer's clip functions only clip things outside of the defined rectangle, not inside. So, instead use Cairo to draw a stroke.

It looks like the black stroke was leftover from before we drew the shadow with `shadow_effect` maybe? It doesn't appear to be necessary anymore

### DIFF
--- a/src/Widgets/WorkspaceClone.vala
+++ b/src/Widgets/WorkspaceClone.vala
@@ -44,19 +44,36 @@ namespace Gala {
         public override void paint (Clutter.PaintContext context) {
             base.paint (context);
 
-            unowned Cogl.Framebuffer fb = context.get_framebuffer ();
+            var surface = new Cairo.ImageSurface (Cairo.Format.ARGB32, (int) width, (int) height);
 
-            pipeline.set_color4ub (0, 0, 0, 100);
-            fb.push_rectangle_clip (0, 0, width, height);
-            fb.draw_rectangle (pipeline, 0, 0, width, height);
-            fb.pop_clip ();
+            var ctx = new Cairo.Context (surface);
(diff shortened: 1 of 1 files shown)


src/Widgets/WorkspaceClone.vala
```diff
@@ -44,19 +44,36 @@ namespace Gala {
         public override void paint (Clutter.PaintContext context) {
             base.paint (context);
 
-            unowned Cogl.Framebuffer fb = context.get_framebuffer ();
-
-            pipeline.set_color4ub (0, 0, 0, 100);
-            fb.push_rectangle_clip (0, 0, width, height);
-            fb.draw_rectangle (pipeline, 0, 0, width, height);
-            fb.pop_clip ();
+            var surface = new Cairo.ImageSurface (Cairo.Format.ARGB32, (int) width, (int) height);
```
tintou, line 47: I have no idea if it is a better thing, but GNOME Shell is using Cairo.Format.A8 to draw its shadow so that it is color agnostic

danrabbit: Hm I'm getting some weird artifacts when trying that

tintou (approved):

Issue #1334, opened 2021-12-05
Hot corner retrigger too sensitive for high resolution monitors or fast mouse cursor speeds


joshuamars:
### What Happened?

When using any of the corners for "Show all windows," the display flashes in and out of all window mode, and typically lands back in normal mode. It looks like it is not tolerating tiny little bounces of the mouse cursor. If I slowly move the mouse cursor into position at the corner, it is stable, and doesn't flash.

### Steps to Reproduce

1 Opened settings > Desktop > Multitasking
2 Set top left corner to "Show all windows"
3 Move mouse quickly into corner causes bug

It looks exactly like what happens when I quickly and repeatedly hit the default keyboard shortcut for show all windows: WINKEY + W or WINKEY + A

### Expected Behavior

I want to quickly move my mouse into the corner, and show all windows, without flashing or being kicked back out into normal mode.

### OS Version

6.x (Odin)

### Software Version

Latest release (I have run all updates)

### Hardware Info

2560x1440 monitor at 144hz

joshuamars:
If I put `notify-send "Poop"` in for "Execute custom command," and then move my cursor into that corner, multiple notifications come up at once, in differing amounts each time.

joshuamars:
This is probably same issue as #1330

joshuamars:
Maybe caused by commit that "fixed" #1100

joshuamars:
Completely disabling the retrigger behavior fixes this issue for me. I commented out line 160 in src/HotCorners/HotCorner.vala and the problem goes away. Adjusting any of the pressure threshold constants wasn't working; I couldn't find a "sweet spot."

joshuamars:
>

@JoseExposito what is the purpose of the retrigger event?

JoseExposito:
Hi! The event is re-triggered  after a certain amount of pressure is applied, check https://github.com/elementary/gala/issues/1078

On my setup, I have to push the corner for a while to trigger the event. Are you using a HiDPI screen or similar? The density of the display might affect this behaviour.

joshuamars:
Yes, I am using a high resolution monitor (2560x1440 at 144hz). My monitor is so big (32") that I have my mouse cursor speed set very fast. I am guessing #1330 is a similar situation (@Lacplesis). For me, the unwanted retriggering happens at anywhere from pressure.abs() = 600 or above. 

[...]

Issue #1330, opened 2021-11-28
Hot Corner for Application Menu glitches on 4K monitor


Lacplesis:
### What Happened?

I have a desktop with a 4K monitor and a laptop with a 1080P monitor and I've used Elementary on both. On the laptop, the applications hot corner is snappy and responsive and always behaves exactly as you would expect. On the 4k monitor however, it's a lot more sensitive and if I flick my cursor into the top left corner, the Applications menu will flicker on and then immediately hide again, sometimes blinking multiple times. If I really ease the cursor in, I can make the menu appear as it should.

Also, it's worth noting that even though the visible applications menu disappears, it still seems to have focus since other windows become unresponsive until I manually click the applications menu and then click off again.

I thought that maybe it was an issue with the hot corners in general at 4k resolution, but I tested that with other hot corner options, and it's only the Applications menu that has the issue. I even tried putting the applications hot corner on other corners and the same glitch occurs.

### Steps to Reproduce

1. with the applications hot corner enabled, on a 4k monitor, drag the cursor to the hot corner
2. Flickering begins on applications menu (if not, repeat and it likely will)

### Expected Behavior

Drag cursor to Applications hot corner and have the applications menu appear.

### OS Version

6.x (Odin)

### Software Version

Latest release (I have run all updates)

### Hardware Info

Ryzen 1600AF CPU
AMD Radeon 570 GPU w/ 4GB RAM
64GB DDR4 RAM

danrabbit:
Transferring to Gala since hotcorners are handled there

danirabbit:
Closing since this was supposedly fixed in #1421

PR #1421, opened 2022-05-16
Changed hotcorners retrigger behavior to adjust for faster mouse pointer speed

joshuamars:
Added a check for mouse pointer speed.
Changed the hotcorners retrigger behavior to adjust for faster mouse pointer speed.

This addresses the problems in issues #1334 and (probably) #1330, while still allowing the retrigger behavior to work properly, as it was implemented for #1078.


src/HotCorners/HotCorner.vala
```diff
@@ -34,6 +34,20 @@ public class Gala.HotCorner : Object {
...
+     * uses this setting to create a scaling factor only for values above 0, 
+     * since smaller values are less sensitive and don't need adjusted
+     */
+    private static GLib.Settings mouse_settings;
+    private static double mouse_speed_scale_factor;
+    static construct {
+        mouse_settings = new GLib.Settings ("org.gnome.desktop.peripherals.mouse");
+        mouse_speed_scale_factor = (mouse_settings.get_double("speed") > 0) ? 
+            (1 + mouse_settings.get_double("speed")) * 2 :
+            1;
```
tintou, line 48: I think that this part should be in a `construct` block as we will also need to watch for the change of the `speed` property

tintou, line 48: Space before starting parenthesis
```suggestion
        mouse_speed_scale_factor = (mouse_settings.get_double ("speed") > 0) ? 
            (1 + mouse_settings.get_double ("speed")) * 2 :
            1;
```

tintou (dismissed):
Thanks for this!
This is indeed a valid approach, just a few things to change.
You can watch for the speed change with:
```vala
mouse_settings.changed["speed"].connect (() => {
    // DO STUFF
});
```

src/HotCorners/HotCorner.vala
joshuamars: Thanks for that! I was wondering how to do that.

joshuamars:
> I wonder whether the issue is the mouse speed or something else. In my case, my mouse speed is also set to 1.0 and I don't experiment this issue:
> 
> ```
> $ gsettings get org.gnome.desktop.peripherals.mouse speed     
> 1.0
> ```
> 
> I think that the issue here are HiDPI screens. @joshuamars are you using any scaling factor?
> 
> image
> 
> If so, a possible approach would be to multiply the pressure by the scaling factor.

My scaling factor is 1.0. I have a larger screen, but not high enough dpi to requiring scaling 🤔. It is 32 US inches, but only 2560x1440

JoseExposito:
Mmm and what's the DPI of your mouse? You can check it online. Let's see if we find what's different in our hardware.

src/HotCorners/HotCorner.vala
```diff
@@ -38,25 +38,57 @@ public class Gala.HotCorner : Object {
      * In order to avoid accidental triggers, don't trigger the hot corner until
      * this threshold is reached.
      */
+    private const int TRIGGER_PRESSURE_THRESHOLD_MIN = 0;
```
JoseExposito, line 41: Why changing the trigger threshold? This threshold is used the first time the hot corner is activated. With my hardware, it makes difficult to trigger it

src/HotCorners/HotCorner.vala
```diff
@@ -38,25 +38,57 @@ public class Gala.HotCorner : Object {
...
     private const int TRIGGER_PRESSURE_THRESHOLD = 50;
 
     /**
      * When the mouse pointer pressures the barrier without activating the hot corner,
      * release it when this threshold is reached.
      */
-    private const int RELEASE_PRESSURE_THRESHOLD = 100;
+    private const int RELEASE_PRESSURE_THRESHOLD_MIN = 30;
```
JoseExposito, line 48: This threshold is mainly used when using more than one screen. The barrier catches the mouse a bit and allows to move it to the next screen if the user keeps pressing the corner. I don't know if we need to change it.

src/HotCorners/HotCorner.vala
```diff
@@ -38,25 +38,57 @@ public class Gala.HotCorner : Object {
...
 
     /**
      * When the mouse pointer pressures the hot corner after activation, trigger the
      * action again when this threshold is reached.
+     * Only retrigger after a minimum delay (milliseconds) since original trigger.
      */
-    private const int RETRIGGER_PRESSURE_THRESHOLD = 500;
+    private const int RETRIGGER_PRESSURE_THRESHOLD_MIN = 60;
+    private const int RETRIGGER_PRESSURE_THRESHOLD = 440;
+    private const int RETRIGGER_DELAY = 600;
```
JoseExposito, line 58: I wonder if just checking for this delay could be a good solution. Do you think it could work?

JoseExposito (changes requested):

src/HotCorners/HotCorner.vala
joshuamars: > I wonder if just checking for this delay could be a good solution. Do you think it could work?

I think so. I don't know why I didn't just try that first. I can test it.

src/HotCorners/HotCorner.vala
joshuamars: Ok

joshuamars:
> Mmm and what's the DPI of your mouse? You can check it online. Let's see if we find what's different in our hardware.

Just for information, my mouse is Logitech G305, with 12000 DPI

JoseExposito:
> Just for information, my mouse is Logitech G305, with 12000 DPI

Wow, that's a good DPI! I think this is what causes the issue. A non-gaming mouse usually have 1500 ~ 2000 DPI.

FWIW, I believe that exploring the timeout only solution could work.

joshuamars:
> > Just for information, my mouse is Logitech G305, with 12000 DPI
> 
> Wow, that's a good DPI! I think this is what causes the issue. A non-gaming mouse usually have 1500 ~ 2000 DPI.
> 
> FWIW, I believe that exploring the timeout only solution could work.

I think that you are right. I over-complicated it to begin with, and was chasing the wrong rabbit anyways.

Latest commit ([c2d8945](https://github.com/elementary/gala/pull/1421/commits/c2d89459198b7939a2536e73683e0e77db9c895f)) removed everything but the timeout solution

JoseExposito (approved):
Thanks for the changes! I like this simple approach, LGTM.

JoseExposito:
@tintou I can't merge until you approve the changes, let us know if you like this approach, please. Thanks!

PR #1424, opened 2022-05-19
Use preferences-desktop-workspaces icon from system

### LINKED ISSUE #957: Ship preferences-desktop-workspaces instead of multitasking-view
## Proposed change
As proposed on Slack I think it would be appropriate to ship `preferences-desktop-workspaces` from the icons theme instead of the current `multitasking-view` because it better reppresent the actual result of the action and also matches the icon used in the Keyboard Plug.
This depends on: https://github.com/elementary/icons/issues/974

danrabbit:
Fixes #957


JoseExposito (approved):
The change seems correct. Are the icons already merged? Not merging the PR myself just in case there is a pending dependency.

danrabbit:
Yes, they are. Thank you!

PR #1427, opened 2022-05-22
Handle GTK4 notification bubbles

ZenitsuDev:
Goes with : https://github.com/elementary/notifications/pull/157

It handles GTK4 notification bubbles by anticipating the unique window title set on the notification bubbles "io.elementary.notifications". The window title is hidden, it just acts more of an identifier. This is done because we can't set the type hint of Windows in GTK4, resulting in the window manager treating notification bubbles as normal windows.

### DIFF
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -1336,7 +1336,7 @@ namespace Gala {
 
             // Notifications are a special case and have to be always be handled
             // regardless of the animation setting
-            if (!enable_animations && window.window_type != Meta.WindowType.NOTIFICATION) {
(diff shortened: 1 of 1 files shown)


danrabbit:
Hm @tintou do you know if there’s a better way to identify these windows if we can’t set type? Or if we can still set type somehow?

tintou:
We need a custom protocol on Wayland, for X11 it's not possible anymore in GTK4 (without custom code poking the X.Window directly which we don't want to invest on)

tintou:
Basically the way to go is to test if https://gitlab.gnome.org/GNOME/vala/-/merge_requests/246 is working and use it

ZenitsuDev:
Closing as non-practical way of solving the problem

PR #1429, opened 2022-05-23
WindowManager: set workspace switch duration directly

danrabbit:
Instead of checking this condition twice, we can just set the value directly when we check it the first time


JoseExposito (approved):
It makes sense, thanks for the fix!

JoseExposito:
@danrabbit there's a conflict with https://github.com/elementary/gala/commit/1f94db16c627f73af5dc69714611815e4691b5e8, you'd need to rebase on top of master.

PR #1432, opened 2022-05-24
Dialogs: use computer-fail for force quit dialog

danrabbit:
Depends on a release of icons with https://github.com/elementary/icons/pull/1107


tintou (approved):

JoseExposito (approved):
@danrabbit now that the other PR is merged, we should be good to go here, shouldn't we?

danrabbit:
@JoseExposito yeah it would probably be okay as long as we remember to re-release icons before releasing another Gala :)

PR #1437, opened 2022-06-09
DBusAccelerator: Use the same version as in the greeter

tintou:
Allows to keep in memory the intend of an accelerator.


jeremypw (approved):
Looks like an accurate copy of the greeter code and all the normal windowing shortcuts still work.

As an aside, it would be less confusing for someone new to the code to rename `action` to `action_id` (as described in the .vala file) but that can be done in another PR if agreed.

PR #1450, opened 2022-08-07
POTFILES: Add src/Dialogs.vala

ryonakano:
So that translators can add translations for the texts in the force quit dialog:

https://github.com/elementary/gala/blob/5f0fcc683f9b0b6b233d1182a33ed429825684c0/src/Dialogs.vala#L152-L159


danrabbit (approved):
Apologies for the oversight! Thanks for fixing

Issue #1413, opened 2022-04-26
Clicking an app in multitasking view doesn't focus that app


danirabbit:
### What Happened?

When using multitasking view, clicking an app doesn't bring that app into focus. There's no effect when clicking apps

### Steps to Reproduce

1. Open multitasking view
2. Click an app to focus it
3. Nothing happens

### Expected Behavior

Clicking an app in the multitasking view should bring it into focus. This is probably a mutter-related regression

### OS Version

7.x (Early Access)

### Software Version

Compiled from git

davidmhewitt:
Fixed in #1464

PR #1457, opened 2022-09-02
Update to latest mutter-11 changes

tintou:
Notably: Meta.Barrier is now a sealed class.


bobby285271:
I am using this change with mutter 43.0, basically it works. multitasking view is still broken to me (don't have time to revisit mutter 42.x but should be the same as #1413) and hot corner does not work (my journal is spammed with `g_signal_emit_valist: assertion 'G_TYPE_CHECK_INSTANCE (instance)' failed` and `invalid (NULL) pointer instance` when I try the hot corner), other things looks fine so far.

(To be clear NixOS is going to continue stick with mutter 3.38 for now)

tintou:
This https://gitlab.gnome.org/GNOME/mutter/-/commit/af65a01057b540cdaa88eb1744fe82c7a82698d8 landed right before the release of Mutter 43 so this has to be reworked

bobby285271:
I can confirm the hot corner now works :partying_face:

PR #1462, opened 2022-09-20
Option to dim wallpapers when dark style is active

### LINKED ISSUE #966: Option to dim bright wallpapers when dark style is active
Pretty useful option that iOS has for brighter wallpapers, it would certainly help with situations like this:
Screenshot from 2020-11-12 23-59-41

lenemter:
Resolves #966

Settings for this branch: elementary/switchboard-plug-pantheon-shell#336

Not sure if it's ok to create a new schema path for one property, but it can be moved to existing one if you so desire

### DIFF
--- a/data/gala.gschema.xml
+++ b/data/gala.gschema.xml
@@ -24,6 +24,14 @@
     </key>
   </schema>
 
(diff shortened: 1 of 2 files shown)


danirabbit (approved):
Works as expected. Nice!

PR #1463, opened 2022-09-26
Fix multitasking view allocation assertions

davidmhewitt:
Fixes the spam of the following assertion that we get when opening and closing the multitasking view:

```
clutter_actor_allocate: assertion '!isnan (real_allocation.x1) && !isnan (real_allocation.x2) && !isnan (real_allocation.y1) && !isnan (real_allocation.y2)' failed
```

This is because `dock_clones` is just an empty actor used to contain the clones of the dock and wingpanel. It initially doesn't have an allocation, and we never change its position, so we don't need to offset by it.

This code is responsible for animating wingpanel and plank in and out during the toggling of multitasking view. It appears to still work manually opening the view, but I haven't tested gestures.

### DIFF
--- a/src/Widgets/MultitaskingView.vala
+++ b/src/Widgets/MultitaskingView.vala
@@ -674,9 +674,6 @@ namespace Gala {
         }
 
         void show_docks (bool with_gesture, bool is_cancel_animation) {
-            float clone_offset_x, clone_offset_y;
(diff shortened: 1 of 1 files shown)


danirabbit (approved):
Can confirm the same behavior as in master with gestures :+1:

PR #1466, opened 2022-10-05
Don't animate workspaces behind multitasking view

### LINKED ISSUE #1354: No window previews in multitasking view after changing workspace
### What Happened?

When entering multitasking view, none of the windows have previews anymore after changing workspaces

### Steps to Reproduce

1. Open multitasking view
2. scroll to switch workspace
3. 
Screenshot from 2022-01-03 11 28 53

### Expected Behavior

Window previews should remain even when switching workspaces

### OS Version

7.x (Early Access)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

davidmhewitt:
Fixes #1354

The multitasking view has its own cloned view of the workspaces and windows that it animates, so we don't need to be animating the actual workspaces and windows behind it, which messes up our window clones.


danirabbit (approved):
Nice fix! Works as expected :)

PR #1468, opened 2022-10-06
Stop using captured_event

### LINKED ISSUE #1467: Unable to drag windows to move them in multitasking view
### What Happened?

When in multitasking view, windows are no longer draggable to other workspaces

### Steps to Reproduce

1. Open some windows
2. Open multitasking view
3. Try to drag a window to another workspace

### Expected Behavior

This is a regression. Behavior should be the same as in OS 6

### OS Version

7.x (Early Access)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

davidmhewitt:
Fixes #1467 

This doesn't compile against previous versions of Mutter, so I still need to add in some conditional compilation and tidy up here.

### DIFF
--- a/lib/DragDropAction.vala
+++ b/lib/DragDropAction.vala
@@ -102,6 +102,10 @@ namespace Gala {
         bool clicked = false;
         float last_x;
         float last_y;
+        Grab? grab = null;
+        static unowned Actor? grabbed_actor = null;
+        InputDevice? grabbed_device = null;
+        ulong on_event_id = 0;
 
         /**
          * Create a new DragDropAction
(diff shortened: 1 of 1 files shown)


davidmhewitt:
There's a lot of code duplication here, but it would just make the `#ifdefs` more complicated, so I'm not going to DRY it. It makes it easier to drop support for older mutter versions in the future if I don't merge all the duplication too.

danirabbit:
This seems to work well until you drag into another workspace or into the workspace containers below and then it looks like it's reverting back to the size of the window clone:

Screenshot from 2022-10-06 12 26 41

If it's not a trivial fix, I think we can track it separately. Functionally this works as expected :)

davidmhewitt:
> This seems to work well until you drag into another workspace or into the workspace containers below and then it looks like it's reverting back to the size of the window clone:

Interesting...

I'll see if I can spot anything tonight. If I've not fixed it by tomorrow, feel free to merge and track separately and I'll do a follow up PR.

danirabbit (approved):
LFG :tada:

PR #1493, opened 2022-11-04
Hide Others

meisenzahl:
Based on a proposal by Reddit user svandragt: https://www.reddit.com/r/elementaryos/comments/ye11rp/hide_others/

@elementary/ux I'm not sure what would be the best solution:

1. hide other windows on all workspaces (https://github.com/elementary/gala/commit/3c0c4e020e77e600a9a4f65e868a439bfc61a1d8)
2. hide only other windows on current workspace (https://github.com/elementary/gala/commit/7e2bc8bf85ddbab68b93f5945d812f19e465f203)

Do we want a keybinding for this?

### DIFF
--- a/daemon/MenuDaemon.vala
+++ b/daemon/MenuDaemon.vala
@@ -33,6 +33,7 @@ namespace Gala {
         private Granite.AccelLabel always_on_top_accellabel;
         private Granite.AccelLabel close_accellabel;
         private Granite.AccelLabel hide_accellabel;
+        private Granite.AccelLabel hide_others_accellabel;
         private Granite.AccelLabel move_accellabel;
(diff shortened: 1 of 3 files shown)


meisenzahl:
I added a keybinding to hide other windows:



@svandragt thank you for your proposal! Do you know if macOS hides

1. other windows on all workspaces
2. only other windows on the current workspace

?

svandragt:
Hi @meisenzahl Sorry I just noticed this, thanks for the PR work.
Tested on the latest macOs (on a macbook pro, no external monitors), it hides other windows on all workspaces.

I was surprised by this, because I expect each workspace to be used for a specific task, so be independent in terms of having a focused window.

PR #1496, opened 2022-11-12
build: require mutter 42

bobby285271:
A quick look at [launchpad](https://launchpad.net/~elementary-os/+archive/ubuntu/stable?field.series_filter=focal) gala is kept on 6.3.1 in elementary OS 6.1, ~~not sure if we want to do this now~~ okay it looks like we want to do this after ~~Odin~~ Horus is out but anyway here is an attempt to remove support for mutter 3.36, 3.38, 40, 41.

It is unclear to me how this is done previously, here is the script I used, so for the 2nd, 3rd, 4th, 5th commits, if you spot any issues (except the manually updated meson.build), it is likely an issue of my script :-)

<details>

Yeah I developed an bad habit of executing C++ files directly as shell script before I retired from competitive programming, so yes you can run this as a shell script :-)

```cpp
#if 0
    g++ -std=c++17 -O2 -Wall -lm "$0" && ./a.out HAS_MUTTER42 $1
    rm ./a.out
    exit
#endif

// WARNING:
//
// I wrote this program in a competitive programming
// way, i.e. very bad code quality and it assumes your
// input is totally valid. But "should works".
//
// Also, YOU NEED TO convert all `#elif` blocks to
// `#if` blocks before running this because this program
// DOES NOT TAKE CARE OF IT.
//
// Place the vala flag you wish to clean up in line 2 (of this file).
// Usage: ./main.cpp <file_location>

#include <bits/stdc++.h>
using namespace std;

const string if_prefix = "#if ";
const string else_prefix = "#else";
const string endif_prefix = "#endif";

bool starts_with(string a, string b)
{
    return a.substr(0, b.size()) == b;
}

bool is_related(string a, string b)
{
    return a.substr(if_prefix.size(), b.size()) == b;
}

struct if_block
{
    bool rm_first;
    bool rm_second;
    bool rm_stat;
    bool is_else_seen;
};

int main(int argc, char **argv)
{
    std::cout << "I assume you know what you are doing :-)\n";
    assert(argc == 3);
    string vala_flag = argv[1];
    string file_location = argv[2];

    string cp_cmd = "cp " + file_location + " " + file_location + ".in";
    system(cp_cmd.c_str());

    ifstream file(file_location + ".in");
    assert(file);
    ofstream ofile(file_location);

    stack<if_block> blocks;
    if_block b;
    b.rm_first = false;
    b.rm_second = false;
    b.is_else_seen = false;
    b.rm_stat = true;
    blocks.push(b);

    string line;
    while (getline(file, line))
    {
        assert(blocks.size() > 0);
        b = blocks.top();
        if (starts_with(line, endif_prefix))
        {
            if (b.rm_stat == false)
            {
                ofile << line << endl;
            }
            blocks.pop();
            continue;
        }
        if (starts_with(line, else_prefix))
        {
            if (b.rm_stat == false)
            {
                ofile << line << endl;
            }
            blocks.pop();
            b.is_else_seen = true;
            blocks.push(b);
            continue;
        }
        if (!starts_with(line, if_prefix))
        {
            if (b.is_else_seen == false)
            {
                if (b.rm_first == false)
                {
                    ofile << line << endl;
                }
            }
            else
            {
                if (b.rm_second == false)
                {
                    ofile << line << endl;
                }
            }
            continue;
        }
        if_block a;
        a.is_else_seen = false;
        if ((b.is_else_seen == false && b.rm_first == true) ||
            (b.is_else_seen == true && b.rm_second == true))
        {
            a.rm_first = true;
            a.rm_second = true;
            a.rm_stat = true;
        }
        else if ((!is_related(line, vala_flag)) && (!is_related(line, "!" + vala_flag)))
        {
            a.rm_stat = false;
            a.rm_first = false;
            a.rm_second = false;
        }
        else
        {
            a.rm_stat = true;
            if (is_related(line, vala_flag))
            {
                a.rm_first = false;
                a.rm_second = true;
            }
            else
            {
                a.rm_first = true;
                a.rm_second = false;
            }
        }
        blocks.push(a);
        if (a.rm_stat == false)
        {
            ofile << line << endl;
        }
    }
    file.close();
    assert(blocks.size() == 1);
    string rm_cmd = "rm " + file_location + ".in";
    system(rm_cmd.c_str());
}
```

</details>

The "vapi: Remove unused files" commit is done in a manual way, you may notice that this PR has 1000+ lines of additions, this is likely because `vapi/mutter-cogl-10.vapi` is updated from a symbolic link to an actual file (and the old "actual file" `vapi/mutter-cogl-6.vapi` is removed). But anyway feel free to request changes or force push this PR if you spot random things are messed up.

This also includes an removal of the `vapi/mutter-cogl.vapi` file, I guess that is probably for even older (< 3.36) mutter? :thinking: 

Compiles against 42 and 43 and briefly tested both.


tintou:
We can only remove the support once elementary Odin is out

danirabbit:
It looks like Gala already can't really be released on OS 6, so I think it's okay to drop support for anything before horus/jammy

lenemter (approved):
Looks good. I'll merge this since OS 7 is out.

Issue #1456, opened 2022-08-25
Gala Crashes when Notifications are Displayed


naftulikay:
### What Happened?

If <kbd>Do Not Disturb</kbd> is <kbd>Off</kbd> and notifications come in from Slack (installed via Flatpak: `com.slack.Slack`) or Signal Desktop (installed via Signal's Debian repository), **Gala often crashes**. It probably also occurs from other applications as well, such as Discord (installed via Flatpak: `com.discordapp.Discord`).

If I enable <kbd>Do Not Disturb</kbd>, I cannot reproduce this issue, so somehow the notification system is crashing Gala.

This means that if I want stability, I have to have <kbd>Do Not Disturb</kbd> on, so I can't see notifications.

### Steps to Reproduce

Install Slack, Discord, and/or Signal Messenger, and have a friend message you, creating notifications. Often, this will lead to Gala crashing. It's certainly possible that other notifications will cause it to crash. It doesn't happen every time, but often enough that I leave <kbd>Do Not Disturb</kbd> on.

### Expected Behavior

Gala should not crash when notifications arrive.

### OS Version

6.x (Odin)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell
Aug 24 13:25:53 hollow gala.desktop[62865]: Window manager warning: Buggy client sent a _NET_ACTIVE_WINDOW message with a timestamp of 0 for 0x1e00003
Aug 24 13:26:20 hollow gala[62865]: clutter_actor_add_effect_with_name: assertion 'CLUTTER_IS_ACTOR (self)' failed
Aug 24 13:26:31 hollow gala.desktop[62865]: Window manager warning: Buggy client sent a _NET_ACTIVE_WINDOW message with a timestamp of 0 for 0x1e00003
Aug 24 13:26:37 hollow gala.desktop[62865]: Window manager warning: Invalid WM_TRANSIENT_FOR window 0x0 specified for 0x9800057.
Aug 24 13:26:43 hollow gala[62865]: clutter_actor_add_effect_with_name: assertion 'CLUTTER_IS_ACTOR (self)' failed
Aug 24 13:31:44 hollow gala[62865]: clutter_actor_set_child_below_sibling: assertion 'child->priv->parent == self' failed
Aug 24 13:31:44 hollow gala[62865]: clutter_actor_set_child_above_sibling: assertion 'child->priv->parent == self' failed
[...]

PR #1497, opened 2022-11-24
Work around crash when receiving notifications

GranPC:
As reported in elementary/notifications#139 and elementary/gala#1456, the compositor can sometimes crash when a notification is received. This can be potentially catastrophic: the compositor does get restarted by the session daemon, but if the crash occurs twice within one minute, the session is made unusable, and all unsaved changes will be lost unless you know the magic incantations.

While I'm not yet 100% sure whether this is an upstream issue or something we are doing wrong on our side (see https://github.com/elementary/notifications/issues/139#issuecomment-1325614603 and the following comment) I do see a lot of value in including this workaround, as a stop-gap solution.

There's still a little bit of bugginess if a notification is received during a workspace switch; they may display on the upper left corner of the screen, rather than the upper right corner, and they may overlap. I don't know how to fix this in an efficient way... but at least, we don't crash anymore.

The original issues can be reproduced by running `watch -n0.2 notify-send hello` in a terminal, and then switching workspaces around a couple times. I have tried these repro steps on both 6.0 and Horus, and applying this patch seems to fix the crash with no adverse side-effects that I have observed so far.

### DIFF
--- a/src/NotificationStack.vala
+++ b/src/NotificationStack.vala
@@ -108,7 +108,10 @@ public class Gala.NotificationStack : Object {
         var y = stack_y + TOP_OFFSET + add_y + ADDITIONAL_MARGIN * scale;
         var i = notifications.size;
         var delay_step = i > 0 ? 150 / i : 0;
-        foreach (var actor in notifications) {
+        var iterator = 0;
+        // Need to iterate like this since we might be removing entries
+        while (notifications.size > iterator) {
+            var actor = notifications.get (iterator);
             if (animate) {
                 actor.save_easing_state ();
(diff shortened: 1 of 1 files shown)


zeebok (commented):
Code-wise this looks alright. Very good find! I am going to do some testing right now before I give it approval.

zeebok (approved):
This seem to do the trick for me. I could not get it to crash again no how much workspace switching I was doing.

zeebok:
That said hopefully someone more familiar with Gala can chime in for what the real fix would be if this really is only a workaround.

PR #1498, opened 2022-12-18
Improve multitasking view animation performance

GranPC:
This animation is currently super stuttery, especially on high-resolution devices. It was getting on my nerves. The reason it's slow is because we constantly reallocate resources instead of just scaling the actors we already have. On top of that, we also allocate a new texture every single time a `FramedBackground` was painted. Allocating stuff in the render loop is generally a bad idea.

So I removed a bunch of seemingly useless code and started caching resources in `FramedBackground` - and now it runs at 240fps on my laptop even with power savings enabled!

This is marked "draft" because I have no idea what I'm doing, and I assume there was a good reason for all the complexity inside `WindowClone.allocate`. That being said, after mercilessly ripping out all that code, everything seems to work OK on my end. Just opening this PR to get some feedback and hopefully understand why/whether this complexity is necessary.

### DIFF
--- a/src/Widgets/WindowClone.vala
+++ b/src/Widgets/WindowClone.vala
@@ -445,8 +445,19 @@ public class Gala.WindowClone : Clutter.Actor {
             return;
 
         clone.set_scale (scale_factor, scale_factor);
-        clone.set_position ((input_rect.x - outer_rect.x) * scale_factor,
-            (input_rect.y - outer_rect.y) * scale_factor);
+
+        // Scaling happens around the pivot point, so we need to move the clone
+        // to compensate for the difference between the pivot point and the
+        // top left corner of the clone.
+
+        float pivot_x, pivot_y;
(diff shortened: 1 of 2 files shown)


GranPC:
Oh, looks like this breaks the little focus ring around windows when you use the arrow keys. Working on it.

GranPC:
Should be all good now.

tintou (approved):
Wow, thanks for working on this!

lenemter (changes requested):
Focus ring around windows seems to be a bit misplaced.

`master`:
image

This branch:
image

GranPC:
Great catch, thanks! Never claimed to be good at math :)

lenemter (approved):

lenemter:
Looks good. I'm going to merge this because @tintou approved and no new issues found. Thank you for your work.

PR #1499, opened 2022-12-21
Fix awkward two-finger scroll in multitasking view

GranPC:
Easing wasn't being correctly used when scrolling between workspaces with the touchpad in the multitasking view. There were two problems with this: scrolling seemed to lag behind the user's fingers sometimes, and letting go of the touchpad would almost always cause a jarring jump.

This little patch fixes both of these issues, making everything silky smooth.

If someone's not clear on what specifically we're solving here, I could record my screen and my hand on my touchpad before & after this patch. It really is night and day.

### DIFF
--- a/src/Widgets/MultitaskingView.vala
+++ b/src/Widgets/MultitaskingView.vala
@@ -316,8 +316,10 @@ namespace Gala {
                                (uint) (AnimationDuration.NUDGE / 2) :
                                (uint) calculated_duration;
 
+                workspaces.save_easing_state ();
                 workspaces.set_easing_duration (duration);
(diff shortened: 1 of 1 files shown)


lenemter (commented):
Fixes the issue. But is it necessary to introduce a one line one-time method? IMO it would be better to just put comment and `workspaces.set_easing_duration (0);` in the else block.

GranPC:
Probably not. I created the method thinking I'd have to reset more state, but it turns out that one line was sufficient.

lenemter (approved):
Nice work! :heart:

GranPC:
Thanks :) Only got a couple things left to fix here, and then Gala will be practically perfect for me! :rocket:

PR #1500, opened 2022-12-21
ShadowEffect: let Clutter know the shadow's size

### LINKED ISSUE #1443: Window switcher shadow persists until overdrawn, after latest updates
### What Happened?

I've found two issues:

1. When you have at least two windows open and there is a desktop background visible between them and you hit alt-tab then some strange artifacts are present after switcher disappears, please see the video attached,
2. I quite often make use of keyboard shortcuts for tiling windows either to left or right. When you use alt-tab this function stops working until you change focus to other window or display multitasking view.

### Steps to Reproduce

1. Open at least two windows
2. Make sure there is a space between so you can see the desktop background
3. Hit alt-tab and observe strange shadow artifacts on the screen
4. Now you can't also tile windows with predefined shortcuts until you change window focus by hand

https://user-images.githubusercontent.com/1751891/177555670-31105413-80f6-417c-84dc-0fea0817b011.mp4

### Expected Behavior

No artifacts should be present and window tile shortcuts should work as expected after using alt-tab.

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

GranPC:
Seems to stop shadows from lingering around. Fixes #1443, hopefully. Have only tested this for a couple minutes.

### DIFF
--- a/src/ShadowEffect.vala
+++ b/src/ShadowEffect.vala
@@ -159,5 +159,18 @@ namespace Gala {
 
             return bounding_box;
         }
+
+        public override bool modify_paint_volume (Clutter.PaintVolume volume) {
+            var size = shadow_size * scale_factor;
+            volume.set_width (volume.get_width () + size * 2);
+            volume.set_height (volume.get_height () + size * 2);
+
+            var origin = volume.get_origin ();
+            origin.x -= size;
+            origin.y -= size;
+            volume.set_origin (origin);
+
+            return true;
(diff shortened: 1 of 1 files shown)


lenemter (approved):
It actually solves the issue. Wow thank you so much :smile:

GranPC:
Thanks for trying it out & merging. More soon. :)

PR #1501, opened 2022-12-22
Fix quick zooming

### LINKED ISSUE #1484: Quickly zooming in and out will break zoom until restart
### What Happened?

Quickly zooming the screen in and out with Super + '+' / Super + '-' will cause the zoom focus to jolt between the top left of the screen, and the mouse cursor. It seems to most frequently happen when zooming in and out of 100%, however it still happens at higher zoom levels.
The effect stays until the computer is restarted. It happens on both OS7 and 6.1, so it is not another byproduct of upgrading to OS7. It also happens regardless of graphics card and drivers, tested on an RX 6700 XT, RTX 3060, and GTX 750 Ti.

### Steps to Reproduce

Quickly zoom the screen in and out a bunch of times using Super + '+' and Super + '-'

**(Serious photosensitivity warning for this video)**

https://user-images.githubusercontent.com/42557645/196416833-6a8169a7-11ef-4e78-aa98-d803a8a36d59.mp4

### Expected Behavior

To smoothly and beautifully zoom in and out, following the mouse with such elegance and grace, as usual :)

### OS Version

7.x (Early Access)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

lenemter:
Fixes #1484.

Review with "Hide whitespace".

While we're here:
* Put Gala namespace inline
* Update copyright header

### DIFF
--- a/src/Zoom.vala
+++ b/src/Zoom.vala
@@ -2,7 +2,7 @@
  * Copyright 2022 elementary, Inc. (https://elementary.io)
  * Copyright 2013 Tom Beckmann
  * Copyright 2013 Rico Tzschichholz
- * SPDX-License-Identifier: LGPL-3.0-or-later
+ * SPDX-License-Identifier: GPL-3.0-or-later
  */
 
 public class Gala.Zoom : Object {
--- a/src/Zoom.vala
+++ b/src/Zoom.vala
@@ -147,7 +147,7 @@ public class Gala.Zoom : Object {
         if (current_zoom <= MIN_ZOOM) {
             current_zoom = MIN_ZOOM;
 
-            if (mouse_poll_timer > 0)  {
+            if (mouse_poll_timer > 0) {
                 Source.remove (mouse_poll_timer);
             }
             mouse_poll_timer = 0;
--- a/src/Zoom.vala
+++ b/src/Zoom.vala
@@ -45,8 +45,8 @@ public class Gala.Zoom : Object {
 
         if (mouse_poll_timer > 0) {
             Source.remove (mouse_poll_timer);
+            mouse_poll_timer = 0;
         }
-        mouse_poll_timer = 0;
     }
 
     [CCode (instance_pos = -1)]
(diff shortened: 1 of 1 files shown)


src/Zoom.vala
```diff
@@ -1,183 +1,180 @@
...
-
-            gesture_tracker = new GestureTracker (ANIMATION_DURATION, ANIMATION_DURATION);
-            gesture_tracker.enable_touchpad ();
-            gesture_tracker.on_gesture_detected.connect (on_gesture_detected);
-        }
+/*
+ * Copyright 2022 elementary, Inc. (https://elementary.io)
+ * Copyright 2013 Tom Beckmann
+ * Copyright 2013 Rico Tzschichholz
+ * SPDX-License-Identifier: LGPL-3.0-or-later
```
tintou, line 5:
```suggestion
 * SPDX-License-Identifier: GPL-3.0-or-later
```

src/Zoom.vala
```diff
@@ -1,183 +1,180 @@
...
-                    wins.set_pivot_point (0.0f, 0.0f);
-                }
+        current_zoom += delta;
+        var animation_duration = animate ? ANIMATION_DURATION : 0;
 
-                return;
+        if (wins_handler_id > 0) {
+            wins.disconnect (wins_handler_id);
+        }
+        wins_handler_id = 0;
```
tintou, line 145: Shouldn't this come inside the if clause?

src/Zoom.vala
```diff
@@ -1,183 +1,180 @@
...
 
-            if ((gesture.fingers == 3 && GestureSettings.get_string ("three-finger-pinch") == "zoom") ||
-                (gesture.fingers == 4 && GestureSettings.get_string ("four-finger-pinch") == "zoom")
-            ) {
-                zoom_with_gesture (gesture.direction);
-            }
+        if (mouse_poll_timer > 0) {
+            Source.remove (mouse_poll_timer);
         }
+        mouse_poll_timer = 0;
```
tintou, line 49: Shouldn't this come inside the if clause?

src/Zoom.vala
```diff
@@ -1,183 +1,180 @@
...
+        }
+        wins_handler_id = 0;
+
+        if (current_zoom <= MIN_ZOOM) {
+            current_zoom = MIN_ZOOM;
+
+            if (mouse_poll_timer > 0) {
+                Source.remove (mouse_poll_timer);
             }
+            mouse_poll_timer = 0;
```
tintou, line 153: Same here :slightly_smiling_face:

src/Zoom.vala
lenemter: Well, such checks here were done this way, so I followed it. I can fix this if you want.

tintou (approved):
Thanks for this!

PR #1504, opened 2022-12-26
Fix close dialog

### LINKED ISSUE #1454: '"Window" is not responding' dialog does stay when triggering program is already closed
### What Happened?

I closed Chrome, because it was hanging. The "not responding" dialogs stayed.

### Steps to Reproduce

0. have a hanging program (maybe you can create a demo for that)
1. get "program is hanging" dialog
2. close program
3. the "hanging" dialog stays

### Expected Behavior

i think the dialogs should disappear, because the hanging situation is resolved. and most of the time that happens, but not always

when i use the "force quit" button from the dialog, all dialogs dissappear (https://youtu.be/WkmW4PAtCI8?t=482)

here they stay: https://youtu.be/WkmW4PAtCI8?t=667 (also gala crashes in the background https://github.com/elementary/gala/issues/1449)

### OS Version

Other Linux

NixOS 22.05.2397.5c211b47aea

### Software Version

gala-6.3.1

### Log Output

_No response_

### Hardware Info

- Intel i9-9900K processor with 3.6GHz
- G.Skill Ripjaws 32 GB DDR4 RAM (3200 MHz)
- Samsung 970 EVO Plus 1TB NVMe M.2 SSD
- AMD Radeon RX 6600 XT graphics card with 8 GB VRAM
- Gigabyte Z390 UD motherboard
- Noctua NH-D15 cooler
- be quiet! Straight Power 11 650 Watt power supply unit
- Antec One Midi Tower case

lenemter:
Fixes #1454 
Addresses https://github.com/elementary/switchboard/issues/245

### DIFF
--- a/src/Dialogs.vala
+++ b/src/Dialogs.vala
@@ -119,6 +119,8 @@ namespace Gala {
     }
 
     public class CloseDialog : AccessDialog, Meta.CloseDialog {
+        private const int ALIVE_TIMEOUT = 5000;
+
         public Meta.Window window {
             owned get { return parent; }
             construct { parent = value; }
@@ -174,7 +176,7 @@ namespace Gala {
 
         public new void show () {
             remove_timeout ();
-            timeout_id = Timeout.add (5000, () => {
+            timeout_id = Timeout.add (ALIVE_TIMEOUT, () => {
                 window.check_alive (window.get_display ().get_current_time_roundtrip ());
                 return Source.CONTINUE;
             });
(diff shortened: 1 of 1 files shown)


Marukesu:
i don't like this, elementary/switchboard#245 only happens because it's handle polkit synchronous, that makes the application froze while the call goes on and then the close dialog appears, it's should be handling those case asynchronous.

for #1454, i believe this was an bug because of the gala crash, and mutter in normal conditions will call hide() on the dialog, but if it not the case, we should be calling hide() on the dialog when the window closes. we can query if an window has an close dialog or not, we already do this when creating dialog to not create more than one dialog per window.

lenemter:
@Marukesu, makes sense. I'll close this.

PR #1506, opened 2022-12-28
Draft: Fix PiP window being undraggable

GranPC:
Dragging PiP views was practically impossible on my end; once I started dragging a PiP window, it would stop listening for any further drag events and I'd need to let go and start dragging again if I wanted to move it more than ~30px.

I assume there was a good reason reactivity was being disabled here, so that's why I'm marking this as a draft. I haven't noticed any adverse effects from removing this line, and PiP is now usable on my end.

Before:

https://user-images.githubusercontent.com/1385480/209759009-eacb95ed-a828-465b-a185-a62d448a523b.mp4

After:

https://user-images.githubusercontent.com/1385480/209759019-743772ff-051c-491f-868b-08122f841953.mp4

### DIFF
--- a/lib/DragDropAction.vala
+++ b/lib/DragDropAction.vala
@@ -316,8 +316,6 @@ namespace Gala {
                             ungrab_actor ();
                             grab_actor (handle, event.get_device ());
 
-                            handle.reactive = false;
-
                             var source_list = sources.@get (drag_id);
                             if (source_list != null) {
                                 var dest_list = destinations[drag_id];


GranPC:
NB: I'm seeing a lot of these messages when interacting with PiP views - even before this patch:

```
(gala:29394): Clutter-WARNING **: 06:10:36.915: (../clutter/clutter/clutter-actor.c:12457):clutter_actor_event: runtime check failed: (retval == CLUTTER_EVENT_PROPAGATE)
```

I'm not 100% sure what it means, but returning `false` instead of `true` in `PopupWindow.enter_event` and `PopupWindow.leave_event` makes them go away with no side-effects as far as I can tell.

lenemter (approved):
I don't know why reactivity was being disabled here too, but no bugs were found, so I'll merge this for now. We can always revert this if new issues are found.

PR #1507, opened 2022-12-28
Fix PiP positioning offset

### LINKED ISSUE #849: Picture-in-Picture area selection being offset
## Prerequisites

## Describe the bug

The area selection in Picture-in-Picture is being offset in both X and Y axis.

In the recording bellow, you can see that only the orange area is selected but the Picture-in-Picture window shows the white portion around the selection.

### Expected behavior

The Picture-in-Picture window should display exactly what was selected.

## Platform Information

The bug is present on all my devices, 2 laptops and 1 desktop. Each with completely different hardware but all using Hera 5.1.5.

One of the notebooks was outdated (Hera 5.1.2) and only started showing the behavior after updating and rebooting (still worked fine before the reboot).

- [x] I'm using the latest released stable version

## Additional context

I've used Picture-in-Picture for a long time and it used to work fine. I've been experiencing this bug for quite some time now (can't tell exactly when it started) and I just lived with it hoping for an update to fix the issue.

GranPC:
Fixes #849 

When choosing a PiP region for certain windows with server-side decorations (such as shadows), the clipped window we displayed in the PiP overlay would be ultimately offset by (what I believe is) the Gala-side decorated area of the window - that is, the window shadow.

What this means is that for most apps, users would see a slightly different region of the window than they chose. This didn't happen for all applications, but it did happen for some very common ones (Firefox, Chrome, Terminal...)

This patch makes it so that we now take this into account before setting the clip rect for the PiP view.

### DIFF
--- a/plugins/pip/Main.vala
+++ b/plugins/pip/Main.vala
@@ -79,6 +79,12 @@ public class Gala.Plugins.PIP.Plugin : Gala.Plugin {
                 int point_x = x - (int)active.x;
                 int point_y = y - (int)active.y;
 
+                // Compensate for server-side window decorations
+                var input_rect = active.get_meta_window().get_buffer_rect ();
+                var outer_rect = active.get_meta_window().get_frame_rect ();
+                point_x -= outer_rect.x - input_rect.x;
+                point_y -= outer_rect.y - input_rect.y;
+
                 var rect = Graphene.Rect.alloc ();
                 rect.init (point_x, point_y, width, height);
 
--- a/plugins/pip/Main.vala
+++ b/plugins/pip/Main.vala
@@ -80,8 +80,8 @@ public class Gala.Plugins.PIP.Plugin : Gala.Plugin {
                 int point_y = y - (int)active.y;
 
                 // Compensate for server-side window decorations
-                var input_rect = active.get_meta_window().get_buffer_rect ();
-                var outer_rect = active.get_meta_window().get_frame_rect ();
+                var input_rect = active.get_meta_window ().get_buffer_rect ();
+                var outer_rect = active.get_meta_window ().get_frame_rect ();
                 point_x -= outer_rect.x - input_rect.x;
                 point_y -= outer_rect.y - input_rect.y;
 


GranPC:
This one's a little tricky to explain if you haven't experienced it before. Here's a screencast showing before/after:

https://user-images.githubusercontent.com/1385480/209762386-bd73e90d-b09e-4750-805b-d0be4cb7b90b.mp4

lenemter (commented):
For me the bug occurs for both server-side and client-side decorations. Can you confirm that it does for you too?

GranPC:
Got an example for both of these? I assumed the window shadow was always server-side rendered, and that it was the root cause.

lenemter:
Here's the video of this bug in Google Chrome and in System Settings

https://user-images.githubusercontent.com/80143868/209781813-073dcc11-748f-4422-a66e-69019cf95d76.mp4

GranPC:
Oh, yup, I actually tried these two exact apps. This PR fixes that.

lenemter:
Can you change the comment in code please since the issue confirmed for both types of shadows?

GranPC:
I do still think it's caused by the shadow (or other decorations) on the Gala side, which clients don't render by themselves. Am I mistaken and do some apps render their own shadows?

lenemter:
AFAIK Gtk apps render their own shadows. I don't know how shadows work in gtk, maybe it somehow reports the shadow size to the gala?

lenemter:
You should ask someone familiar with this stuff like @tintou or @davidmhewitt

GranPC:
Yeah agreed. I'm honestly not really sure what the exact difference between `buffer_rect` and `frame_rect` is - I just intuitively coded this up after having lived with this bug for a few years. Definitely would be great to have someone who knows this stuff look into it :)

tintou:
Here is the documentation for get_buffer_rect:
```c
/**
 * meta_window_get_buffer_rect:
 * @window: a #MetaWindow
 * @rect: (out): pointer to an allocated #MetaRectangle
 *
 * Gets the rectangle that the pixmap or buffer of @window occupies.
 *
 * For X11 windows, this is the server-side geometry of the toplevel
 * window.
 *
 * For Wayland windows, this is the bounding rectangle of the attached
 * buffer.
 */
```
and get_frame_rect:
```c
/**
 * meta_window_get_frame_rect:
 * @window: a #MetaWindow
 * @rect: (out): pointer to an allocated #MetaRectangle
 *
 * Gets the rectangle that bounds @window that is what the user thinks of
 * as the edge of the window. This doesn't include any extra reactive
 * area that we or the client adds to the window, or any area that the
 * client adds to draw a client-side shadow.
 */
```
So here we aren't taking the client side shadow into account in both cases

lenemter (approved):

PR #1510, opened 2023-01-03
Release 7.0.0

danirabbit:
GranPC has some other branches waiting for review, if we can get to more that would be awesome. But this already contains a lot of performance fixes and crash fixes so really would like to get this in

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,6 +1,6 @@
 project('gala',
     'c', 'vala',
-    version: '6.3.3',
+    version: '6.4.0',
     meson_version: '>= 0.50.0',
     license: 'GPL3',
 )
--- a/meson.build
+++ b/meson.build
@@ -1,6 +1,6 @@
 project('gala',
     'c', 'vala',
-    version: '6.4.0',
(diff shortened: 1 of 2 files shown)


meson.build
```diff
@@ -1,6 +1,6 @@
 project('gala',
     'c', 'vala',
-    version: '6.3.3',
+    version: '6.4.0',
```
meisenzahl, line 3: You have increased the version of Feedback to 7.0.0. Does that make sense here too?

meisenzahl (changes requested):

meson.build
danirabbit: I'll spin up a VM today to test, but I think this might still be releasable on 6.1 maybe. If it contains breaking changes then yes we should bump to 7

danirabbit: Okay yeah gonna bump to 7. It looks like we're relying on some newer icons and this doesn't actually fix some of the artifacting issues etc in OS 6, so probably not much point trying to backport it.

Issue #1505, opened 2022-12-26
"Not responding" dialog and not responding app close after inactivity


lenemter:
### What Happened?

"Not responding" dialog and not responding app close after inactivity because
`Dialogs.vala:198: Timeout was reached`

### Steps to Reproduce

1. Open "Not responding" dialog (e.g. elementary/switchboard#245)
2. Wait a while

### Expected Behavior

Dialog and an application stay until user or application responds

### OS Version

7.x (Early Access)

### Software Version

Compiled from git

Marukesu:
That's the DBus timeout. This should be easily fixed by setting the call timeout to `Int.MAX`, vala support setting the timeout with the `[DBus (timeout = value)]` attribute, but it only accept literals, so we can't use the constant here.

PR #1515, opened 2023-01-18
Dialogs: Remove DBus Timeout From AccessDialog Call

Marukesu:
we don't want to calls to the portal to be timedout before the user sent a response, so disable the timeout.

while here, fix a issue with the show() call in the CloseDialog class creating new dialogs even when a dialog is already open, also handle the portal response only in the AccessDialog base class now and use a class callback to convert betwwen portal and close dialog responses.

fixes https://github.com/elementary/gala/issues/1505

### DIFF
--- a/src/Dialogs.vala
+++ b/src/Dialogs.vala
@@ -1,11 +1,12 @@
 /*
- * Copyright 2021 elementary, Inc. (https://elementary.io)
+ * Copyright 2021-2023 elementary, Inc. (https://elementary.io)
  * SPDX-License-Identifier: GPL-3.0-or-later
  */
 
 namespace Gala {
     [DBus (name = "org.freedesktop.impl.portal.Access")]
     public interface AccessPortal : Object {
+        [DBus (timeout = 2147483647)] // timeout = int.MAX; value got from <limits.h>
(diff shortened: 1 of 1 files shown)


src/Dialogs.vala
```diff
@@ -32,12 +37,13 @@ namespace Gala {
...
         public string deny_label { get; set; }
 
-        public signal void response (uint response);
-
         const string PANTHEON_PORTAL_NAME = "org.freedesktop.impl.portal.desktop.pantheon";
         const string FDO_PORTAL_PATH = "/org/freedesktop/portal/desktop";
         const string GALA_DIALOG_PATH = "/io/elementary/gala/dialog";
 
+        [CCode (delegate_target = false)]
+        protected class OnResponseCallback on_response;
```
tintou, line 45: Wouldn't it be more elegant/idiomatic to have a `protected virtual uint on_response (uint response)`

tintou:
Good catch!

src/Dialogs.vala
Marukesu: done!

tintou (approved):

PR #1516, opened 2023-01-18
WindowManager: `confirm_display_change()` Fixes

Marukesu:
the value of 30 seconds here is wrong, and the dialog was outliving the timeout used by mutter internally. instead of hardcoding the right values here too, use the mutter provided function to query the real timeout.

also, make `AccessDialog.show()` base handler run first than signal callbacks, and that we only call `AccesDialog.close()` when we don't receive a response.

### DIFF
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -2122,17 +2122,21 @@ namespace Gala {
 
         public override void confirm_display_change () {
             var timeout = Meta.MonitorManager.get_display_configuration_timeout ();
+            var summary = ngettext (
+                "Changes will automatically revert after %i second.",
+                "Changes will automatically revert after %i seconds.",
+                timeout
+            );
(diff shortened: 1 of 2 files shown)


src/WindowManager.vala
```diff
@@ -2121,24 +2121,34 @@ namespace Gala {
         }
 
         public override void confirm_display_change () {
+            var timeout = Meta.MonitorManager.get_display_configuration_timeout ();
+            uint dialog_timeout_id = 0;
+
             var dialog = new AccessDialog (
                 _("Keep new display settings?"),
-                _("Changes will automatically revert after 30 seconds."),
+                _("Changes will automatically revert after %i seconds.").printf (timeout),
```
tintou, line 2129: We now need to use ngettext for such string

tintou (approved):

PR #1529, opened 2023-02-06
Fix multitasking view animations

lenemter:
- Fix weird icon animation when opening multitasking view
- Animate maximized windows shadow opacity with gestures

### DIFF
--- a/src/Widgets/MultitaskingView.vala
+++ b/src/Widgets/MultitaskingView.vala
@@ -733,7 +733,7 @@ namespace Gala {
                     }
 
                     clone.save_easing_state ();
-                    clone.set_easing_mode (AnimationMode.EASE_OUT_QUAD);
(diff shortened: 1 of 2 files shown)


peteruithoven:
Could you clarify which shadow opacity animation you're referring to?

lenemter:
@peteruithoven This PR animates the shadow opacity when opening multitasking view with gestures. In `master` when opening/closing multitasking view with gestures shadow suddenly appears/disappears when you end the gesture.

peteruithoven:
I think I might have captured that shadow issue in the following capture? 
[Gala master shadow issue.webm](https://user-images.githubusercontent.com/523210/219873672-5f56bc8d-c5e5-49a6-a7ab-e08b95c21902.webm)

lenemter:
@peteruithoven No, that is different bug that has been fixed in #1531. There is an update planned forward the end of February, so you need to wait for the fix :)

danirabbit (approved):
Looks good, nice work!

PR #1533, opened 2023-02-10
WorkspaceInsertThumb: Code style fixes

lenemter:
Review with "Hide whitespace"

* Inline namespace
* Explicit Clutter and Meta namespaces
* Explicit `private` keyword
* Update copyright


danirabbit (approved):
Looks good, thanks!

PR #1540, opened 2023-02-17
Dialogs: Improve force quit string

davidmhewitt:
I don't think this makes grammatical sense currently, does this improve things?

### DIFF
--- a/src/Dialogs.vala
+++ b/src/Dialogs.vala
@@ -154,7 +154,7 @@ namespace Gala {
                 title = _("Application is not responding");
             }
 
-            body = _("You may choose to wait a short while for it to continue or force the application quit entirely.");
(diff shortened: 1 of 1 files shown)


src/Dialogs.vala
```diff
@@ -154,7 +154,7 @@ namespace Gala {
                 title = _("Application is not responding");
             }
 
-            body = _("You may choose to wait a short while for it to continue or force the application quit entirely.");
+            body = _("You may choose to wait a short while for it to continue or force the application to quit entirely.");
```
jeremypw, line 157: I think it better to use the specific noun first and "it" second.

-> "You may choose to wait a short while for the application to continue, or force it to quit entirely."

davidmhewitt: Agreed! Done

danirabbit (approved):
Much better. Thank you!

PR #1544, opened 2023-02-18
Remove some deprecations and fix compile warnings

lenemter:
ModuleFlags.BIND_LOCAL was deprecated in 0.46, so I set required Vala version to 0.46 too.


src/AccentColor/AccentColorManager.vala
```diff
@@ -151,7 +151,12 @@ public class Gala.AccentColorManager : Object {
             return null;
         }
 
-        return metadata.get_tag_string (TAG_ACCENT_COLOR);
+        try {
+            return metadata.try_get_tag_string (TAG_ACCENT_COLOR);
+        } catch (Error e) {
+            warning (e.message);
+            return null;
```
danirabbit, line 158: Instead of return null here should we return a fallback color?

lenemter: No, this function is used when getting accent color from wallpaper, it tries to get it from metadata and if it fails (returns null) update_accent_color selects accent color based on a picture itself.

lenemter:
I removed deprecated `get_pkgconfig_variable` and `meson.source_root` from meson files.

danirabbit (approved):
Let's go :tada:

PR #1554, opened 2023-02-20
IconGroup: Make close button respect animations key

lenemter:


danirabbit (changes requested):
LGTM can we get a line in AppData? :)

danirabbit (approved):
:tada:

PR #1557, opened 2023-02-21
Fix window maximize and unmaximize animations

lenemter:
Supersedes #1512 
Supersedes #1550 
Resolves #1185 
Resolves #1536

PR #1559, opened 2023-02-22
Cleanup after #1557

lenemter:
Fix compile warning and reuse window snapshot in maximize animation


danirabbit (approved):
Nice job! :rocket:

Issue #975, opened 2020-11-25
Middle click to kill workspace (in overview) should be optional


hanaral:
## Problem
When testing the gestures branch I've noticed how often I swipe up, lift my fingers off, and then place them back down on the touchpad. As you can probably guess, it is annoying to loose all the apps in a workspace because of a shortcut you didn't know about.

## Proposal
Just make it a toggle in the new gestures panel under 'mouse and touchpad' in settings

ibrahimrahman:
Making it a toggle is still a better solution than removing it altogether as suggested in #977, which would take away functionality from mouse users for whom gestures are of little use. I personally use the middle click in workspace overview a lot and hope that it would stay.

Something else that could be done is making mouse clicks register after the mouse button has been released. This way the app will not close unless you lift your fingers without moving them at all. 

IIRC, Windows, unlike elementary, doesn't take any action until the mouse button is released. On button press, it just gives a visual feedback relating to what was clicked. I think this is a much more pleasant approach as it gives an impression of something actually being pressed, while also giving users a chance to rectify a miss-click

danrabbit:
I haven't looked at the code yet, but yeah if it's activating on press instead of on release that's an issue

If that doesn't resolve it, we should probably make sure that we're only listening to mouse events and ignoring three finger taps

cassidyjames:
@danrabbit I don't know what the proper solution is here to ensure things don't conflict, but I personally use three-finger tap to close apps all the time; it would be super frustrating to lose that. I guess this is where I prefer four fingers for workspace gestures so there's less of a possibility of a conflict.

It might be non-ideal, but I wonder if we can only respond to a three-finger _physical_ click on touchpads, and just ignore the tap-to-click if we _have_ to ignore something. That would mean we retain the ability to three-finger close apps, but you just have to physically click down on a touchpad instead of accidentally tapping when trying to do a gesture.

hanaral:
>@danrabbit I don't know what the proper solution is here to ensure things don't conflict, but I personally use three-finger tap to close apps all the time; it would be super frustrating to lose that. I guess this is where I prefer four fingers for workspace gestures so there's less of a possibility of a conflict. ...

@cassidyjames my ThinkPads touchpad isn’t terribly accurate, so I some times kill my entire workspace by accident despite only having 4 finger gestures like you.

cassidyjames:
[...]

PR #1560, opened 2023-02-22
DragDropAction: Activate click signal on release

lenemter:
Mentioned here: https://github.com/elementary/gala/issues/975#issuecomment-752770338

Allows dragging with any mouse button , not only primary one
In multitasking view, clicking on workspace now closes the view on button release


danirabbit (approved):
I think this can probably be considered to close #975 tbh. I don't think we need to add a toggle there. I'm no longer able to reproduce OP's issue and even doing a lot of sloppy movements it's much harder here to accidentally trigger a close. Nice work!

PR #1561, opened 2023-02-22
WindowIcon: Cleanup

lenemter:


danirabbit (approved):
Love that, false by default and nothing ever sets it true lol

PR #1565, opened 2023-03-04
vapi: Sync with greeter

bobby285271:
https://github.com/elementary/greeter/pull/639#issuecomment-1454746972.

Should introduce no change other than file rename.

You may review the commits one by one so the diff won't look too horrible to you.

/cc @tintou


tintou (approved):

tintou:
thank you!

PR #1573, opened 2023-03-12
Added new menu entries

### LINKED ISSUE #1572: New desktop menu items
### Problem

It would be nice to add new items to the context menu of the desktop.

### Proposal

You can add a terminal startup and a system shutdown item.

### Prior Art (Optional)

_No response_

camellan:
Fixes #1572

Screenshot from 2023-03-12 19 04 53


camellan:
Fix #1572

lenemter (changes requested):
I don't think it is a good idea to show either of this two items on desktop. elementary OS is not focused on terminal usage, so I think it's bad to add Terminal to desktop menu. Secondly, adding shutdown option is quite silly as you can do it in 2 clicks using Wingpanel.

However I do agree that desktop menu could have more entries, but not Terminal and shutdown.

@danirabbit Do you have any thoughts on this?

---

For future, translation files are being update automatically, so remove them please. If you want to translate elementary OS please follow guide at https://docs.elementary.io/contributor-guide/localization/translations

danirabbit:
Yeah I agree, I don't think these items are a great fit for this menu

PR #1574, opened 2023-03-13
Daemon: Workaround broken window menu for some apps

### LINKED ISSUE #1347: Parity between right-clicking titlebars/headerbars on mouse and touchpad
### What Happened?

I have noticed with some apps, right-clicking the headerbar using my mouse doesn't bring up the context menu (Take Screenshot; Always on Top; etc.), while a two-finger tap on my laptop's touchpad does. This menu is often very useful, and it would be great to have more consistent behavior here.

Worth noting that all the Elementary apps I've tried have the desired behavior and bring up the menu no matter the method. Newer Gnome apps (GTK4) also seem to work. Some examples of apps that don't work properly for me:

* Mailspring
* Slack
* Visual Studio Code
* Alacritty
* SpeedCrunch
* Signal
* GIMP
* KeePassXC
* LibreOffice
* MuseScore

Touchpad double-tap does work to bring up the context menu in all of these cases. At the moment I have only tried one mouse, a Lenovo Thinkpad-branded wireless mouse with which I've never had any other issues.

### Steps to Reproduce

Using a mouse, right-click on the header bar of a non-native app
See no context menu appearing
Compare to the behavior when doing the same using a laptop touchpad

### Expected Behavior

Mouse right-clicking should produce the same behavior as touchpad double-tapping.

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

### LINKED ISSUE #1387: Window menu sluggish/inoperative for apps with flathub origin
### What Happened?

It was noticed certain apps did not seem to produce a window menu when the headerbar was right-clicked (sometimes a menu appeared after right-clicked many times).  Other windows produced a window menu immediately.  After some research the problem seems to be related to Flatpak apps with origin "flathub".  Native apps and Flatpak apps with origin "appcenter" are unaffected (at least the one I have installed).

### Steps to Reproduce

1. Install from AppCenter an app with origin "flathub"  (e.g. Aisleriot).
2. Start the app
3. Secondary-click on the headerbar
4. No menu appears
5. Repeatedly secondary-click on the headerbar
6. Eventually a menu appears

### Expected Behavior

The secondary-click behaviour should not depend on the origin of the Flatpak.

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

Hardware

lenemter:
Resolves #1347 
Resolves #1387 

Not the best solution but gets the job done.


zeebok:
Can you elaborate more on "best solution"? I am less familiar with Gala so this seems like just a solution. What do you think would be better?

lenemter:
> Can you elaborate more on "best solution"? I am less familiar with Gala so this seems like just a solution. What do you think would be better?

I really don't know how to fix this properly. Gtk.Menu.popup works weirdly here, that's why this workaround is made. I don't think there is even something like Gtk.Menu in GTK4. So when porting to GTK4 a new solution will have to be implemented.

We can do the same thing as Gnome Shell and draw menus using Clutter, but this will require a lot more work. If someone know better and simpler solution please let me know.

zeebok (approved):

PR #1582, opened 2023-03-23
Remove texture.png

lenemter:
In my opinion this texture looks really dirty and old (looks like it was created 9 years ago), so I really want to remove it and instead use the default background color RGB(46, 52, 54) (the one Greeter uses).

The ability to set custom image as a background for multitasking view via gsettings key is retained.

### Before
old

### After
new


danirabbit (approved):
Looks good!

1280px:
woah... this whole time I though it's just my dirty display :o

I feel it could've look better if a subtle noise was used for texture instead.

PR #1583, opened 2023-03-25
Ignore autorepeat for some keybindings

### LINKED ISSUE #913: Holding on keyboard shortcut to activate the action only once
Holding any shortcut is making the action to do many times until it's pressed instead it could be better if it can only do it once like in super+a when holding it the actions is repeated instead it would be nice if it can do it once that would make the animations complete otherwise it looks incomplete and broken.
Things like switching windows can stay the same as it would be useful to cycle windows with the key pressed the whole time
I am in elementary os 5.1.7

lenemter:
Resolves #913


danirabbit (approved):
This looks good! I wonder if it makes sense to allow repeating for screenshot? I suppose we’ll hear about it if that breaks someone’s workflow 😂

PR #1590, opened 2023-03-29
WindowSwitcher: Select a window on button release

lenemter:
Select a window on button release instead of button press.


danirabbit (approved):
👍

PR #1591, opened 2023-03-29
DragDropAction: Check that actor was actually clicked

lenemter:
In master `actor_clicked` is triggered even if you only release mouse button on actor without actually clicking it.

### DIFF
--- a/lib/DragDropAction.vala
+++ b/lib/DragDropAction.vala
@@ -214,7 +214,7 @@ namespace Gala {
                         actor.get_transformed_position (out x, out y);
 
                         // release has happened within bounds of actor
-                        if (x < ex && x + actor.width > ex && y < ey && y + actor.height > ey) {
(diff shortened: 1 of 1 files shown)


lib/DragDropAction.vala
```diff
@@ -214,7 +214,7 @@ namespace Gala {
                         actor.get_transformed_position (out x, out y);
 
                         // release has happened within bounds of actor
-                        if (x < ex && x + actor.width > ex && y < ey && y + actor.height > ey) {
+                        if (clicked && x < ex && x + actor.width > ex && y < ey && y + actor.height > ey) {
```
tintou, line 217: Do we also want to ungrab_actor in this case?

lenemter: I guess so? It's already there on line 221.

lenemter: Makes sense. Fixed.

davidmhewitt (approved):

PR #1594, opened 2023-03-29
Allow the notification stack to be moved using DBus

### LINKED ISSUE #91: Move notifications when indicators are open
Thought this had been filed, but couldn't find it anywhere. We should either defer or move notifications when WingPanel indicators are open since they occupy the same space and get in each others' way.

stan-janssen:
Adds a DBUS interface to move the notifications stack. This can be called by WingPanel to move the notifications stack out of the way when indicator menus are opened. This can be used to close #91.

Please see https://github.com/elementary/wingpanel/pull/481 for more details.

### DIFF
--- a/src/NotificationStack.vala
+++ b/src/NotificationStack.vala
@@ -22,7 +22,7 @@ public class Gala.NotificationStack : Object {
 
     // we need to keep a small offset to the top, because we clip the container to
     // its allocations and the close button would be off for the first notification
-    private const int TOP_OFFSET = 2;
+    private int TOP_OFFSET = 2;
     private const int ADDITIONAL_MARGIN = 12;
     private const int MARGIN = 12;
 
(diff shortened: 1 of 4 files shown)


src/NotificationStack.vala
```diff
@@ -22,7 +22,7 @@ public class Gala.NotificationStack : Object {
 
     // we need to keep a small offset to the top, because we clip the container to
     // its allocations and the close button would be off for the first notification
-    private const int TOP_OFFSET = 2;
```
lenemter, line 25: Since `TOP_OFFSET` is no longer a constant we should name it lowercase and move it away from other constants.

stan-janssen: I agree. I was thinking if I should use `TOP_OFFSET` as a `const` again, and use a second variable to control the offset. Because now I have the magic number 2 in two different places.

lenemter: Makes sense. That would be even better

lenemter (approved):

stan-janssen:
I just noticed that we need to account for the resolution scale when doing this. This is now fixed in a follow-up commit.

PR #1595, opened 2023-03-29
Prototype colorblind correction filters

### LINKED ISSUE #1091: Accessibility grayscale and colourblind filters
## Problem

No options currently to make screen grayscale or apply (colourblind or otherwise) colour filters.

## Proposal

Add a 'colour' pane to System Settings > Display. Used to manage both software filters and their strength, as well as display colour cal. in the future (there was a lot of talk on Wayland getting it built-in)

## Prior Art

image

lenemter:
Resolves #1091

Big thank you to https://github.com/G-dH/gnome-colorblind-filters for these beautiful shaders.

image

image

image

image

image


danirabbit:
This is awesome! Fantastic work!

Something I'm thinking about here is focus modes and if there's anything we need to do to plan ahead for that. For example, I know that Android offers monochrome as part of their "Wind Down" feature before bed. So, I wonder if there's anything we need to do here to plan ahead of in terms of a workflow that, for example, has a colorblindness correction shader during the day, wants to red shift that in the evening, and then go monochrome during "Sleep" time

lenemter:
@danirabbit This is mostly ready. I've just added transitions when changing the filter.
One more thing, seems like Windows in addition offers 'Inverted' and 'Grayscale inverted'. Is it worth adding them?

danirabbit:
@lenemter

Hmm that makes me wonder if saturation should be its own separate value that could be applied to any of the other filters? Not really sure tbh how flexible this needs to be. I’ll try to find some further guidance on this. I just pinged an a11y expert for feedback hopefully she has a moment 😅

lenemter:
@danirabbit I've already added a configurable strength for filters :)

danirabbit:
Sorry i meant like if it would be useful to later saturation on top of another filter. So for example layering monochrome with inverted or with red/green colorblind filtering

I was directed to this site, which I haven’t dug into yet but wanted to make sure I left it here for reference: https://wearecolorblind.com/resources/collection/resources-for-designing-for-the-colorblind/

tintou:
Wow, that looks really great!
I think that for even better maintainability I would suggest to add `data/shaders/colorblindness-correction.vert` and `data/shaders/monochrome.vert` and add the content of the shader inside (so that you can properly add the license data and get syntax coloration too) and use them as GResources

in /data/gala.gresource.xml:
```xml
<gresource prefix="/io/elementary/desktop/gala">
  <file compressed="true">shaders/colorblindness-correction.vert</file>
  <file compressed="true">shaders/monochrome.vert</file>
</gresource>
```
Then you can simply
```vala
try {
    var bytes = GLib.resources_lookup_data ("/io/elementary/desktop/gala/shaders/colorblindness-correction.vert", GLib.ResourceLookupFlags.NONE);
    set_shader_source (bytes.get_data ());
} catch (Error e) {
    critical ("Unable to load colorblindness-correction.vert: %s", e.message);
}
```

danirabbit:
My brother has protanopia and he said these are the best filters he’s ever seen :)

danirabbit:
@lenemter any reason this is still drafting?

lenemter:
@danirabbit Yes, currently transitions work pretty badly here because of the way I made them. I'll try to fix them on this weekends

danirabbit:
@lenemter I'm getting a consistent crash if I first enable the monochrome filter and then enable a color blindness filter

lenemter:
@danirabbit Fixed

danirabbit (approved):
Works perfectly! These are fantastic, thank you so much for working on this :tada:

PR #1596, opened 2023-03-29
Avoid to use Gdk constants when Clutter should be used

tintou:
Also avoid to use Gdk.Display when we have already everything from Mutter.

### DIFF
--- a/plugins/pip/PopupWindow.vala
+++ b/plugins/pip/PopupWindow.vala
@@ -55,10 +55,6 @@ public class Gala.Plugins.PIP.PopupWindow : Clutter.Actor {
             || window_type == Meta.WindowType.MODAL_DIALOG;
     }
 
-    private static void get_current_cursor_position (out int x, out int y) {
-        Gdk.Display.get_default ().get_default_seat ().get_pointer ().get_position (null, out x, out y);
-    }
-
     public PopupWindow (Gala.WindowManager wm, Meta.WindowActor window_actor) {
         Object (wm: wm, window_actor: window_actor);
     }
@@ -185,7 +181,7 @@ public class Gala.Plugins.PIP.PopupWindow : Clutter.Actor {
         resize_button.opacity = 255;
         resize_button.restore_easing_state ();
 
-        return Gdk.EVENT_PROPAGATE;
+        return Clutter.EVENT_PROPAGATE;
     }
 
     public override bool leave_event (Clutter.CrossingEvent event) {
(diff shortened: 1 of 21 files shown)


plugins/pip/PopupWindow.vala
```diff
@@ -217,40 +213,42 @@ public class Gala.Plugins.PIP.PopupWindow : Clutter.Actor {
...
-            return Gdk.EVENT_STOP;
+            return Clutter.EVENT_STOP;
         }
 
         resizing = true;
 
-        get_current_cursor_position (out resize_start_x, out resize_start_y);
+
+        Graphene.Point coords;
+        Clutter.ModifierType mods;
```
lenemter, line 223: These two variables are unused.

src/Widgets/WindowSwitcher.vala
lenemter, line 1: Window switcher sometimes doesn't automatically close in this branch.

lenemter (changes requested):

src/Widgets/WindowSwitcher.vala
tintou: Yeah it needs more testing, I went too far with the removal of the use of Gdk but I was really sad to see that we were using it to query pointer position (!!) it means that it does X11 queries to get it instead of simply get it via Mutter.

src/Zoom.vala
```diff
@@ -99,6 +99,20 @@ public class Gala.Zoom : Object {
         gesture_tracker.connect_handlers (null, (owned) on_animation_update, null);
     }
 
+    private inline Graphene.Point compute_new_pivot_point () {
+        unowned var wins = wm.ui_group;
+        unowned Meta.CursorTracker cursor_tracker = wm.get_display ().get_cursor_tracker ();
+        Graphene.Point coords;
+        Clutter.ModifierType mods;
+        cursor_tracker.get_pointer (out coords, out mods);
```
lenemter, line 107:
```suggestion
        cursor_tracker.get_pointer (out coords, null);
```

tintou: After looking at Mutter source code, that's indeed possible, opened https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/2951

src/Widgets/WindowSwitcher.vala
tintou: fixed !

src/Zoom.vala
```diff
@@ -99,6 +99,18 @@ public class Gala.Zoom : Object {
         gesture_tracker.connect_handlers (null, (owned) on_animation_update, null);
     }
 
+    private inline Graphene.Point compute_new_pivot_point () {
+        unowned var wins = wm.ui_group;
+        Graphene.Point coords;
+        wm.get_display ().get_cursor_tracker ().get_pointer (out coords, null);
+        var new_pivot = new Graphene.Point () {
```
lenemter, line 106:
```suggestion
        var new_pivot = Graphene.Point () {
```

lenemter (approved):
Works the same as in master. The code looks good. Great job! :tada:

Issue #1136, opened 2021-05-14
Low fps overview animation


lilvadim:
Multitasking animations have noticeable FPS drops in the beginning and in the end of gesture (when windows are bigger).
Especially with multiple Chrome windows.

It's about lags that I see when using 1:1 touchpad gestures.
_Specs: 
AMD Ryzen 5 3500U with mobile Vega 8 graphics (laptop).
FullHD screen, Large text selected._

lilvadim:
Maybe it's possible to somehow optimize performance of windows scaling animations? Then your DE will be greater than GNOME, because GNOME 40 has huge laggy animations problems when it comes to multitasking view with more than zero windows opened.

Condor157:
Hi. Have the same problem. Animations in the end are shuttering. This is a bit sad. I am a shop assistant and have tested it on several new laptops with AMD 4700U, AMD 3500U, Intel 10300H and others. But even on my Intel® Celeron® G4920 there is the same problem. It would be great to fix that. I can add photos and videos if it is needed. I am always open to help improoving this great OS.

okitoxo:
I have the same problem with a graphic card RX 570. FPS have low when I open multitask or close it only

okitoxo:
This is a video recorded with slow motion

https://user-images.githubusercontent.com/5513927/129086201-34210fa7-3de1-4ef7-9054-163897c4bcfc.mp4

danirabbit:
@lenemter should this be closed?

lenemter:
@danirabbit Nope, multitasking view still lags for me even after latest optimizations.

lenemter:
I'll close this because #1603 improved performance a lot and multitasking view animation no longer lags for me.

PR #1603, opened 2023-04-01
WindowClone: optimize ActiveShape

lenemter:
Addresses #1136

ActiveShape.allocate  while animating:
0.002864 -> 0.000057 seconds (~50 times faster)


davidmhewitt (dismissed):
Nice work! :tada: 

No regressions found

davidmhewitt (approved):
I thought I'd found an issue where the currently focused window isn't highlighted when opening the multitasking view, but this also doesn't work properly in `master` on my machine.

So, this is fine as it is.

PR #1611, opened 2023-04-02
Don't manually map notifications while nudge animation is playing

lenemter:
Fixes this:
image


danirabbit:
Hm, I was still able to reproduce that issue with this branch, but admittedly its very hard to reproduce :sweat_smile:

danirabbit (approved):
It's much easier to produce in master than in this branch so :+1:

Issue #1287, opened 2021-11-03
Visual artifacts when swiping workspaces with touchpad


kdwk:
### What Happened?

https://user-images.githubusercontent.com/39268094/140052282-16d69df0-40fd-43f6-873e-1994a2d24900.mp4

As can be seen, gray bars sometimes appear when three-finger swiping to the left when there is no more workspace to the left

### Steps to Reproduce

1. Go to the leftmost workspace
2. Open an app like Web
3. Swipe with three fingers on a touchpad to the left, such that a rebound animation is shown
4. Repeat 1-2 until a gray bar appears, obstructing the content

### Expected Behavior

No gray bar appears to obstruct content on the workspace

### OS Version

6.x (Odin)

### Software Version

Latest release (I have run all updates)

### Hardware Info

CPU: Intel Core i7-9700
GPU: Intel UHD 630
X11

PR #1612, opened 2023-04-02
Use translation-x for nudge animation

lenemter:
Addresses #1287

`translation-x` seem to produce a lot less glitches than `x`


danirabbit (approved):
Works for me :rocket:

PR #1630, opened 2023-04-07
Support floating point scale for resize button

davidmhewitt:


lib/Utils.vala
```diff
@@ -379,11 +379,16 @@ namespace Gala {
...
-            var height = 36 * Utils.get_ui_scaling_factor ();
-            if (resize_pixbuf == null || resize_pixbuf.height != height) {
+        public static Gdk.Pixbuf? get_resize_button_pixbuf (float scale) {
+            var height = scale_to_int (36, scale);
+
+            if (resize_pixbufs == null) {
+                resize_pixbufs = new Gee.HashMap<int, Gdk.Pixbuf?> ();
+            }
+
+            if (resize_pixbufs[height] == null) {
```
tintou, line 389: I guess that this would be safer (also forgotten with the close button)
```suggestion
            if (!(height in resize_pixbufs)) {
```

tintou (approved):

lib/Utils.vala
tintou: We can revisit it afterward

PR #1638, opened 2023-04-09
WindowSwitcher: Fix rounded corners

### LINKED ISSUE #1355: Issue with rounded corners of window switcher
### What Happened?

Just installed the latest updates and noticed that there’s a regression on the WindowSwitcher when drawing the rounded corners (There is a faint square border around it and the space between the border and the rounded corners is transparent.)

I can’t take a screen capture of it because X11 was released in 1984 :)

Also the colour seems different and/or there is less shadow (?) – I don’t know what it is but it blends into the other windows instead of standing out.

### Steps to Reproduce

1. Open a few apps
2. Make sure you have a light wallpaper in dark mode so you can see the issue
3. Use the WindowSwitcher.

### Expected Behavior

  - There shouldn’t be a square border around it.
  - It should stand out from the other windows instead of blending into them.

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

lenemter:
Resolves #1355

This PR does 2 things:
- Increases border radius of the shadow
- Cleans the canvas before drawing background

### DIFF
--- a/data/gala.appdata.xml.in
+++ b/data/gala.appdata.xml.in
@@ -21,6 +21,7 @@
       </description>
       <issues>
         <issue url="https://github.com/elementary/gala/issues/536">Enable "Alt + ~" to switch between windows of the same application</issue>
+        <issue url="https://github.com/elementary/gala/issues/1355">Issue with rounded corners of window switcher</issue>
       </issues>
     </release>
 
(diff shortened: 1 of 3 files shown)


danirabbit (approved):
Nice fix :rocket:

PR #1653, opened 2023-04-12
IconGroup: Improve animation

lenemter:
* Animate icon group highlights with gestures
* Run animation as soon as the gesture stops, not when the workspace is actually switched.


danirabbit (approved):
:fire:

PR #1670, opened 2023-05-02
Fix double icon when crossing icon group

lenemter:
Fixes this:
image


danirabbit:
@lenemter could you clarify what the issue is and steps to reproduce? I'm unsure what I should be testing for here

lenemter:
@danirabbit When you drop a window in an icon group two icons of the same window appear for a brief moment

danirabbit (approved):
Yup, confirmed!

Issue #1674, opened 2023-05-17
Get accent from settings portal


danirabbit:
### What Happened?

Currently we get the accent color from the stylesheet, which means if the stylesheet gets changed from one of the elementary ones, our accent color stuff breaks

### Steps to Reproduce

1. Change the stylesheet to something like Adwaita
2. See that highlighting in the alt tab window switcher is broken

### Expected Behavior

There's now an accent color on the settings portal, so that seems the cleaner way to get the accent color. See: https://github.com/elementary/settings-daemon/pull/60

### OS Version

7.x (Horus)

### Software Version

Latest release (I have run all updates)

PR #1682, opened 2023-06-13
Wayland fixes

### LINKED ISSUE #970: ShadowEffect class causes hangs/crashes on Wayland
Opening this for tracking against the Wayland project board.

Because the `ShadowEffect` class uses GTK CSS, it tries to initialise GTK upon construction.

It is not possible to initialise GTK inside the Wayland compositor, so we need to go back to using Cogl/Clutter to draw shadows.

lenemter:
Closes #970

Basically makes Wayland session launchable.

We can get around issues with Gtk.StyleContext and Gtk.WidgetPath if we create them a bit later, e.g. create window switcher's Gtk objects when user opens window switcher for the first time, or just create them in Clutter.Actor.paint :)

Since there is #1674 we might want to hardcode the theme colors, which is done here.


tintou:
Wow, great, can you break it in meaningful commits, I'd prefer this to not being squashed-and-merged :wink:

PR #1683, opened 2023-06-13
Hardcode accent colors

lenemter:


danirabbit:
Instead of hardcoding accent colors, can we get the RGB value directly over the settings portal?

Screenshot from 2023-06-13 11 36 35

lenemter:
@danirabbit I think we can get only color of the currently applied accent color, not all of them.

danirabbit:
Oh okay, I see what this is about. This is for setting the accent color based on the wallpaper. Sorry for the confusion! Makes sense

danirabbit (approved):
Works as expected. Nice job! :)

PR #1686, opened 2023-06-13
WindowSwitcher: Use Gtk later

lenemter:


src/Widgets/WindowSwitcher/WindowSwitcher.vala
```diff
@@ -119,8 +126,8 @@ namespace Gala {
             style_context.add_class ("unified");
```
tintou, line 126: You forgot to remove this here

tintou (approved):

PR #1689, opened 2023-06-13
ShadowEffect: Only use Gtk widgetry at the pain stage

tintou:
Avoid using Gtk at creation as the current display might not be existing.


tintou:
@lenemter What do you say about this instead of deferring ShadowEffect creation?

danirabbit:
Fixes #970 ?

lenemter (commented):
Works perfectly and is much cleaner. But we need to copy this to plugins/pip/ShadowEffect.vala as well (We need to deduplicate this class, but it's out of scope of this PR)

tintou:
> Fixes #970 ?

Maybe :shrug:

lenemter (approved):

PR #1709, opened 2023-07-04
Use dark wallpaper

### LINKED ISSUE #1704: Use dark wallpaper in dark mode when available
### Problem

We currently only support a single wallpaper, but in dark mode folks might want to use a darker wallpaper that is better for dark environments, easier on the eyes, less contrasty, etc.

### Proposal

There is a `picture-uri-dark` key in `org.gnome.desktop.background`. We should use that (if it isn't empty) when the system is in dark mode.

Currently use use `picture-uri` here:
https://github.com/elementary/gala/blob/dd98aefaa09a697519ad4dd051eeaecc19780dcd/src/Background/BackgroundSource.vala#L25
https://github.com/elementary/gala/blob/dd98aefaa09a697519ad4dd051eeaecc19780dcd/src/AccentColor/AccentColorManager.vala#L75

### Prior Art (Optional)

_No response_

lenemter:
Closes #1704

### DIFF
--- a/src/Background/BackgroundSource.vala
+++ b/src/Background/BackgroundSource.vala
@@ -23,6 +23,7 @@ namespace Gala {
             "picture-opacity",
             "picture-options",
             "picture-uri",
+            "picture-uri-dark",
             "primary-color",
             "secondary-color"
         };
@@ -67,6 +68,9 @@ namespace Gala {
                     }
                 }
             });
+
+            unowned var granite_settings = Granite.Settings.get_default ();
+            granite_settings.notify["prefers-color-scheme"].connect (() => changed ());
         }
 
         private void monitors_changed () {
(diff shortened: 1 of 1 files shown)


src/Background/BackgroundSource.vala
```diff
@@ -91,7 +95,10 @@ namespace Gala {
 
             var style = settings.get_enum ("picture-options");
             if (style != GDesktop.BackgroundStyle.NONE) {
-                var uri = settings.get_string ("picture-uri");
+                var uri = settings.get_string (
+                    Granite.Settings.get_default ().prefers_color_scheme == DARK ?
+                    "picture-uri-dark" : "picture-uri"
+                );
```
tintou, line 101: We need to fallback to picture-uri if there is not file in picture-uri-dark

tintou:
Seeing this made me do this https://github.com/elementary/gala/pull/1711

danirabbit (approved):
Does the thing. Nice job!

PR #1710, opened 2023-07-04
Keep track of dimmed windows

### LINKED ISSUE #1331: Dialog dim won't go off on gtk app
### What Happened?

After the November 2021 update I notice a strange behavior with the dialogs in some gtk apps, specifically with "double commander", when the dialog closes, the dim won't go off.

I tried with version 0.9.1 and 1.0.1 of "double commander".

And if you call multiple times the dialog and then cancel, the dim just get over and over till you see just a black area.

Didn't happen this before the November update.

dialog1
dialog2
dialog3
dialog4
dialog5

### Steps to Reproduce

1. run doublecmd
2. press F7 to make dir
3. cancel
4. repeat steps 2 and 3 many times

### Expected Behavior

the dim should disappear when the dialog closes

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

dell e7240 (intel gfx)

lenemter:
Fixes #1331

That's an ugly solution but I guess it's the only working one. For some reason when double commander destroys its dialogs they no longer have transient window, because of that the undim failed.

### DIFF
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -795,12 +795,9 @@ namespace Gala {
             return (proxy in modal_stack);
         }
 
-        struct DimData {
-            Meta.Window child;
-            unowned Meta.WindowActor parent_actor;
-        }
+        private HashTable<Meta.Window, unowned Meta.WindowActor> dim_table =
+            new HashTable<Meta.Window, unowned Meta.WindowActor> (GLib.direct_hash, GLib.direct_equal);
 
-        private List<DimData?> dim_data = new List<DimData?> ();
         private void dim_parent_window (Meta.Window window, bool dim) {
             unowned var transient = window.get_transient_for ();
             if (transient != null && transient != window) {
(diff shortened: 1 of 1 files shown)


src/WindowManager.vala
```diff
@@ -795,22 +795,61 @@ namespace Gala {
             return (proxy in modal_stack);
         }
 
+        struct DimData {
+            Meta.Window child;
+            unowned Meta.WindowActor parent_actor;
+        }
+
+        private List<DimData?> dim_data = new List<DimData?> ();
```
tintou, line 803: What about using a `GLib.HashTable<Meta.Window child, unowned Meta.WindowActor>` ?

lenemter: I'm have no experience with `HashTable` in Vala,  what hash and equal functions should it use?

tintou: GLib.direct_hash and GLib.direct_equal for pointer-based types :)

lenemter: Thanks!

src/WindowManager.vala
```diff
@@ -795,21 +795,35 @@ namespace Gala {
...
+
+                return;
+            }
+
+            // fall back to dim_data (see https://github.com/elementary/gala/issues/1331)
+            if (window in dim_table && dim_table[window].get_effect ("dim-parent") != null) {
+                dim_table[window].remove_effect_by_name ("dim-parent");
+                dim_table.remove (window);
+                debug ("Removed dim using dim_data");
             }
```
tintou, line 827: Cache the value so that we don't have to do the table lookup twice
```suggestion
            unowned Meta.WindowActor? transient_actor = dim_table.take (window);
            if (transient_actor != null) {
                transient_actor.remove_effect_by_name ("dim-parent");
                debug ("Removed dim using dim_data");
            }
```

tintou (approved):

tintou:
That's an ugly solution but I guess it's the only working one. For some reason when double commander destroys its dialogs they no longer have transient window, because of that the undim failed.

PR #1711, opened 2023-07-04
Background: Use GLib.File instead of dealing with Paths

tintou:
Simplify the code and avoids to deal with File paths directly.

### DIFF
--- a/src/Background/Animation.vala
+++ b/src/Background/Animation.vala
@@ -17,20 +17,20 @@
 
 namespace Gala {
     public class Animation : Object {
-        public string filename { get; construct; }
-        public string[] key_frame_files { get; private set; default = {}; }
+        public GLib.File file { get; construct; }
+        public GLib.GenericArray<GLib.File> key_frame_files { get; private set; default = new GLib.GenericArray<GLib.File> (); }
         public double transition_progress { get; private set; default = 0.0; }
         public double transition_duration { get; private set; default = 0.0; }
         public bool loaded { get; private set; default = false; }
 
         private Gnome.BGSlideShow? show = null;
 
-        public Animation (string filename) {
-            Object (filename: filename);
+        public Animation (GLib.File file) {
+            Object (file: file);
         }
 
         public async void load () {
-            show = new Gnome.BGSlideShow (filename);
+            show = new Gnome.BGSlideShow (file.get_path ());
 
             show.load_async (null, (obj, res) => {
                 loaded = true;
@@ -42,7 +42,7 @@ namespace Gala {
         }
 
         public void update (Meta.Rectangle monitor) {
-            string[] key_frame_files = {};
+            var new_key_frame_files = new GLib.GenericArray<GLib.File> ();
 
             if (show == null)
                 return;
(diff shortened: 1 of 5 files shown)


lenemter (commented):
Can we omit the `GLib` namespace? It makes some lines pretty long.

tintou:
I personally really prefer having the GLib namespace specified (when it's not a native type like string/int/bool) so that we know that it is not an object from the project but from GLib

src/Background/Animation.vala
```diff
@@ -17,20 +17,20 @@
 
 namespace Gala {
     public class Animation : Object {
-        public string filename { get; construct; }
-        public string[] key_frame_files { get; private set; default = {}; }
+        public GLib.File file { get; construct; }
+        public GLib.GenericArray<GLib.File> key_frame_files { get; private set; default = new GLib.GenericArray<GLib.File> (); }
```
Marukesu, line 21: Does this need to be a GenericArray? given that we known that the max number of files we will use is 2, isn't a fixed array enough.

src/Background/Animation.vala
```diff
@@ -52,19 +52,21 @@ namespace Gala {
...
+            }
 
-            if (file2 != null)
-                key_frame_files += file2;
+            if (file2 != null) {
+                new_key_frame_files.add (GLib.File.new_for_path (file2));
+            }
 
-            this.key_frame_files = key_frame_files;
+            this.key_frame_files = new_key_frame_files;
```
Marukesu, line 69: i don't think the `this` accessor is needed anymore here.

src/Background/Background.vala
```diff
@@ -28,39 +28,39 @@ namespace Gala {
...
-                    filename: filename);
+                    file: file);
         }
 
         construct {
             background = new Meta.Background (display);
             background.set_data<unowned Background> ("delegate", this);
 
-            file_watches = new Gee.HashMap<string,ulong> ();
+            file_watches = new Gee.HashMap<GLib.File, ulong> ();
```
Marukesu, line 55: Use `GLib.File.hash` and `GLib.File.equal` here? Gee will fallback to `direct_{hash,equal}` if the functions aren't provided here.

src/Background/Background.vala
```diff
@@ -121,18 +121,18 @@ namespace Gala {
             }
         }
 
-        private void watch_file (string filename) {
-            if (file_watches.has_key (filename))
+        private void watch_file (GLib.File file) {
+            if (file_watches.has_key (file))
                 return;
```
Marukesu, line 126: Code style.

```suggestion
            if (file_watches.has_key (file)) {
                return;
            }
```

src/Background/Background.vala
```diff
@@ -242,20 +242,22 @@ namespace Gala {
             }
         }
 
-        private void load_file (string filename) {
-            if (filename.has_suffix (".xml"))
-                load_animation.begin (filename);
-            else
-                load_image (filename);
+        private inline void load_file (GLib.File file) {
+            if (file.get_basename ().has_suffix (".xml")) {
```
Marukesu, line 246: i think we should use `GLib.ContentType.guess()` here, so that we can identify xml files without a extension. like the ones that the Gtk portal provides.

src/Background/BackgroundSource.vala
```diff
@@ -87,26 +87,26 @@ namespace Gala {
...
-                    filename = uri;
+                    file = File.new_for_path (uri);
             }
 
             // Animated backgrounds are (potentially) per-monitor, since
             // they can have variants that depend on the aspect ratio and
             // size of the monitor; for other backgrounds we can use the
             // same background object for all monitors.
-            if (filename == null || !filename.has_suffix (".xml"))
+            if (file == null || !file.get_basename ().has_suffix (".xml"))
```
Marukesu, line 105: Same thing here, should check the mimetype instead of extension.

PR #1715, opened 2023-07-05
Block GSD OSD requests with `level` set to -1

### LINKED ISSUE #1714: Disabling touchpad notification/indicator includes empty volume bar
## What Happened

When you disable the touchpad via keyboard shortcut (fn+f9 on mine) — a notification-like indicator that appears has an empty volume bar like if I'm changing the sound. The icon is for the touchpad, tho.

Screenshot from 2021-09-05 09 01 34

## Expected Behavior

No volume bar, maybe textual explanation.

## Steps to Reproduce

1. Locate the keyboard button that disables the touchpad, somewhere in the "f" row. It should have a touchpad-like icon.
2. Press it, maybe while holding the "fn" key (depends on the keyboard).
3. See notification in the upper-right corner.

## Platform Information

Screenshot from 2021-09-05 09 08 00

lenemter:
Closes #1714

### DIFF
--- a/src/DBusAccelerator.vala
+++ b/src/DBusAccelerator.vala
@@ -162,5 +162,5 @@ namespace Gala {
 
             MediaFeedback.send (icon, level);
         }
-    } 
+    }
 }
--- a/src/DBusAccelerator.vala
+++ b/src/DBusAccelerator.vala
@@ -154,7 +154,7 @@ namespace Gala {
             string label = "";
             if (parameters.contains ("label"))
                 label = parameters["label"].get_string ();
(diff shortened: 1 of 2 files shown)


Marukesu:
I don't think this is right, we want theses. We should fix Confirmation notifications in the Daemon.

lenemter:
> We should fix Confirmation notifications in the Daemon.

@marukesu Do you mean we should patch gnome-settings-daemon?

Marukesu:
i mean, we should make our notification daemon know how to deal with icon-only confirmations.

PR #1717, opened 2023-07-07
Fix warnings

lenemter:
Fixes warnings spam from WindowClone introduced in https://github.com/elementary/gala/commit/14767bfde76e2d141b5ba88e0de7efe3496b9d56


danirabbit (changes requested):
I get like massive glitches and issues with this branch :grimacing: Multitasking view is really messed up with icon placement and no close buttons and wingpanel indicators won't open

lenemter:
@danirabbit Can you try to test this again. I sometimes randomly encounter such issues while developing. Maybe that's the case because I don't have any glitches on my side.

danirabbit (approved):
Yep, weird. After trying this again everything works as expected!

PR #1725, opened 2023-07-12
WindowManager: A small cleanup

lenemter:
1. Remove unused constant in MultitaskingView
2. Make WindowManagerGala.NUDGE_GAP integer
3. Add missing `unowned`
4. Move `switch_workspace`'s `window_created` listener to the method itself.


danirabbit (approved):
Looks good. afaict behavior is the same as in master. Nice work!

Issue #1004, opened 2020-12-28
Wallpaper in workspace clone and wallpaper clone should have the same style


danirabbit:
## What Happened

If you go into multitasking view, you can see that wallpapers have a shadow and inner highlight, but when swiping workspaces there is no box shadow or highlight

## Expected Behavior

There should be consistency between the two and they should be styled the same

## Steps to Reproduce

1. Push into multitasking view and notice that wallpapers have a box shadow
2. Stop half way through a swipe and notice wallpapers do not have a box shadow

PR #1726, opened 2023-07-12
Add shadow to wallpaper clones while switching workspace

lenemter:
Addresses #1004

### DIFF
--- a/data/gala.css
+++ b/data/gala.css
@@ -46,7 +46,6 @@
 }
 
 .workspace.decoration {
-    border-radius: 4px;
     box-shadow:
         0 0 0 1px alpha(#000, 0.2),
         0 8px 10px 1px alpha(#000, 0.14),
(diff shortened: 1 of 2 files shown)


zeebok:
I went to test this in a VM and the code can't be grabbed from that branch? When I have the chance to get the VM an ssh key setup I will try to just clone and checkout the branch.

vjr:
> I went to test this in a VM and the code can't be grabbed from that branch? When I have the chance to get the VM an ssh key setup I will try to just clone and checkout the branch.

You can also just git clone the http(s) link to the repo so no ssh keys required? Then git checkout the this branch from there?

About your question for downloading ZIP for the PR/branch, I wonder if you just download the ZIP for the repo/main branch and will it have the .git folders so you can git checkout to switch branches that way too.

danirabbit (approved):
Works as expected. Nice job!

PR #1730, opened 2023-07-19
Background: Cleanup

lenemter:
- Don't use custom hacky way to set dimmed wallpaper
- Handle every settings change in BackgroundSource
- Remove background_sources HashMap with only one item and everything that used it

### DIFF
--- a/src/Background/BackgroundSource.vala
+++ b/src/Background/BackgroundSource.vala
@@ -17,57 +17,36 @@
 
 namespace Gala {
     public class BackgroundSource : Object {
-        // list of keys that are actually relevant for us
-        private const string[] OPTIONS = {
-            "color-shading-type",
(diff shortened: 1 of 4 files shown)


src/Background/BackgroundSource.vala
```diff
@@ -17,57 +17,36 @@
...
-            "picture-uri-dark",
-            "primary-color",
-            "secondary-color"
-        };
-
         public signal void changed ();
 
         public Meta.Display display { get; construct; }
-        public Settings settings { get; construct; }
+        public GLib.Settings gnome_background_settings;
```
danirabbit, line 23: should this be private set?

danirabbit (changes requested):
afaict background dimming is no longer working in this branch

PR #1734, opened 2023-07-23
Round corners of workspaces in multitasking view

### LINKED ISSUE #1545: Draw rounded corners on workspaces in multitasking view
### Problem

The corners of workspaces in the multitasking view are perfectly square and sharp even though we mask off corners of the desktop

### Proposal

Round the corners of workspaces in multitasking view with the same radius as we use for the whole desktop

### Prior Art (Optional)

GNOME Shell has rounded corners on its workspace previews these days

lenemter:
Closes #1545


danirabbit:
If you zoom in a bunch you can see something is drawing a square here

Screenshot from 2023-10-30 20 06 12

lenemter:
@danirabbit I increased outline border radius, does it look better on your screen?

danirabbit:
@lenemter I think what's actually happening is that in the cairo rounded rectangle function the corners are drawn offset by like a half pixel which gives that weird aliasing. Gonna try to see if I can figure out how to fix it

danirabbit:
@lenemter okay I don't know enough about cairo to fix this lol. But I think just making it 1px larger is what it should be and not 3px larger

danirabbit:
Actually it should be smaller right? Like it doesn't make sense lol something is weird here
border-radius

danirabbit (approved):
We can fiddle with values later if we need to, rather not let perfect be the enemy of good :)

PR #1735, opened 2023-07-25
Kill workspace switch when opening multitasking view

### LINKED ISSUE #1728: invisible windows in multitasking view after workspace switch, if done quickly
### What Happened?

When I switch workspace and then open multitasking view in quick succession, the window previews in the view are invisible, with only outlines and icons showing where the window previews should be. This happens when I am in hurry, most people will probably not run into it, so it's rather a corner case.

Here is how the issue looks in a screenshot:

Snímek z 2023-06-12 13 08 27

Typically, I switch to another workspace to access some app, which can be hidden behind another window, so I open multitasking view to get to the app from there. It doesn't matter if I do this with touchpad gestures or keyboard shortcuts, I can 100% reproduce it with both whenever I want.

**The key to reproduce the issue is doing things quickly** - in particular, you need to start the second action before the first action finishes. It's very easy to see just by quickly pressing `Super+Tab` quickly followed by `Super+S`, or with a horizontal 3-finger swipe quickly followed by 3-finger swipe up.

### Steps to Reproduce

1. Have at least two workspaces with some apps open.
2. Switch between workspaces, using either a touchpad gesture (horizontal 3-finger swipe) or a keyboard shortcut, such as `Super+Tab`
3. Before the switch finishes, open multitasking view, with 3-finger swipe up, or by pressing `Super+S`
4. See the bug

### Expected Behavior

The window previews should be visible.

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

Slimbook Essential 14.

lenemter:
Fixes #1728

### DIFF
--- a/src/Widgets/MultitaskingView.vala
+++ b/src/Widgets/MultitaskingView.vala
@@ -652,6 +652,11 @@ namespace Gala {
             opened = !opened;
             var opening = opened;
 
+            // https://github.com/elementary/gala/issues/1728
+            if (opening) {
+                wm.kill_switch_workspace ();
+            }
+
             foreach (var container in window_containers_monitors) {
                 if (opening) {
                     container.visible = true;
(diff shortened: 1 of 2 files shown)


zeebok (approved):
Code seems reasonable and testing looks good

PR #1736, opened 2023-07-25
Background: Remove `background_sources` HashMap

lenemter:
This HashMap always contains only one item, so we can safely remove it.


lenemter:
Closing in favor of #1759.

PR #1741, opened 2023-08-11
MaskCorners: get radius from Stylesheet

danirabbit:
Instead of hardcoding, get this from the stylesheet


tintou:
This won't be possible with GTK4

PR #1746, opened 2023-08-29
Add support for Mutter 45

### LINKED ISSUE #1743: Support mutter 45 / libmutter-13
<copy-pasted-from-#1569>

Every mutter development cycle the API version bumps, so here is the biannual issue :see_no_evil: 

[mutter 45.beta.1](https://gitlab.gnome.org/GNOME/mutter/-/tags/45.beta.1) is out, it will be nice to bring support for mutter 45 / libmutter-13 so this can still be built on distributions that cannot pin mutter to old versions (e.g. Fedora and Arch).

We started [packaging GNOME 45.beta](//github.com/NixOS/nixpkgs/pull/247766) on NixOS, though we are [not affected](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/desktops/gnome/core/mutter/43/default.nix) but I am happy to help with testing when needed :-)

</copy-pasted-from-#1569>

---

Actually we are affected this time, https://github.com/GNOME/gsettings-desktop-schemas/commit/9726c5e92dbb65d410033d478e94ce900438884c removes a `toggle-shaded` key, which will cause crash to mutter 43 (which is what we are pinning), I am aware of that https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/2884  is in mutter 44 and will see if we should backport that MR or we just move to mutter 44 (or 45?)

Edit: Thanks to the work from budgie we are now applying https://github.com/BuddiesOfBudgie/magpie/commit/4177c466375462ca8ed8fdb60913df4422f19144 on our mutter 43. So as of GNOME 45.beta I don't aware of major issues on Pantheon from our side right now.

tintou:
Closes #1743


tintou:
Please note that I've only tested that it builds, not that it runs

bobby285271:
Seems to start for me

<details><summary>Screenshot</summary>




</details>

PR #1748, opened 2023-08-29
NotificationStack: Check if notification was already shown

lenemter:
After a reboot i cannot reproduce https://github.com/elementary/notifications/issues/218 anymore, so this is my first wild guess on what might be going on.


lenemter:
@TomiOhl Can you check if it makes any difference on your machine please? I can give instructions if you need them.

src/NotificationStack.vala
```diff
@@ -49,6 +49,10 @@ public class Gala.NotificationStack : Object {
     }
 
     public void show_notification (Meta.WindowActor notification, bool animate) {
+        if (notifications.contains (notification)) {
+            return;
+        }
```
Marukesu, line 54: not really a review, but i believe you could instead do:

```vala
  public void show_notification (Meta.WindowActor notification, bool animate) requires (!notifications.contains (notification)) {
```

and then take a look in the logs to see how often this is triggered, or if it's happens when the issue occurs.

TomiOhl:
> @TomiOhl Can you check if it makes any difference on your machine please? I can give instructions if you need them.

I've also rebooted and currently I cannot reproduce it either 😅 Not sure what triggers the behavior. Would be useful to know in order to test properly.
Instructions: the ones in the readme are up-to-date?

lenemter:
> Instructions: the ones in the readme are up-to-date?

Yes, they are. You can quickly install build dependencies using `sudo apt build-dep gala`

TomiOhl:
So I've tried to reproduce the issue on my other laptop, but I couldn't do so neither on this branch nor on master. Would be useful to know what causes it really.
Things that I haven't tried here but do cause other issues in gala:
- multiple displays
- different aspect ratios
I did get the issue again on my work laptop today though. I don't really want to use it for testing but right now I feel like that would be the easiest thing that I could do on my end

danirabbit (approved):

PR #1751, opened 2023-09-09
Remember window position

### LINKED ISSUE #297: Add an API to remember window position
It sounds like there is an intent in Gtk4 to remove functions related to setting window position. To prevent regression, we need to create some kind of API for applications to request that Gala remember their window position.

I think if we're smart we'll do this using relative units like percentage instead of absolute units so that we can preserve the users' intent if displays change such as moving from a docked laptop at work to the internal display at home.

lenemter:
Fixes #297


vjr:
Thanks for this PR @lenemter , looking forward to this feature :-)

According to the comment here: https://github.com/elementary/gala/pull/1751/files#diff-ba3fbdef1da3bc4b5454c2ad54c4a2751d0df4ed0dd64ee57c7c623bc161a570R63

Is there a hurdle tracking multiple windows of the same app - anyone have any ideas to solve this?

lenemter:
> Is there a hurdle tracking multiple windows of the same app - anyone have any ideas to solve this?

I think if an application has two windows (e.g. Files opened with 2 windows), after the first (primary) window is being closed, gala should track the second window. I'll try to implement this now, but it's worth asking @danirabbit for opinion.

vjr:
> > Is there a hurdle tracking multiple windows of the same app - anyone have any ideas to solve this?
> 
> I think if an application has two windows (e.g. Files opened with 2 windows), after the first (primary) window is being closed, gala should track the second window. I'll try to implement this now, but it's worth asking @danirabbit for her opinion.

What about the use case of multiple windows of the same app and not closing any of them (or closing only some of them, still leaving multiple open) and user maybe logs out or restarts?

I mean, say an app like Files or some third party say Firefox if I open 3 windows and move them around to my preferred positions, then close just 1 window (that should have its position saved) then while remaining 2 are still open, I logout/restart?

Is it possible to handle this kind of scenario in this PR? Of course, yes, will leave it to core devs / UX team opinion.

lenemter:
> Is it possible to handle this kind of scenario in this PR? Of course, yes, will leave it to core devs / UX team opinion.

That's interesting. I'll look into this, shouldn't be too hard to implement.

vjr (commented):
This is not a code review but just wanted to mention I tried this PR and it appears to work well already in its current state.

I was going to ask about also saving apps' sizing information but looks like each app mostly does that for itself already. Tried Files, Music and Firefox.

vjr:
@lenemter my quick attempt at tracking all windows: https://github.com/elementary/gala/pull/1752

One thought I had is what to do about potential DB buildup (cleanup old/stale saved entries) over time?

lenemter:
> One thought I had is what to do about potential DB buildup (cleanup old/stale saved entries) over time?

It shouldn't be a issue. Even if user launches unique 20 windows everyday for 5 years, it would be only 36k entries which I think isn't too much.

lenemter:
@vjr I tried to implement your idea. Can you check if it is what you have in mind please?

lenemter:
If someone wants to test this PR:

There are two commits of this branch: b3203ac27fba137eb84e4d22993781669b7e4970 that implements https://github.com/elementary/gala/pull/1751#issuecomment-1712768393 and df0db8621549eca1b54a09ddec3df10bc3dee49b that implements https://github.com/elementary/gala/pull/1751#issuecomment-1712769604

danirabbit:
I think what @jeremypw had implemented in Files before was to only track the last closed window, but afaict this is working well to track all windows :) I'm sure he might have some opinions though?

Nice work! Definitely super nice to have every window remembering position <3

jeremypw:
> I think what @jeremypw had implemented in Files before was to only track the last closed window, but afaict this is working well to track all windows :) I'm sure he might have some opinions though?

Only the last closed window was remembered because only one window was restored. Saving and restoring all app windows was something I have considered but never got around to - I don't think there is a wishlist request for this?  The state of each window (open tabs etc) would need to be saved so it would add to the complexity of the schema but is certainly doable if it is something that people want.

src/WindowStateSaver.vala
```diff
@@ -0,0 +1,163 @@
...
+        if (!(app_id in app_windows)) {
+            app_windows[app_id] = new GLib.Array<Meta.Window?> ();
+        }
+
+        var window_index = find_window_index (window, app_id);
+        app_windows[app_id].insert_val (window_index, window);
+
+        var tracking_window = false;
+        db.exec (
+            "SELECT last_x, last_y, last_width, last_height FROM apps WHERE app_id = '%s' AND window_index = '%d';".printf (app_id, window_index),
```
jlnr, line 76: This does not seem to escape `app_id`. I guess app_id will never contain ' characters, but is there an easy way to escape it? (Same for the `INSERT` statement below.)

src/WindowStateSaver.vala
```diff
@@ -0,0 +1,163 @@
...
+            return;
+        }
+
+        Sqlite.Statement stmt;
+        rc = db.prepare_v2 (
+            """
+            CREATE TABLE IF NOT EXISTS apps (
+                id           INTEGER PRIMARY KEY,
+                app_id       TEXT,
+                window_index INTEGER,
```
jlnr, line 29: What do you think about using `(app_id, window_index)` as the primary key and always upserting them, to avoid having duplicates in the database?

jlnr (commented):
Even for X11, this looks much nicer than forcing every app to implement it by itself :+1:

I wanted to take a look at this PR to see how it handles multi-monitor setups. What macOS seems to do is that it remembers the last screen that a window was actively moved to. Disconnecting a secondary screen collects all windows on the remaining display, re-connecting it restores things as they were before.

I don't mean to ask for extra code in this PR to cover more edge cases, but do you think it makes sense to add a column for storing the preferred screen index(?) now to avoid having to migrate the DB later on?

(Although I guess one could always come up with more fields in the future, like was_fullscreen, workspace_index etc. so maybe YAGNI should win for now?)

danirabbit (approved):
I'm inclined to get this in early so we can get lots of testing and potentially break the DB early if we have to

src/WindowStateSaver.vala
```diff
@@ -0,0 +1,162 @@
...
+public class Gala.WindowStateSaver : GLib.Object {
+    private static unowned WindowTracker window_tracker;
+    private static GLib.HashTable<string, GLib.Array<Meta.Window?>> app_windows;
+    private static Sqlite.Database db;
+
+    public static void init (WindowTracker window_tracker) {
+        WindowStateSaver.window_tracker = window_tracker;
+        app_windows = new GLib.HashTable<string, GLib.Array<Meta.Window?>> (GLib.str_hash, GLib.str_equal);
+
+        var path = Path.build_filename (Environment.get_home_dir (), ".local", "share", "io.elementary.gala-windowstate.db");
```
tintou, line 15:
```suggestion
        var dir = Path.build_filename (GLib.Environment.get_user_data_dir (), "io.elementary.gala");
        Posix.mkdir (dir, 0775);
        var path = Path.build_filename (dir, "windowstate.db");
```

vjr (approved):
I might not get to testing the PR until later in the weekend but I agree with @danirabbit comment to get this PR in early and work on any other issues , if discovered , as we go along.

lenemter:
@danirabbit I think currently window position is not saved before shutdown/sleep and doesn't save monitors correctly, but I'm too busy to implement it now. If this is ok for you we can merge this.

orowith2os:
Late concern, and I'm not sure if I can make sense of it from the patch - what happens if you store a position and restore it when the setup changes such that the stored coords are out of bounds of the current setup?

lenemter:
@orowith2os I don't know. As said it's tested very badly and was merged to get more feedback.

orowith2os:
I guess I'll be bug testing Elementary at some point in the future then - because this sounds like it could be Not Great.

PR #1752, opened 2023-09-10
Track all windows

vjr:
Track positions of all windows using `Meta.Window.get_id ()` to uniquely identify.

Tested with Firefox which has option to restore windows upon launch.

### DIFF
--- a/src/WindowStateSaver.vala
+++ b/src/WindowStateSaver.vala
@@ -5,12 +5,10 @@
 
 public class Gala.WindowStateSaver : GLib.Object {
     private static unowned WindowTracker window_tracker;
-    private static GLib.GenericSet<string> opened_app_ids;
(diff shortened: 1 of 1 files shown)


vjr:
Not sure if this is right, need to compare this PR vs stock gala.

lenemter:
I think `Meta.Window.get_id ()` returns unique window id and gala can't rely on it to track window in any way?

vjr:
> I think `Meta.Window.get_id ()` returns unique window id and gala can't rely on it to track window in any way?

I was looking for a way to ID the windows across sessions so that they all could be restored to their last positions, but I guess it might not be feasible. Your PR should be just fine whatever is the final code that gets merged, I appreciate it :-)

vjr:
Closing since the main PR covers everything.

PR #1756, opened 2023-09-13
Don't block `Alt + Shift` when there is only one keyboard layout

### LINKED ISSUE #1261: `Alt + Shift` unnecessarily blocked when there is only one keyboard layout.
## What Happened

See https://github.com/elementary/switchboard-plug-keyboard/issues/387  & https://github.com/elementary/switchboard-plug-keyboard/issues/270

## Expected Behavior

When there is only one keyboard layout in use (the most common situation?) then the "Switch Layout" shortcut should not be blocked for use in other shortcuts.

## Steps to Reproduce

see See https://github.com/elementary/switchboard-plug-keyboard/issues/387

## Platform Information

Odin daily

lenemter:
Fixes #1261


tintou:
The thing is, how does it react when you assign the shortcut to something else then add a keyboard layout?

lenemter:
@tintou the `Alt + Shift` layout will trigger layout switch and if a shortcut with these accelerators was set, it will be broken. This branch just blocks layout switch shortcut if there is only one keyboard layout in order to avoid the issue on such systems.

tintou (approved):

PR #1760, opened 2023-09-18
Redraw tooltip to avoid corner glitches

lenemter:
Fixes this

Screenshot from 2023-09-30 13 46 14


danirabbit:
Can you provide reproduction steps for this? I'm not seeing any clipping here

lenemter:
@danirabbit It usually appears when tooltip changes its size e.g. you opened some long path in Files and then opened a smaller one

danirabbit (approved):
Okay yup, I can confirm the issue in master and that this fixes it. Nice job!

Issue #1694, opened 2023-06-14
Panel and Dock are treated as normal windows on Wayland


danirabbit:
### What Happened?

On Wayland, the panel and dock are treated as normal windows and aren't treated specially like panel and dock class windows should be

### Steps to Reproduce

1. Run a wayland session
2. Open the multitasking view
3. See that the panel and dock are treated as normal windows instead of animated out etc

### Expected Behavior

The panel and dock should be special cased so that they have different behavior from normal windows

### OS Version

7.x (Horus)

### Software Version

Compiled from git

Issue #1693, opened 2023-06-14
Notifications appear in the center of the screen on Wayland


danirabbit:
### What Happened?

On Wayland, notification windows appear in the center of the screen instead of in the top right (or top left for RTL)

### Steps to Reproduce

1. Run a Wayland session
2. Receive a notification
3. See that the notification window is centered on the display instead of in the top corner

### Expected Behavior

Notifications should be aligned top right for LTR and top left for RTL

### OS Version

7.x (Horus)

### Software Version

Compiled from git

Marukesu:
(stealing this issue a bit for a more overall view to how integrate system components with gala in wayland, this text is also valid as solution to #1694 and elementary/switchboard-plug-display#352)

I though about this a little some time ago, i think we can use a "wayland protocol as DBus interfaces" concept, and then make it a proper wayland protocol when we have the tools to do it.

For now, we can go with something like this:
```
 // base interface that would be also implemented by the "derived" interfaces,
 // something like the "xdg_surface" providing a role property to be set by others 
 // interfaces.
 @interface io.elementary.Gala.Surface {
     @property role: (s) // possibly values would be "notification", "bubble", and "panel"
 }

 // for wingpanel and dock
 @interface io.elementry.Gala.Panel {
     @property anchor: (i) // top, bottom, left, right
 }

 // for the display labels that switchboard uses
 @interface io.elementary.Gala.Bubble {
     @property anchor: (i) // top, top_left, top_right, etc... 
 }
```

The way this would work is that applications would export the interfaces in the window's object path set with `gtk_shell.set_dbus_properties` method in wayland, or the `_GTK_WINDOW_OBJECT_PATH` xprop in x11. In gala, we would lookup the path to see if the interfaces exists and handle the Meta.Window as expected.

The major issue with this is that `Gdk.Wayland.Toplevel.set_dbus_properties()` is private in Gtk4. However, this would allow us to design and make use of what would look like, more or less, a wayland surface, making more easy to move from DBus interfaces to a wayland protocol in the future.

davidmhewitt:
I haven't looked at it for a while, but Mutter has a concept of shell subprocesses built in (I can't remember what they call it now).

But if we used that, it would essentially make Gala responsible for starting the Dock and the panel and maybe the notification server. But, the fact that they're trusted subprocesses allows them to set some of the custom hints, like removing themselves from the dock and Alt+Tab switcher, and because Gala is tracking those processes, it could position their windows into the correct places.

[...]

PR #1761, opened 2023-09-25
Use Meta.WaylandClient for shell components

leolost2605:
Tries to fix stuff like #1693 and #1694 by using `Meta.WaylandClient` to launch the respective apps as trusted subprocesses and by managing their position etc. from within gala as [suggested](https://github.com/elementary/gala/issues/1693#issuecomment-1593531836) by @davidmhewitt.
As a proof of concept this PR currently positions the new dock in the top left corner (I didn't bother getting the maths right for the correct position with the possibility we might not even use this :)) of the primary monitor, keeps it above all other windows and hides it from the window list.
It's based on the `wayland-testing` branch because for some reason `master` still crashes for me but can be easily rebased.
@tintou @Marukesu is this worth pursuing?

The code is probably not very good, it's the first time I do something window manager related and this is just the result of me playing around with wayland :) Still it works and shouldn't require too much additional effort to get it ready 🤞


src/PantheonShell.vala
```diff
@@ -0,0 +1,46 @@
...
+        display.window_created.connect ((window) => {
+            if (dock_client != null && dock_client.owns_window (window)) {
+                setup_dock_window (window);
+            }
+        });
+
+        var subprocess_launcher = new GLib.SubprocessLauncher (NONE);
+        try {
+            dock_client = new Meta.WaylandClient (subprocess_launcher);
+            string[] args = {"io.elementary.dock"};
```
leolost2605, line 26: For a more useful way to test this replace dock with wingpanel and remove the TryExec for io.elementary.wingpanel from the wayland session settings and it works really fine apart from maximized windows disappearing behind the panel :)

tintou:
Hi there, you should take a look at https://github.com/elementary/gala/pull/1705

leolost2605:
Hi, yes I've seen that and even tried to do some direct xdg shell wayland access from a client side to see how easy/hard using custom wayland protocols is but at some point (I think when I just couldn't figure out how to use the headers generated for wayland protocols from within vala without manually writing vapi, although I can imagine I probably did something very wrong🙂) just gave up and found this to be at least IMHO a much simpler solution that (as @davidmhewitt pointed out) doesn't require new API and any (?) client side effort

leolost2605:
Also with this it's easier for new people, like me, to contribute without having to dig deep down into how wayland works :)

PR #1762, opened 2023-09-26
Don't use Gtk.init

lenemter:
For some reason it breaks Wayland session. Needs to be tested on other distros.


leolost2605:
I can confirm this fixes the issue for me, I'm not sure about unintended side effects and the other distros as you said though so I'll leave a review to someone else 🤷

tintou:
I'm tempted to just merge this, it's easy enough to revert if we see something wrong

tintou (approved):

meisenzahl:
This change causes an X11 session to crash on Fedora 38.

A Wayland session crashes both with and without the change.

@tintou do you have any idea how to better identify the reason for the crash in a wayland session?

PR #1766, opened 2023-09-29
Fix crash

lenemter:
This should fix crash in described in https://github.com/elementary/gala/discussions/1764.

~~I decided to add `instance` field to `WindowManagerGala` to avoid passing it over object constructors down to `BackgroundSource`. I'll prepare similar branch for other objects later. Is this a good idea?~~

Moves handling of `BackgroundSource` to `BackgroundManager`, which allows passing `wm` into it.


tintou:
Hmm no we shouldn't create static objects like this all over the place, the proper fix should be elsewhere

PR #1767, opened 2023-09-29
BackgroundSource: Keep a reference to the monitor manager

tintou:
Also use a GLib.HashTable instead of Gee.HashMap as it is enough for here.


lenemter (approved):
Looks good, thanks for this! I guess this should fix https://github.com/elementary/gala/discussions/1764, shouldn't it?

tintou:
It should theoretically, but I haven't tested it

PR #1768, opened 2023-09-29
WorkspaceManager: Fixes and cleanup

### LINKED ISSUE #506: After login there is always 2 unpopulated workspaces open.
1- Login.
2- Press Super+Down to go to the workspace overview

**Result:** There is 2 unpopulated workspaces. 1 2 +

elementary OS Juno fully updated. New profile,

PS: I'm sorry if this is a duplicated issue.

### LINKED ISSUE #695: Wrong workspaces behavior
By default on login there are two workspaces open (screen attached). This issue leads to strange fullscreen behavior for fullscreen applications. For example videos in firefox app opens fullscreen in same workspace with firefox, but this behaviour changes after pressing multitask button or open other app fullscreen.
Снимок экрана от 2020-01-11 01 51 56
Also it makes impossible to switch fullscreen-on-new-workspace behaviour, it stops working again after multitasking button is pressed or some other app opened in fullscreen.

lenemter:
Fixes #506
Fixes #695

Changes:
1. On boot there is now only a single "+" workspace
2. If a window is removed from the second to last workspace and it becomes empty, it will become the "+" workspace

### DIFF
--- a/data/gala.gschema.xml
+++ b/data/gala.gschema.xml
@@ -355,6 +355,7 @@
   <enum id="GestureSwipeHorizontal">
     <value nick="none" value="0" />
     <value nick="switch-to-workspace" value="1" />
+    <value nick="move-to-workspace" value="2" />
   </enum>
   <enum id="GestureSwipeUp">
     <value nick="none" value="0" />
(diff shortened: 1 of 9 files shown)


danirabbit (approved):
Makes sense and looks good

Issue #1301, opened 2021-11-08
Make Window Switcher accessible


aral:
### What Happened?

Last week [Catts](https://github.com/small-tech/catts) was brought in-tree to gala as the default Window Switcher.

However, it is entirely inaccessible.

(e.g., for a person who uses a screen-reader, it does not exist at all)

I was able to cobble together some basic accessibility for Catts from whatever scant documentation and code I could find on the subject in the GNOME/Gtk/Vala/elementary OS worlds and, while it is better than nothing, it should be possible to improve it.

I hope that someone with the necesary knowledge can help with this and also share their knowledge to teach other developers both on the elementary OS core team and those making apps for the operating system how to create accessible widgets by default.

I’m opening this issue as per @cassidyjames’s request that [making the widget accessible should be tracked in a separate issue](https://github.com/elementary/gala/pull/1234#issuecomment-930397772).

Related: https://github.com/elementary/hig/discussions/51

### Steps to Reproduce

1. Use the Window Switcher with the screen reader

Nothing is heard.

### Expected Behavior

The Window Switcher should be accessible.

### OS Version

6.x (Odin)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell
n/a
```

### Hardware Info

n/a

Tireg:
Probably a duplicate of #1326 (all Mutter version impacted since Mutter 40)

danirabbit:
I wonder if we should make this UI element into a GTK subprocess like we've done with some other shell elements. That would give us a11y for basically free and we could remove a lot of custom drawing

PR #1769, opened 2023-09-29
Use clutter for the background menu

### LINKED ISSUE #1695: No context menu on wallpaper in Wayland
### What Happened?

I'm not able to reliably open the context menu when running under Wayland

### Steps to Reproduce

1. Run a wayland session
2. secondary click the wallpaper
3. Nothing happens

### Expected Behavior

Context menu with options for changing the wallpaper, display settings, etc

### OS Version

7.x (Horus)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

### LINKED ISSUE #1696: Window context menu doesn't automatically close on Wayland
### What Happened?

The context menu that appears when secondary clicking on window titles doesn't close when clicking outside of it

### Steps to Reproduce

1. Run a wayland session
2. Secondary click on a window title
3. See the context menu with window management options
4. Click outside of the menu to dismiss it
5. nothing happens

### Expected Behavior

Clicking outside of the menu should dismiss it

### OS Version

7.x (Horus)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

leolost2605:
The current way of handling window and desktop menus doesn't work under wayland because the daemon can't position the menus and AFAIK they don't work properly without a parent (see @Marukesu's comment [here](https://github.com/elementary/gala/issues/1695#issuecomment-1593373538))

With this PR we add a `Gala.Menu` Clutter Actor that is then used to create the background menu. In a follow up the same can be used (with some additional effort) for the window menu.

The appearing and disappearing of the menu is currently slightly animated not sure if we want to keep that (I just took it from the WindowSwitcher).
The original Gtk.Menu's add some additional space to the end of the menu items. When porting to GTK4 we don't do that when replacing Menus with popovers, and popovermenus don't do it by themselves so I didn't do it here.
Fixes #1695 and sets the foundation for a follow up to f*x #1696


lenemter:
Seems like the menus don't get properly destroyed and still take input after being closed.


https://github.com/elementary/gala/assets/80143868/b4d6ee3a-9be5-4f06-9d62-a83751467f36

---

Second, they can go off screen when clicked close to the screen edge.

image

---

Third, seems like there is no black outline in dark mode unlike other menus. I briefly looked into the stylesheet code and couldn't find why it's broken.

And the last, the menu is shorter for some reason. Menus' min-width is clearly 150px in stylesheet, but in reality it's 205?

danirabbit:
I’m a little concerned about like needing to create and maintain half a toolkit here to support everything that Gtk does with a11y and localization etc. maybe @tintou has thoughts/suggestions.

I wonder if it would be better to just have a full screen transparent window so that it could position its own popover?

tintou:
Yes I think that the background menu is an exact usecase of Meta.WaylandClient as it is tightly tied to Gala

lenemter:
@danirabbit @tintou Gala still needs a Gala.Menu for window menus. So why don't just use Gala.Menu for everything?

And talking about gala's toolkit, it has many a11y issues e.g. #1301, but that doesn't mean we need to replace say window switcher with Gtk implementation. We should put more effort into developing gala instead of searching hacky solutions with Gtk.

leolost2605:
As @lenemter said the window menu was my primary concern too. I think we would have to go very hacky to get that working with only GTK...?
Because the main issue I had was being able to close the menu by clicking outside of it.
I tried doing a push modal with the `WindowActor` for a daemon `Gtk.Window` which didn't work though but would that be possible in theory and I just did something wrong? Or maybe there is a way to get all mouse button clicks in gala before they get consumed by a child?

lenemter (commented):
Oops I forgot to click submit button

leolost2605:
See #1844

PR #1776, opened 2023-10-23
Background: Cleanup

### LINKED ISSUE #1774: Scheduled switch to dark style does not dim the wallpaper after cold reboot on elementary OS 7.1
### What Happened?

dim_bug_01

dim_bug_02

dim_bug_03

dim_bug_04

### Steps to Reproduce

Prerequisite:
The desktop is in the default light style

Step 1:
Go to System Settings --> Desktop --> Wallpaper Tab
Enable (switch on) the feature "Dim with dark style"

Step 2:
Go to System Settings --> Desktop --> Appearance Tab
Set the Schedule From [actual time + 2 minutes] To [actual time + 1 hour]
Now shut down the OS

Step 3:
Wait at least 3 minutes before booting up again
Log in
The desktop should now have switched to dark style.
As you can see plank the dock appears now dark
BUT the wallpaper has not been dimmed.

Step 4:
To confirm the bug:
Go to System Settings --> Desktop --> Wallpaper Tab
Manually switch off and on again the feature "Dim with dark style"
NOW the wallpaper is dimmed

### Expected Behavior

After the cold reboot, the desktop switches from light style to dark style AND the wallpaper should also switch from normal state to dimmed state.

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

lenemter:
Fixes #1774

- Don't use custom hacky way to set dimmed wallpaper
- Handle every settings change in `BackgroundSource`

### DIFF
--- a/data/gala.metainfo.xml.in
+++ b/data/gala.metainfo.xml.in
@@ -33,6 +33,7 @@
         </ul>
       </description>
       <issues>
+        <issue url="https://github.com/elementary/gala/issues/1774">Scheduled switch to dark style does not dim the wallpaper after cold reboot on elementary OS 7.1</issue>
         <issue url="https://github.com/elementary/gala/issues/1261">Alt + Shift unnecessarily blocked when there is only one keyboard layout</issue>
       </issues>
     </release>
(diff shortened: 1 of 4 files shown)


danirabbit (approved):
Looks good, changes make sense, and I can confirm that it fixes the issue!

PR #1778, opened 2023-10-25
Don't create window snapshots if animations are disabled

lenemter:


danirabbit (approved):
I can confirm that this variable is only used in functions that already return early if animations are disabled, so this makes sense. Nice little optimization!

PR #1779, opened 2023-10-25
Background: respect animations key

lenemter:


danirabbit (approved):
Works as expected, nice little code style fix there too. Thank you!

Issue #1739, opened 2023-07-09
Support for transform current focused app to other/new workspace


sauntor:
### Problem

The shortcut `Meta` + `Alt` + `1/2/3/4/...` is to hard for notebook users

### Proposal

maybe we can transform focused app to another workspace via touchpad gestures, i.e., four finger swipe,  or `Meta` + mouse drag, etc.

PR #1780, opened 2023-10-26
Add "Move to workspace" gesture

### LINKED ISSUE #1109: Use gestures for "move to workspace"
## Problem

When using keyboard shortcuts you can either switch to a workspace with super + arrows or move a window to a workspace with super + alt + arrows. With gestures however, there is only the option to switch to a workspace, it's not possible to move a window to another workspace

## Proposal

It would be nice to be able to configure a horizontal swipe gesture to move the focused window in the direction of the swipe. For example, by default swiping horizontally with 3 fingers switches workspaces, and optionally one could move the focused window with a 4-finger swipe.

lenemter:
Closes https://github.com/elementary/gala/issues/1739
Fixes #1109

### DIFF
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -564,7 +564,7 @@ namespace Gala {
                 unowned var manager = display.get_workspace_manager ();
 
                 var direction = gesture_tracker.settings.get_natural_scroll_direction (gesture);
-                
+
                 moving = display.focus_window;
                 moving.change_workspace (manager.get_active_workspace ().get_neighbor (direction));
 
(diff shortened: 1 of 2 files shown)


src/WindowManager.vala
```diff
@@ -531,17 +531,47 @@ namespace Gala {
...
+
                 var direction = gesture_tracker.settings.get_natural_scroll_direction (gesture);
+
+                moving = display.focus_window;
+                moving.change_workspace (manager.get_active_workspace ().get_neighbor (direction));
+
                 switch_to_next_workspace (direction);
+                return;
             }
+
```
lenemter, line 574:
```suggestion
```

danirabbit (approved):
This is perfect! Feels super good, love it. Nice code cleanup too, makes the logic way clearer here

PR #1781, opened 2023-10-26
Increase `window-clone` shadow size

lenemter:
Fixes clipped shadows in multitasking view:
Screenshot from 2023-10-26 14 04 21


danirabbit:
Can you provide reproduction steps for this? Is this happening on CSD or SSD windows or all windows or windows of a certain size?

lenemter:
@danirabbit the bug occurs on every maximized window

danirabbit (approved):
Ah okay, I have it set so maximized windows go to their own workspace :sweat_smile: I can confirm the fix. Nice work!

PR #1782, opened 2023-10-26
Use different cursors for dnd

lenemter:


danirabbit (approved):
Thank you for your attention to detail! Looks great <3

PR #1787, opened 2023-10-28
MultitaskingView: cleanup and bugfixes

lenemter:
- Don't disable gesture tracker as it can stuck if you switch workspace and immediately close the multitasking view
- Switch to target workspace early
- Use `Meta.Workspace` objects directly instead of dealing with their indexes


danirabbit (approved):
I can confirm that closing immediately after switching can end up breaking gestures and I can't reproduce that in this branch :rocket:

PR #1792, opened 2023-10-30
Gtk.init only in X11 session

meisenzahl:
The change introduced in #1762 causes a X11 session to crash on Fedora 38. After reverting the change, the session doesn't crash.

On Fedora 38 a Wayland session crashes both with and without the change introduced in #1762.

To fix the issue, GTK should be initialized only in a X11 session.


tintou (approved):
Let's do it for now

Issue #485, opened 2019-03-01
Deduplicate Multitasking View and window spread


cassidyjames:
Right now we have the standard Multitasking View (Super+S or Super+Down) plus a window spread (Super+W). These are largely duplicated, with each having slightly different features (advantages **bolded**):

| Multitasking View                       | Window Spread                      |
|-----------------------------------------|------------------------------------|
| Super+S, Super+Down                     | Super+A                            |
| Shows windows on current workspace      | Shows windows on all workspaces |
| Can close windows w/X icon              | Can close windows w/X icon         |
| Middle-click close windows              | Middle-click close windows         |
| Supports keyboard navigation        | Supports keyboard navigation             |
| Shows context of other workspaces   | Shows windows on all workspaces           |
| **Can move windows across workspaces**  | No workspace functionality         |
| **Discoverable via Hotcorner settings** | Explicitly removed from Hotcorners |
| Only shows on primary display           | **Shows windows on secondary displays** |

We should standardize on the Multitasking View since it's just as fast but more useful, showing the window spread _and_ workspaces, plus supporting keyboard navigation. If we wish to continue to support a window spread across all workspaces (Super+A), we could leave that part as-is—though I personally would be fine dropping it to reduce complexity and duplication in the codebase.

The one feature I know of that the Window Spread handles better is windows on secondary displays, so that would need to be ported over to the Multitasking View.

As part of the migration, we should probably switch Super+W to the Multitasking View to keep muscle memory the same.

Blast-City:
Hi.
The following threads talk about this issue:

#287 
#415 

Wouldn't it be better to have only on place to discuss this?

Blast-City:
The show all windows open on the overview could be beneficial to reorganizing workspaces. Pressing Super+A and then grab each window to desired workspace on the bottom seems to be fast. 

Having said that i think it wouldn't be a feature that people would miss that much.

janxkoci:
So this explains why I can no longer assign window spread to hot corner. I don't like this too much.

First of all, in my experience from running elementary on lower-spec laptop, I can say that the animation might be of same speed, but performance-wise the window spread is much lighter on the resources. The multitasking view can freeze my laptop for a few seconds, while window spread rarely does that (only under heavy load).

[...]

PR #1793, opened 2023-10-30
MultitaskingView: move keyboard navigation code into WindowCloneContainer

### LINKED ISSUE #6: Can't navigate open windows with keyboard when doing Super+A
Every time I use super+A/super+w or super+s I have to move my hands from the keyboard to the mouse to select the window I want to navigate to. 
We should be able to navigate the windows using the keyboard arrows and select with [Enter].

It would be cool if each window had a shortcut tag above then to fast switch. Eg: Press supar+a then 1 to switch to first window.

lenemter:
Fixes #6


leolost2605:
Any reason this is still drafting? Works pretty good in my testing and would be - if I understand it correctly - useful for #1833

lenemter:
@leolost2605 I was aware of https://github.com/elementary/gala/issues/485 and thought we might remove window spread altogether, but since it has found a new purpose I'll finish this branch... someday.

leolost2605 (approved):
LGTM 🎉🎉

PR #1794, opened 2023-10-30
IconGroup: remove close button

### LINKED ISSUE #348: UX: It's too easy to trigger the action "Close All Windows in Workspace" when switching between workspaces with mouse
It's a destructive action, without any confirmation dialog and IMHO not very useful.
Happened to me several times. Specially with that delayed effect before "X" appears. You don't see it, go for a clicking a BUM! the "X" apperead and you did an unwanted destructive action.

worspace close all

lenemter:
Fixes #348


danirabbit (approved):
Approving but waiting to merge in case we want to make that point release first before we start breaking things for OS 8 😅

danirabbit:
Let's go :rocket:

PR #1777, opened 2023-10-25
Release 7.1.3

lenemter:
Waiting on:
- https://github.com/elementary/gala/pull/1748
- https://github.com/elementary/gala/pull/1760
- https://github.com/elementary/gala/pull/1786
- https://github.com/elementary/gala/pull/1787
- https://github.com/elementary/gala/pull/1789
- https://github.com/elementary/gala/pull/1788

PR #1795, opened 2023-10-30
Remove `expose-windows` shortcut

lenemter:
The first step towards https://github.com/elementary/gala/issues/485
Removes window spread for one workspace. AFAIK it does the exact same thing as the multitasking view, so it can be removed.

Do not merge until #1777

### DIFF
--- a/data/gala.gschema.xml
+++ b/data/gala.gschema.xml
@@ -7,11 +7,10 @@
     <value nick="minimize-current" value="3" />
     <value nick="open-launcher" value="4" />
     <value nick="custom-command" value="5" />
-    <value nick="window-overview" value="6" />
-    <value nick="window-overview-all" value="7" />
-    <value nick="switch-to-workspace-previous" value="8" />
-    <value nick="switch-to-workspace-next" value="9" />
-    <value nick="switch-to-workspace-last" value="10" />
+    <value nick="window-overview-all" value="6" />
+    <value nick="switch-to-workspace-previous" value="7" />
+    <value nick="switch-to-workspace-next" value="8" />
+    <value nick="switch-to-workspace-last" value="9" />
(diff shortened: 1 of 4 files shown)


data/gala.gschema.xml
```diff
@@ -7,11 +7,10 @@
     <value nick="minimize-current" value="3" />
     <value nick="open-launcher" value="4" />
     <value nick="custom-command" value="5" />
-    <value nick="window-overview" value="6" />
-    <value nick="window-overview-all" value="7" />
-    <value nick="switch-to-workspace-previous" value="8" />
-    <value nick="switch-to-workspace-next" value="9" />
-    <value nick="switch-to-workspace-last" value="10" />
+    <value nick="window-overview-all" value="6" />
```
danirabbit, line 10: I wonder if changing the value here will mess up system settings? It seems odd that the value is a number here tbh. But might need to preserve the numbers for now even though there would be a missing option just to not break things

data/gala.gschema.xml
```diff
@@ -148,7 +147,7 @@
       <summary>Zoom out</summary>
     </key>
     <key type="as" name="expose-windows">
```
danirabbit, line 149: Do we need this key at all? If not we should change the description of the key to reflect that it's deprecated as of this version and will be removed in the next major version

lenemter: I marked it as deprecated now.

data/gala.gschema.xml
lenemter: I think this enum is used only in hotcorners key, and settings plug set the value using string and not actual enum value. So it shouldn't break anything.

danirabbit (approved):
Let's go :rocket:

PR #1798, opened 2023-10-31
Remove  `workspace-switcher-background` key

lenemter:
Review with hidden whitespace changes

While we're here:
- Inline namespace
- Update copyright header


danirabbit (approved):
:rocket:

Issue #1317, opened 2021-11-26
Add a Gesture option to scroll and switch between open windows, similar to Alt Tab


capttrousers:
### Problem

Alt Tab allows you to hold Alt, and continually press Tab until you get to the open app window you want to switch to, but it doesnt allow you to use the arrow keys. And forces you to hit the correct keys.

Using gestures is a more natural UX, considering people's familiarity with gestures and touch screens with modern smart phones.

### Proposal

Add the ability to use 3 or 4 finger Right/Left swipes to scroll through open windows. When using Alt Tab, the currently selected window shows up on top, with the rest being grayed out beneath, while you are still holding Alt, so a similar animation of control could be used.

Once you pick up your fingers, the currently selected app window is now the top window with focus, similar to how letting go of Alt will focus on that window. 

The "Scroll Through Open Windows" option would be in the drop down of the Gestures settings page, for Swipe Horizontally, 3 or 4 fingers. Now there is just one option for switch to workspace, but this would add an additional option to that drop down menu. 

Stopping and scrolling the other way should scroll naturally, both left and right and back and so forth, without picking up your fingers.

cassidyjames:
I think this makes sense; the bigger question for me is if it would be possible to have touchégg and Gala coordinate that gesture. cc @JoseExposito

JoseExposito:
This is already possible just with Touchégg, check example 2 or add the shortcuts with [Touché](https://github.com/JoseExposito/touche) for testing:
https://github.com/JoseExposito/touchegg#keyboard-shortcut-send_keys

The user experience can be improved by, for example, not allowing to jump to the start of the window list from the end, but that should give you enough to test and get a good idea of the raw edges we'd need to polish.

In my experience implementing this kind of "repeat action" gestures it's important to:

1. Keep the physical distance you need to move your fingers to jump one item constant.
For example, imagine 2 scenarios where we have 2 and 5 windows open. It's important that the amount of movement required in both scenarios to change from item 1 to 2 is the same for muscle memory.

2. I found that 10 steps are the sweet spot. Allowing more steps makes selecting items precisely very difficult, specially on small touchpads.
The problem is handling more than 10 items, in this case, when the user has more than 10 windows open. Personally, I'd not allow to select the window number 11, because of what I explained in 1. and because it will get really difficult to select it precisely.

In case you want to explore other values, change this constant and re-compile:
https://github.com/JoseExposito/touchegg/blob/master/src/actions/repeated-action.cpp#L46

[...]

Issue #979, opened 2020-12-02
Add gesture to switch between windows on a touchpad


fs3o5:
## Problem

There is no way to cycle through open windows on a touchpad.

## Proposal

Since we are now getting a lot of gestures I would like to see a way to cycle through windows using a touchpad.
Perhaps simular to how windowes does it. [Example at 0:51](https://youtu.be/cN-Y4LR5r-8?t=51)

JoseExposito:
I cannot tell you if this is something the elementary team wants to include by default. But, you will be able to configure it if you want:

https://github.com/JoseExposito/touchegg#keyboard-shortcut-send_keys

```
<gesture type="SWIPE" fingers="3" direction="LEFT">
  <action type="SEND_KEYS">
    <repeat>true</repeat>
    <modifiers>Alt_L</modifiers>
    <keys>Shift_L+Tab</keys>
    <decreaseKeys>Tab</decreaseKeys>
  </action>
</gesture>

<gesture type="SWIPE" fingers="3" direction="RIGHT">
  <action type="SEND_KEYS">
    <repeat>true</repeat>
    <modifiers>Alt_L</modifiers>
    <keys>Tab</keys>
    <decreaseKeys>Shift_L+Tab</decreaseKeys>
  </action>
</gesture>
```

PR #1802, opened 2023-11-12
Add "Switch windows" gesture

lenemter:
Fixes https://github.com/elementary/gala/issues/1317
Fixes https://github.com/elementary/gala/issues/979

### DIFF
--- a/data/gala.gschema.xml
+++ b/data/gala.gschema.xml
@@ -355,6 +355,7 @@
     <value nick="none" value="0" />
     <value nick="switch-to-workspace" value="1" />
     <value nick="move-to-workspace" value="2" />
+    <value nick="switch-windows" value="3" />
   </enum>
   <enum id="GestureSwipeUp">
     <value nick="none" value="0" />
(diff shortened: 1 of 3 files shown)


danirabbit (commented):
This works for me! Is there a reason that only swiping to the right works? Can we initiate the switch when swiping either direction?

lenemter:
> Is there a reason that only swiping to the right works? Can we initiate the switch when swiping either direction?

This is a limitation of our current gesture system and I have no idea how to avoid it. So basically gesture have a direction and a progress. If you initiate swipe to left, you cannot change your mind and swipe to the right to select different windows. Which might be unexpected so I removed it. You can see how it worked in [00920b6](https://github.com/elementary/gala/pull/1802/commits/00920b6e18cb5745950609bbbac0987b1e786b75).

TomiOhl:
Not sure if this is any help, but currently this can be done via touchégg [like this](https://github.com/JoseExposito/touchegg#keyboard-shortcut-send_keys:~:text=Example%202%3A%20Switch%20between%20windows%20(Alt%2BTab)). Probably it's different as it's a repeated gesture instead of a long continuous one.
Maybe there is a way to cancel the gesture when the direction changes and start one to the opposite direction?

lenemter:
I decided to allow swiping to the left, but you cannot jump to the start of the window list from the end, which hides the flaw of the gesture system and emulates Windows 10 gestures behaviour.

danirabbit (approved):
Works perfectly, great job :)

PR #1804, opened 2023-11-18
Add pinch to screenshot gesture

lenemter:
I think it's a good gesture for a pinch, which currently is used only for zoom.

### DIFF
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -529,7 +529,6 @@ namespace Gala {
             }
         }
 
-
         private void on_gesture_detected (Gesture gesture) {
             if (workspace_view.is_opened ()) {
                 return;
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -263,12 +263,12 @@ namespace Gala {
             display.add_keybinding ("switch-input-source", keybinding_settings, Meta.KeyBindingFlags.IGNORE_AUTOREPEAT, (Meta.KeyHandlerFunc) handle_switch_input_source);
             display.add_keybinding ("switch-input-source-backward", keybinding_settings, Meta.KeyBindingFlags.IGNORE_AUTOREPEAT, (Meta.KeyHandlerFunc) handle_switch_input_source);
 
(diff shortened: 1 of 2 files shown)


src/WindowManager.vala
```diff
@@ -530,51 +534,69 @@ namespace Gala {
...
+                switch_to_next_workspace (natural_scroll_direction);
                 return;
             }
 
+            var three_fingers_move_to_workspace = fingers == 3 && three_finger_swipe_horizontal == "move-to-workspace";
+            var four_fingers_move_to_workspace = fingers == 4 && four_finger_swipe_horizontal == "move-to-workspace";
+
             switch_workspace_with_gesture = three_fingers_move_to_workspace || four_fingers_move_to_workspace;
-            if (switch_workspace_with_gesture) {
+            if (switch_workspace_with_gesture && is_horizontal_direction) {
```
zeebok, line 562: This looks to be the same check as on 552?

danirabbit:
I've been trying to figure out for a while why I'm feeling negatively about this, and I think I figured it out. For all of the other swipe gestures, the action has a continuous 1:1 response. So for example swiping to switch a workspace your getting continuous feedback as the action happens. But here there's no way to have a kind of continuous feedback for a very discrete action. So the gesture feels kind of unrelated to the thing that happens in response to the gesture.

I wonder for some action like this if it would be better to have like a double tap with 3 fingers or something like that

zeebok:
> I've been trying to figure out for a while why I'm feeling negatively about this, and I think I figured it out. For all of the other swipe gestures, the action has a continuous 1:1 response. So for example swiping to switch a workspace your getting continuous feedback as the action happens. But here there's no way to have a kind of continuous feedback for a very discrete action. So the gesture feels kind of unrelated to the thing that happens in response to the gesture.
> 
> I wonder for some action like this if it would be better to have like a double tap with 3 fingers or something like that

I can see that. Though I could also imagine the gesture causing some sort of overlay to appear, "framing" the shot so to speak, that way there's visual feedback to the gesture, and if you swipe quick you just don't see it or something. But that is like a pipe dream kind of implementation.

lenemter:
I'll close the PR for now and revisit pinch gesture ideas later.

Issue #1727, opened 2023-07-16
openning multitasking view can crash gala


janxkoci:
### What Happened?

In recent weeks gala occasionally crashes after opening the multitasking view (typically with a touchpad gesture - three-finger swipe up). It even happened once right in front of my boss when I needed to show him something and he was not entertained.

The crash manifests as all windows disappearing while in multitasking view, with only a thin outlines where windows should be. Closing the view can make windows appear again, but if I switch to another workspace, then the windows will no longer be visible even after returning to the original workspace.

The only way to fix this AFAIK is to `killall gala` from another TTY (I get there by pressing e.g. ctrl+alt+F1). Then I can go back to a GUI session (with ctrl+alt+F7) and watch gala and all windows appear again.

Below I attach partial output of `journalctl` as `coredumpctl` reports no coredumps and `gdb` cannot find gala. Note also that the line directly above `...skipping...` was from May, so I guess the journalctl output may be truncated.

### Steps to Reproduce

1. open multitasking view with touchpad gesture
2. roll a dice I guess - it only happens sometimes

### Expected Behavior

Gala should not crash, especially while interacting with desktop in normal ways that didn't bring it to crash previously.

Gala was very solid from Freya through Loki up to Juno and Hera, but since Jolnir (I skipped Odin) it started to crash a lot and even though it's more stable in Horus it still crashes for unknown reasons.

### OS Version

7.x (Horus)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell
...skipping...
čec 16 15:20:38 slimbook gala[1683]: Can't update stage views actor <unnamed>[<ClutterActor>:0x555f27e178d0] is on because it needs an allocation.
čec 16 15:20:38 slimbook gala[1683]: Can't update stage views actor <unnamed>[<ClutterActor>:0x555f27e17290] is on because it needs an allocation.
čec 16 15:20:38 slimbook gala[1683]: Can't update stage views actor <unnamed>[<GalaWorkspaceClone>:0x555f28199180] is on because it needs an allocation.
čec 16 15:20:38 slimbook gala[1683]: Can't update stage views actor <unnamed>[<GalaFramedBackground>:0x555f2819a830] is on because it needs an allocation.
čec 16 15:20:38 slimbook gala[1683]: Can't update stage views actor <unnamed>[<MetaBackgroundActor>:0x555f2820d550] is on because it needs an allocation.
[...]

PR #1805, opened 2023-12-02
Improve debug messages in notifications stack and add more checks

lenemter:
Should fix the crash described in https://github.com/elementary/gala/issues/1727#issuecomment-1836488852 and adds more debug messages.


danirabbit (approved):
:rocket: lgtm!

Issue #1648, opened 2022-04-16
Option to Disable/Enable Hot Corners 


hrstwn:
### Problem

Hot corners is a really nice feature that I use a lot, but on certain apps, it's quite annoying to have it activated. Especially when playing games that expect the cursor to go from corner to corner. In my case, (on some Proton games, so it's probably not elementary's fault) sometimes the _multitasking view_ makes the game crash and the _maximize current window_ moves the game window to another monitor which makes it necessary for me to turn it off. 

The problem is, the only way right now to make this work is to open settings, and set all my corners manually (one by one) to _'do nothing'_ whenever I want to play games, and reactivating it (one by one) when I'm done with my game. It's probably a very niche problem and not many people have the issue but I'd be very thankful if it's solved.

### Proposal

I think it'd be nice to have an switch to disable/enable hot corners or maybe a checkbox to disable hot corners for full-screen apps only. I'm not sure.

danirabbit:
I think it probably makes sense for us to disable hotcorners altogether for fullscreen apps.

Thoughts @lenemter ?

danirabbit:
@lenemter that makes sense to me. Just a single setting to enable/disable all hot corners for full screen apps probably solves the issue

PR #1806, opened 2023-12-02
Add an option to disable hotcorners in fullscreen

lenemter:
Fixes https://github.com/elementary/gala/issues/1648
Also fixes an issue in enum introduced in https://github.com/elementary/gala/pull/1795

### DIFF
--- a/data/gala.gschema.xml
+++ b/data/gala.gschema.xml
@@ -71,7 +71,7 @@
       </description>
     </key>
     <key type="b" name="enable-hotcorners-in-fullscreen">
-      <default>false</default>
+      <default>true</default>
       <summary>Whether hotcorners should be enabled when fullscreen window is opened</summary>
       <description></description>
     </key>
--- a/data/gala.gschema.xml
+++ b/data/gala.gschema.xml
@@ -7,10 +7,11 @@
     <value nick="minimize-current" value="3" />
     <value nick="open-launcher" value="4" />
     <value nick="custom-command" value="5" />
-    <value nick="window-overview-all" value="6" />
-    <value nick="switch-to-workspace-previous" value="7" />
-    <value nick="switch-to-workspace-next" value="8" />
-    <value nick="switch-to-workspace-last" value="9" />
+    <value nick="window-overview" value="6" />
+    <value nick="window-overview-all" value="7" />
+    <value nick="switch-to-workspace-previous" value="8" />
+    <value nick="switch-to-workspace-next" value="9" />
+    <value nick="switch-to-workspace-last" value="10" />
   </enum>
   <enum id="GalaWindowOverviewType">
     <value nick='grid' value='0'/>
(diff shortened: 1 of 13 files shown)


data/gala.gschema.xml
```diff
@@ -69,6 +70,11 @@
         followed by a ':' and then the command. ';;' is used as delimiter between command definitions.
       </description>
     </key>
+    <key type="b" name="disable-hotcorners-in-fullscreen">
```
danirabbit, line 73: Can we make this enabled instead of disabled? I think for the setting in the UI we probably want to bind to a switch and it would be kind of confusing logic to turn a switch on to turn the feature off, so we'd want to do an inverse bind 😅

danirabbit (changes requested):

src/HotCorners/HotCornerManager.vala
```diff
@@ -71,6 +71,13 @@ public class Gala.HotCornerManager : Object {
         var hot_corner = new HotCorner (display, (int) x, (int) y, scale, hot_corner_position);
 
         hot_corner.trigger.connect (() => {
+            if (
+                display.get_monitor_in_fullscreen (display.get_primary_monitor ()) &&
+                !behavior_settings.get_boolean ("enaable-hotcorners-in-fullscreen")
```
lenemter, line 76:
```suggestion
                !behavior_settings.get_boolean ("enable-hotcorners-in-fullscreen")
```

data/gala.gschema.xml
```diff
@@ -69,6 +70,11 @@
         followed by a ':' and then the command. ';;' is used as delimiter between command definitions.
       </description>
     </key>
+    <key type="b" name="enable-hotcorners-in-fullscreen">
```
danirabbit, line 73: Ah sorry one more thing, can we add this to the `io.elementary.desktop.wm` namespace instead of to `org.pantheon.desktop.gala`? At some point I think we'll want to break this so that we're namespacing all the settings correctly, but at least a new setting it would be nice to have in the new namespace so we don't have to break it later too

danirabbit (changes requested):
Works as expected, just one more minor comment sorry!

data/gala.gschema.xml
lenemter: Done

tintou:
I wonder, do we need a setting here? Shouldn't we disable hotcorners in fullscreen in all cases?

lenemter:
@tintou I think some people still want to use hotcorners while watching videos in fullscreen or doing some other non gaming stuff

danirabbit (approved):
Let's default to `false` and see if anyone complains and if we don't get any feedback we can remove the setting in a future version. This seems like a pretty non-invasive option so shouldn't be hard to change later if we want

mirzaindra:
i just installed elementary OS 8, and i noticed hot screen no longer works in full screen app, how can i fix this? i need hot screen to work in full screen app. thank you

leolost2605:
If you want to do it graphically you can install dconf editor (just search in appcenter) and navigate to `/io/elementary/desktop/wm/behavior/` where you can toggle `enable-hot-corners-in-fullscreen`.

Alternatively just run `gsettings set io.elementary.desktop.wm.behavior enable-hotcorners-in-fullscreen true` in the terminal.

PR #1809, opened 2023-12-17
BackgroundManager: Don't position background actors

lenemter:
Branch based on https://github.com/elementary/gala/pull/1808

This code seems to be unnecessary.

### DIFF
--- a/src/Background/BackgroundManager.vala
+++ b/src/Background/BackgroundManager.vala
@@ -1,169 +1,157 @@
-//
-//  Copyright (C) 2014 Tom Beckmann
-//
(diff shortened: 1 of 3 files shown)


danirabbit:
It looks like this was introduced in 2014 as part of supporting the mutter version at that time, so it's very possible it's not needed anymore. It looks like there is an instance of `BackgroundManager` in src/Background/BackgroundContainer.vala that wasn't setting `control_position` to false though

danirabbit (approved):
I'm not seeing any issues here though, so let's go! Easy enough to revert if a problem appears

Issue #1796, opened 2023-10-30
Blur wallpaper behind multitasking view


danirabbit:
### Problem

The grey background behind multitasking view is fine. No real problem 

### Proposal

Instead of the grey background, blur the current wallpaper behind the workspace card. Rule of cool 🤷🏻‍♀️ 

### Prior Art (Optional)

Incredibly popular GNOME shell extension that does this: https://extensions.gnome.org/extension/3193/blur-my-shell/

PR #1810, opened 2023-12-17
MultitaskingView: Add blurred wallpaper

lenemter:
Fixes https://github.com/elementary/gala/issues/1796


danirabbit:
Very cool! Is there a reason we didn't use the Clutter blur effect and implemented our own?

I submitted a branch here to add a brightness effect for some extra contrast, thoughts? https://github.com/elementary/gala/pull/1811

I'm getting a consistent crash on logging into the session and entering the multitasking view. If I switch workspaces first it doesn't crash

lenemter:
> Very cool! Is there a reason we didn't use the Clutter blur effect and implemented our own?

I think it uses a simple shader with a small blur radius and there's no way to change it.

> I'm getting a consistent crash on logging into the session and entering the multitasking view. If I switch workspaces first it doesn't crash

Can you provide the backtrace please? It doesn't crash on my side.

lenemter:
@danirabbit I actually found a typo I made that made it crash with certain blur radius. I hope it fixes the crash for you too

danirabbit (approved):
LFG!

PR #1811, opened 2023-12-18
Add brightness effect for contrast

danirabbit:
Adds a brightness effect based on dark or light mode to add some extra contrast between the background and the workspace

Screenshot from 2023-12-18 13 32 52
Screenshot from 2023-12-18 13 32 38


lenemter (approved):
Looks great. Thanks!

PR #1812, opened 2023-12-24
Show notifications on the left in RTL languages

### LINKED ISSUE #765: The notifications appear on the right side, they should be in Arabic on the left
The notifications appear on the right side, they should be in Arabic on the left

lenemter:
Fixes #765


danirabbit (approved):
I can confirm this works as expected :)

PR #1724, opened 2023-07-12
Use sound indicator to handle volume change

lenemter:

PR #1816, opened 2023-12-30
Handle sound confirmations in the wingpanel indicator

lenemter:


Marukesu:
The way i would go with this is disabling the `volume-*` keybindings in GSD, and handling they in the sound indicator.

We already expose the `org.gnome.Shell.GrabAccelerators` method in gala, there's no need for creating a new DBus interface.

lenemter:
@Marukesu there is https://github.com/elementary/gala/pull/1724 with your proposal. I'd like to revisit it later separately.

> We already expose the org.gnome.Shell.GrabAccelerators method in gala, there's no need for creating a new DBus interface.

Yes, but we need to trigger sound change feedback in the indicator, so we can't use `org.gnome.Shell.GrabAccelerators` since it's exposed by gala, not the indicator. Or you meant something different?

Marukesu:
> > We already expose the org.gnome.Shell.GrabAccelerators method in gala, there's no need for creating a new DBus interface.
> 
> Yes, but we need to trigger sound change feedback in the indicator, so we can't use `org.gnome.Shell.GrabAccelerators` since it's exposed by gala, not the indicator. Or you meant something different?

I mean, make the indicator call org.gnome.Shell.GrabAccelerators, and handle both the volume change and the feedback, so instead of this:
```
  GSD -> GrabAccelerators -> gala -> AcceleratorActivated -> GSD -> ShowOSD -> gala -> HandleOSD -> sound indicator.
```
we have this:
```
  sound indicator -> GrabAccelerators -> gala -> AcceleratorActivated -> sound indicator
```

lenemter:
Closing in favour of https://github.com/elementary/wingpanel-indicator-sound/pull/270.

Issue #1377, opened 2022-02-04
API for launching or switching a particular app ID


danirabbit:
### Problem

For the new Dock, we can't handle any window management functions to be compatible with Wayland. However, in our recent UI study we found that users expect to be able to switch to a running app and switch between windows of running apps in the dock.

### Proposal

To facilitate these features, we need a need API in Gala to handle launching and switching since the dock isn't aware of open windows etc. The simplest way forward here would be a dumb API from the Dock's perspective where we only send an app ID that we would like Gala to act on.

If there are no open windows associated with that app ID, Gala should launch the appinfo associated with that ID.

If there is one open window associated with that app ID, Gala should switch to that window, across workspaces if necessary.

If there are multiple open windows associated with that app ID, Gala should show them in a window spread (like the multitasking view)

### Prior Art (Optional)

I'm fairly certain this is how the Ubuntu Unity launcher worked, but I don't have any evidence of that on hand

PR #1817, opened 2024-01-02
Allow focusing a window via DesktopIntegration

leolost2605:
Adds a dbus method that focuses the window with the given uid.
Is there a better way to find a window by uid instead of iterating over all windows (I'm not too familiar with gala/mutter methods)?

Goes with elementary/dock#196


danirabbit:
I haven't tested this yet but just wanna reference #1377 which was the design we were looking at based on survey results

leolost2605:
Would a window list with window titles in the context menu still be part of the design? I think for this and for cycling through windows by scrolling over the icon, this method here would still be useful. Also since we now have the DesktopIntegration we know quite a lot about the windows of applications so I wonder whether that was a limiting factor for the design?
Or should the window spread be the only way to cycle through windows?

leolost2605:
@danirabbit should we go forward on this? Or drop the window list from the context menu entirely? I personally don't use it but it's not a lot to maintain so we might as well keep it 🤷

lenemter (approved):

danirabbit:
@leolost2605 sorry for taking so long to follow up on this. I believe the scrolling behavior is at least a somewhat popular feature so that makes sense to keep imo. I think I'd rather hold off on adding a list of windows to the menu until we get some feedback that indicates this is a useful feature for people. The feedback we've received in the past about people wanting to see a preview of the actual window in order to choose leads me to believe that the list isn't useful

PR #1820, opened 2024-01-05
Struts with pantheon protocol

leolost2605:
This PR adds a `PanelManager` that handles struts for panel windows and in the future possibly setting the anchor for a panel. The struts are already implemented and working, set_anchor is still waiting for a mutter MR.

Regarding the actual API it can be easily changed. Currently it's `make_exclusive` and `unmake_exclusive` to closest resemble what's actually happening but it can be changed to something like` set_exclusive (bool exclusive)` or KDE's way of setting a flag how the panel should behave.

Opinions on this? :)

This PR is based on #1705 (the PR is currently against that branch to reduce the diff)


src/PantheonShell.vala
```diff
@@ -243,7 +247,55 @@ namespace Gala {
             return;
         }
 
-        // TODO
+        Meta.Side anchor_side = TOP;
+        switch (anchor) {
```
leolost2605, line 251: Not sure whether we can just cast this or whether mutters enum order is likely to change

Marukesu:
Is there a use case for changing the exclusive are after creating the panel surface?

I believe both wingpanel and dock won't change it after set the first time. so making exclusive a parameter on get_panel() should be enough.

leolost2605:
I thought about changing dock behavior from auto hide to always show and back 🤷

Issue #1823, opened 2024-01-08
API for display labels


danirabbit:
### Problem

In System Settings we have colored display labels that we set while the display configuration page is open. However we won't be able to position those windows in GTK 4/Wayland

### Proposal

Some kind of API either to position those windows as with the Dock and Panel

OR

Draw these labels in Gala and a simple API to show/hide them

### Prior Art (Optional)

https://github.com/elementary/switchboard-plug-display/blob/master/src/Widgets/DisplayWindow.vala

Marukesu:
GNOME Shell has `org.gnome.Shell.ShowMonitorLabels` and `org.gnome.Shell.HideMonitorLabels` DBus methods. We can implement those on our version of the Shell interface.

I think the labels are simply enough to be written in clutter. otherwise we could move them to gala-daemon and use Meta.WaylandClient for positioning.

PR #1827, opened 2024-01-16
API for display labels

lenemter:
Fixes https://github.com/elementary/gala/issues/1823


src/Widgets/MonitorLabel.vala
```diff
@@ -0,0 +1,103 @@
...
+                create_components ();
+            }
+        });
+
+        canvas.draw.connect (draw);
+    }
+
+    private bool draw (Cairo.Context ctx, int width, int height) {
+        if (style_context == null) { // gtk is not initialized yet
+            create_gtk_objects ();
```
Marukesu, line 52: if we are gonna to use gtk here, isn't better to use the daemon to create the labels? we should be reducing our usage of Gtk from inside the gala process, not increasing it.

PR #1830, opened 2024-01-17
switch_to_next_workspace: require timestamp

lenemter:


tintou:
As a side note, we should replace all our `Gtk.get_current_event_time` with `Clutter.get_current_event_time`

tintou (approved):

PR #1833, opened 2024-01-20
Add window spread for an app

### LINKED ISSUE #1377: API for launching or switching a particular app ID

leolost2605:
This is going to be used by dock when more than one window is open for an app and the app icon is clicked on.

Uses the WindowOverview but with only the windows that belong to the specific app. Once the dock is properly integrated it will stay as well and don't go away.

Also opinions in general on this? An alternative would be something more classical like the alt tab switcher but with window previews over the dock icon but I like this a bit more and it is as far as I understood it what's meant in #1377.

Mostly fixes #1377 (probably together with #1817, alternatively that can be absorbed here with this handling deciding between focus for one or spread for two or more but I think a window list in the context menu would still be useful in the dock)

### DIFF
--- a/src/Widgets/WindowOverview.vala
+++ b/src/Widgets/WindowOverview.vala
@@ -16,7 +16,6 @@ public class Gala.WindowOverview : Clutter.Actor, ActivatableComponent {
 
     // the workspaces which we expose right now
     private List<Meta.Workspace> workspaces;
-    private List<Meta.Window> minimized_windows;
 
     public WindowOverview (WindowManager wm) {
         Object (wm : wm);
@@ -92,19 +91,12 @@ public class Gala.WindowOverview : Clutter.Actor, ActivatableComponent {
 
                 if (window.window_type != Meta.WindowType.NORMAL &&
                     window.window_type != Meta.WindowType.DIALOG ||
-                    window.is_attached_dialog ()
+                    window.is_attached_dialog () ||
+                    (window_ids != null && !(window.get_id () in window_ids))
                 ) {
                     unowned var actor = (Meta.WindowActor) window.get_compositor_private ();
                     actor.hide ();
 
-                    continue;
-                }
-
-                if (window_ids != null && !(window.get_id () in window_ids)) {
-                    if (!window.minimized) {
-                        window.minimize ();
-                        minimized_windows.append (window);
-                    }
                     continue;
                 }
 
@@ -286,10 +278,6 @@ public class Gala.WindowOverview : Clutter.Actor, ActivatableComponent {
 
         foreach (unowned var child in get_children ()) {
             ((WindowCloneContainer) child).close ();
-        }
-
-        foreach (var window in minimized_windows) {
-            window.unminimize ();
         }
 
         Clutter.Threads.Timeout.add (MultitaskingView.ANIMATION_DURATION, () => {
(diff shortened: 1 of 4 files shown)


leolost2605:
Currently it minimizes and then unminimizes other windows but since that's done while activating the overview it can be a bit overwhelming and look a bit flickery. Another option would be to just hide which doesn't look too bad and maybe even a bit better? Opinions/any other ideas @elementary/ux ?

With just hiding:

[With just hiding](https://github.com/elementary/gala/assets/106322251/1842ac17-5021-42f3-9c10-4d485631e4c5)

With minimizing/unminizing:

[With minimizing/unminizing](https://github.com/elementary/gala/assets/106322251/aadbaa18-5cf3-40c3-be27-7b4a75f2fda6)

leolost2605:
For testing one can use elementary/dock#210

danirabbit:
Yeah I agree the minimize animation is heavy. Is there no animation on hide? I'm sure it's outside of the scope of this branch but perhaps we should make sure there's a crossfade or something for hide and maybe speed up this window spread animation a little bit. But otherwise, this is great!

leolost2605:
> perhaps we should make sure there's a crossfade or something for hide

Should be easy enough but for some reason the WindowActors become just grey rectangles instead of going transparent when setting opacity to 0. Not quite sure what to do here 🤷

danirabbit (approved):
Approving on behalf of UX but requested Leo for code review since he's more active in this repo than I am

src/Widgets/WindowOverview.vala
```diff
@@ -85,7 +90,9 @@ public class Gala.WindowOverview : Clutter.Actor, ActivatableComponent {
 
                 if (window.window_type != Meta.WindowType.NORMAL &&
                     window.window_type != Meta.WindowType.DIALOG ||
-                    window.is_attached_dialog ()) {
+                    window.is_attached_dialog () ||
+                    (window_ids != null && !(window.get_id () in window_ids))
```
lenemter, line 94: I think you can remove `window_ids != null` check if you set initialize window_ids to be {} by default.

src/WindowManager.vala
```diff
@@ -2427,6 +2427,14 @@ namespace Gala {
...
             actor.unref ();
         }
+
+        public void show_window_spread (HashTable<string,Variant>? hints = null) {
+            if (window_overview.is_opened ()) {
+                window_overview.close ();
+            }
+
+            window_overview.open (hints);
+        }
```
lenemter, line 2437: Maybe better would be to make `window_overview` public, and move this method to the DesktopIntegration?

leolost2605: Right thank you!

src/Widgets/WindowOverview.vala
leolost2605: I'm currently using the check here as an indicator whether to show all windows or only certain windows

lenemter (approved):
LGTM

PR #1837, opened 2024-01-21
wayland: Launch daemon as WaylandClient

leolost2605:
On wayland launches the daemon as a wayland client. For testing this disables the autostart in the desktop file. 
It will now position the menus correctly.

Todo (in this branch or follow up PRs):

- [ ] Menus don't close on outside clicks. (Push modal with the WindowActor makes it not receive any inputs itself.)
- [ ] Find a way to launch daemon on X only. Alternatively we can also launch it ourselves.
- [ ] Probably remove the session stuff from the daemon?


leolost2605:
#1844

PR #1840, opened 2024-01-22
Wayland daemon display labels

### LINKED ISSUE #1823: API for display labels

### LINKED ISSUE #1826: Missing gala-daemon systemd service
### What Happened?

I am trying to [drop `--builtin`](https://github.com/elementary/session-settings/blob/3476c89bbb66564a72c6495ac0c61f8f9ed7a3ec/xsessions/pantheon.desktop#L4) from session-settings and gala-daemon no longer autostarts.

`gala-daemon.desktop` was set hidden under systemd:

https://github.com/elementary/gala/blob/167862ae223b3483336ad3ed45bc2326ac3c167a/data/gala-daemon.desktop#L13

***P.s.:** https://github.com/elementary/session-settings/issues/17#issuecomment-1562589852 is now part of gnome-session 46.alpha.* /cc https://github.com/NixOS/nixpkgs/pull/140429

### Steps to Reproduce

[Drop `--builtin`](https://github.com/elementary/session-settings/blob/3476c89bbb66564a72c6495ac0c61f8f9ed7a3ec/xsessions/pantheon.desktop#L4), open switchboard, right click the titlebar (which is why gala-daemon was introduced: https://github.com/elementary/gala/pull/179)

### Expected Behavior

I imagine we add services and targets here and add it as a [required components](https://github.com/elementary/session-settings/blob/3476c89bbb66564a72c6495ac0c61f8f9ed7a3ec/systemd/meson.build#L14) in session-settings like other GSD stuff

### OS Version

NixOS

### Software Version

leolost2605:
I had this wayland-daemon branch lying around and thought I might as well put it to use. 

This removes the gala-daemon autostart and instead starts it as a wayland client on wayland or as a normal subprocess on x. It therefore also removes all the session stuff (not sure whether we have to do somewhere else something else too). 

Furthermore it implements an API to show MonitorLabels which is pretty much taken from #1827 and from the original displaylabel in settings.
I've hardcoded the CSS and didn't gresource it since we need some hardcoded css anyways. If it should still be gresourced where would be the best place to put it? Normal `data` folder?
I'm not sure since what mutter version `Meta.Context` is required for `WaylandClient` creation so I guessed 43 which would fit with the commit date. Is there a good way to find this version?

This is an alternative to #1827 and works on both X and wayland.

Closes #1827 
Fixes #1823
Probably fixes #1826


leolost2605:
For testing one can use elementary/switchboard-plug-display#372

leolost2605:
> Given that [mutter!2718](https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/2718) is not in 43, do you mind replacing the `HAS_MUTTER43` in `src/DaemonManager.vala` with `HAS_MUTTER44` since NixOS pins Mutter to 43 on Pantheon and we do need that to get this build 🙇

Sure seems I guessed wrong then 😅

danirabbit:
@lenemter Can you take a look at this? I think these labels are currently the biggest blocker for this gtk4 port :)

leolost2605:
I would actually kinda prefer to go with #1844 first and then I'll push a new branch with display labels based on that one since it's quite a bit better than this one and with less changes. The bigger diff there is just because I moved both menus into separate files.

lenemter:
@leolost2605 resolve conflicts here please

leolost2605:
See #1861

PR #1844, opened 2024-02-06
Wayland menus

### LINKED ISSUE #1826: Missing gala-daemon systemd service

leolost2605:
This removes the gala-daemon autostart and instead starts it as a wayland client on wayland or as a normal subprocess on x. It therefore also removes all the session stuff (not sure whether we have to do somewhere else something else too). 

It spawns a transparent window that covers the whole display and makes it responsible for positioning the menus.
This is done in preparation for the wayland transition.

This currently fully works under X but doesn't work under wayland. The reason for that is that on wayland gtk always closes menus automatically/doesn't even spawn them if they are opened without an event. However this would/will be an easy fix with popovers but for now they can't be used as context menus because in gtk3 they always draw an arrow. So I would argue this doesn't block since positioning etc. on wayland works and there is no change in x behavior

Hide whitespaces might help for reviewing since some namespaces were changed to inline
Probably fixes #1826


leolost2605:
This currently fully works under X but doesn't work under wayland. The reason for that is that on wayland gtk always closes menus automatically/doesn't even spawn them if they are opened without an event. However this will be an easy fix with popovers but they can't be used as context menu in gtk3 because they always draw an arrow. So I would argue this doesn't block since positioning etc. on wayland works and there is no change in x behavior

danirabbit (commented):
Don't forget to add the new BackgroundMenu and WindowMenu files to POTFILES :)

danirabbit:
afaict this works exactly the same on X11. My wayland session is crashing right now so couldn't test that :sweat_smile: Would love an ack from @tintou or @lenemter so we can move forward :)

daemon/MenuDaemon.vala
```diff
@@ -15,353 +15,103 @@
...
-        private static GLib.Settings keybind_settings;
-        private static GLib.Settings gala_keybind_settings;
+    construct {
+        Bus.watch_name (BusType.SESSION, DBUS_NAME, BusNameWatcherFlags.NONE, gala_appeared, lost_gala);
+    }
 
-        static construct {
-            keybind_settings = new GLib.Settings ("org.gnome.desktop.wm.keybindings");
-            gala_keybind_settings = new GLib.Settings ("org.pantheon.desktop.gala.keybindings");
+    private void on_gala_get (GLib.Object? o, GLib.AsyncResult? res) {
```
lenemter, line 40:
```suggestion
    private void on_gala_get (GLib.Object? obj, GLib.AsyncResult? res) {
```

daemon/MenuDaemon.vala
```diff
@@ -15,353 +15,103 @@
...
 //
 
-namespace Gala {
+[DBus (name = "org.pantheon.gala")]
+public interface Gala.WMDBus : GLib.Object {
+    public abstract void perform_action (Gala.ActionType type) throws DBusError, IOError;
+}
+
+[DBus (name = "org.pantheon.gala.daemon")]
+public class Gala.Daemon.MenuDaemon : Object {
```
lenemter, line 24: Interface above uses explicit GLib namespace. I'd like to be consistent here.

```suggestion
public class Gala.Daemon.MenuDaemon : GLib.Object {
```

daemon/WindowMenu.vala
```diff
@@ -0,0 +1,194 @@
+/*
+ * Copyright 2023 elementary, Inc. (https://elementary.io)
```
lenemter, line 2:
```suggestion
 * Copyright 2024 elementary, Inc. (https://elementary.io)
```

daemon/BackgroundMenu.vala
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright 2023 elementary, Inc. (https://elementary.io)
```
lenemter, line 2:
```suggestion
 * Copyright 2024 elementary, Inc. (https://elementary.io)
```

daemon/Main.vala
lenemter, line 1: Please update copyright header here as well :)

daemon/MenuDaemon.vala
lenemter, line 1: And here

daemon/Window.vala
```diff
@@ -0,0 +1,55 @@
+/*
+ * Copyright 2023 elementary, Inc. (https://elementary.io)
```
lenemter, line 2:
```suggestion
 * Copyright 2024 elementary, Inc. (https://elementary.io)
```

data/gala-daemon.css
lenemter, line 1: I think this file needs a copyright too.

src/DaemonManager.vala
```diff
@@ -0,0 +1,137 @@
+/*
+ * Copyright 2024 elementary, Inc. (https://elementary.io)
+ * SPDX-License-Identifier: LGPL-3.0-or-later
+ *
+ * Authored by: Leonhard Kargl <leo.kargl@proton.me>
+ */
+
+public class Gala.DaemonManager : Object {
```
lenemter, line 8:
```suggestion
public class Gala.DaemonManager : GLib.Object {
```

src/DaemonManager.vala
```diff
@@ -0,0 +1,137 @@
...
+    [DBus (name = "org.pantheon.gala.daemon")]
+    public interface Daemon: GLib.Object {
+        public abstract async void show_window_menu (WindowFlags flags, int width, int height, int x, int y) throws Error;
+        public abstract async void show_desktop_menu (int display_width, int display_height, int x, int y) throws Error;
+    }
+
+    public Meta.Display display { get; construct; }
+
+    private Meta.WaylandClient daemon_client;
+
```
lenemter, line 22:
```suggestion
```

src/DaemonManager.vala
```diff
@@ -0,0 +1,137 @@
...
+                start_x.begin ();
+                return Source.REMOVE;
+            });
+        } catch (Error e) {
+            warning ("Failed to create daemon subprocess with x: %s", e.message);
+        }
+    }
+
+    private void handle_daemon_window (Meta.Window window) {
+        warning ("HAndle daemon window");
```
lenemter, line 87: You forgot to remove this :)

src/WindowManager.vala
```diff
@@ -367,6 +343,8 @@ namespace Gala {
 
                 plugin_manager.load_waiting_plugins ();
 
+                daemon_manager.start ();
```
lenemter, line 346: Any reason it's here and not in `start`? I think it would make sense to create DaemonManager and start it in a single function.

lenemter (changes requested):
I haven't tested it with multiple monitors (yet) but still it looks very good. I left some code suggestions, after fixing them I'll happily approve this branch :)

src/WindowManager.vala
leolost2605: I'm not sure I think it's a remnant from an earlier version I had but I changed it to start in construct and it still works 🤷

lenemter (approved):
Code looks good.

There is one thing that bothers me, you can switch workspaces while the menu is open. We should probably close the menu or disable workspace switch while the menu is open. This should be fixed later.

Issue #1815, opened 2023-12-26
Grey lines on primary monitor top and bottom


leolost2605:
### What Happened?

My primary monitor has two grey strips on top and bottom both on desktop and multitasking view.

Desktop:
![Screenshot from 2023-12-26 18 10 15](https://github.com/elementary/gala/assets/106322251/9c3eec9e-e35b-467c-9966-6ff750706016)

Multitasking view:
![Screenshot from 2023-12-26 17 42 13](https://github.com/elementary/gala/assets/106322251/6a5892e7-5285-48ae-bdda-29566ebdc1bc)

### Steps to Reproduce

1. Compile gala from master
2. Restart gala

### Expected Behavior

No rounded corners and no grey strips

### OS Version

7.x (Horus)

### Software Version

Compiled from git

### Hardware Info

Dual monitor setup

Nouveau nvidia drivers

leolost2605:
I just realized that the desktop on the secondary monitor also has rounded corners

kerunaru:
I think those rounded corners on every display are intended... Aren't they, @danirabbit ?

lenemter:
@kerunaru the issue is about gray lines on the top and bottom.

kerunaru:
@lenemter, I supposed that but I've thrown the question just in case. The original issue description doesn't make clear what is the complaint: the grey lines, the rounded corners or both :sweat_smile:

leolost2605:
Oh i didn't know the rounded corners on desktop are intentional 😅
I just thought them connected to the strips because of the recent introduction of rounded corners on multitasking view but I'll change the issue description

leolost2605:
@lenemter I've got really no idea what any of this is doing but removing [`content.set_rounded_clip_bounds (rect);`](https://github.com/elementary/gala/blob/c62024acd20c2b82bffac960e9239aea9624eb3a/src/Background/BackgroundManager.vala#L127) makes the grey lines go away and I've still got rounded corners in the multitasking view 🤷

PR #1845, opened 2024-02-08
Fix rounded background

### LINKED ISSUE #1841: Changing wallpaper makes the background go grey
### What Happened?

When changing wallpaper from settings the background goes gray with some fading. The multitasking view shows the correct wallpaper. Logging in and out fixes the issue.
Also if you play around with it a little and change wallpaper then immediately open multitasking view funny things will happen. For example now the normal desktop background has the correct wallpaper while the multitasking view is grey and sometimes even offset...?

### Steps to Reproduce

1. (Probably) compile gala from latest master
2. Open settings
3. Change wallpaper
4. Observe a grey background

### Expected Behavior

Show the new wallpaper without logging in and out

### OS Version

7.x (Horus)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

leolost2605:
Fixes #1841
Closes #1843 

Ok I did some digging so here is what I found and what I think is happening:
When clip bounds aren't set the full texture size is used as clip bounds. But when the texture isn't loaded yet the size is seen as 0 so clip bounds are set to 0 and the whole texture is clipped away because clip bounds aren't updated when the texture is loaded.

Anyways regardless of whether that's actually what's happening this PR fixes the issue for me without reintroducing #1815 :)

### DIFF
--- a/src/Background/BackgroundManager.vala
+++ b/src/Background/BackgroundManager.vala
@@ -89,6 +89,7 @@ public class Gala.BackgroundManager : Meta.BackgroundGroup {
         var background = new_content.background.get_data<unowned Background> ("delegate");
 
         if (background.is_loaded) {
+            new_content.rounded_clip_radius = Utils.scale_to_int (6, wm.get_display ().get_monitor_scale (monitor_index));
             swap_background_actor (animate);
             return;
         }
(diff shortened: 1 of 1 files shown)


danirabbit (approved):
Works for me. Nice fix! Makes sense

Issue #26, opened 2017-06-23
Workspaces on multiple monitors [Wishlist]


jackbow:
Currently when using multiple monitors and the workspace is switched, only the main monitor will switch to a new workspace, and the other monitors will keep their current windows. I think it would be useful to be able to toggle this behavior per monitor, perhaps in the system settings under 'display'.

cassidyjames:
iirc the reason we can't do this naively is because of Mutter. There has been interest in per-display workspaces expressed several times when talking to GNOME and Mutter folks, though.

PR #1846, opened 2024-02-08
Remove corner mask in favor of rounded background corners

leolost2605:
By making the system background entirely black we can remove the maskcorners plugin since we already round the background corners.

The only difference is that now windows or anything that's in the corner doesn't get rounded so I guess that's a design question 🤷


danirabbit (approved):
So the only place you notice this afaict is like fullscreen apps. Which I'm fine with fullscreen apps figuring out their own corners. This looks good to me! I'm happy if @lenemter is happy :)

lenemter (changes requested):
You also need to deprecate the gsettings keys the plugin used

leolost2605:
@lenemter done!

lenemter:
For some reason windows are still getting clipped at corners and corners stay black even if I change background color to the old gray one. So nothing has changed for me.

@leolost2605 Could you revert the change to the background color please? The color is still shown when switching workspaces and black looks bad here.

leolost2605:
> For some reason windows are still getting clipped at corners and corners stay black even if I change background color to the old gray one. So nothing has changed for me.

I think that's because the plugin isn't actually uninstalled when you install the branch so it's still there. I tested it by disabling the plugin 🤷

> @leolost2605 Could you revert the change to the background color please? The color is still shown when switching workspaces and black looks bad here.

Ah right I almost thought I might have missed a use :) Will change it

leolost2605:
@lenemter this should be fixed now :)

lenemter (changes requested):
Nice! Using `background_color` is a good solution but now the black corners are visible during workspace switch animation. To fix this you can change their color to transparent.

leolost2605:
You mean by workspace switch switch via touchpad swipe/super + left/right? 
Because they aren't black for me there 🤔

lenemter:
@leolost2605 Yes

screenshot

It's more obvious if I change system bg to white:

screenshot white

leolost2605:
@lenemter ok so I'm guessing this has something to do with whether you have workspaces on primary only. Btw how do I disable this cause I didn't find it anywhere? Is there a dconf key somewhere?

I pushed your mentioned fix, couldn't test it though. Also if you have a better name for the method just change it I just couldn't figure out a  good name :)

lenemter:
@leolost2605 it's /org/gnome/mutter/workspaces-only-on-primary. I don't have second monitor, so I don't know how it would affect it

lenemter:
@leolost2605 it crashes wm for me

```
(gdb) bt
#0  0x000055e194671399 in gala_background_container_black_background (self=0x55e1959408f0, black=0)
    at ../src/Background/BackgroundContainer.vala:52
#1  0x000055e194659a1a in gala_window_manager_gala_prepare_workspace_switch (self=0x55e19593b250, from=0, to=1, direction=META_MOTION_RIGHT)
    at ../src/WindowManager.vala:1941
#2  0x000055e19465c64a in gala_window_manager_gala_real_switch_workspace (base=0x55e19593b250, from=0, to=1, direction=META_MOTION_RIGHT)
    at ../src/WindowManager.vala:2121
#3  0x00007fc62b2f35d9 in meta_workspace_activate_with_focus (workspace=0x55e195596cd0, focus_this=0x0, timestamp=0)
    at ../src/core/workspace.c:684
#4  0x000055e19464d5c8 in gala_window_manager_gala_real_switch_to_next_workspace
    (base=0x55e19593b250, direction=META_MOTION_RIGHT, timestamp=0) at ../src/WindowManager.vala:594
#5  0x00007fc62c452bf3 in gala_window_manager_switch_to_next_workspace (self=0x55e19593b250, direction=META_MOTION_RIGHT, timestamp=0)
    at ../lib/WindowManager.vala:183
```

leolost2605:
@lenemter IIUC if workspaces_on_primary_only is enabled the workspace switch only uses the backgroundmanager of the primary/only monitor. But if it's disabled it uses the whole backgroundcontainer (where the black background is set) no matter whether you have one or multiple monitors.

Also the crash should be fixed now

src/Background/BackgroundContainer.vala
```diff
@@ -47,6 +48,10 @@ namespace Gala {
             monitor_manager.monitors_changed.disconnect (update);
         }
 
+        public void black_background (bool black) {
```
lenemter, line 51: Method names usually use verbs. Rename this to `set_black_background` please :)

lenemter (commented):
Looks very good. One last comment

leolost2605:
Done!

Btw is workspaces on all monitors something we might want to support in the future? Because currently it's entirely broken with multiple monitors?

lenemter (approved):
:rocket:

lenemter:
@leolost2605 Personally, I'd like to add support for https://github.com/elementary/gala/issues/26 when/if mutter implements this, and don't find workspaces on all monitors useful. Maybe @danirabbit has an opinion on this?

PR #1848, opened 2024-02-10
Rework close button

lenemter:
Adds a `CloseButton` class with a simple scale animation + `triggered` signal to avoid using `ClickAction`


danirabbit (approved):
LGTM :rocket:

PR #1849, opened 2024-02-10
WindowCloneContainer: When no window is active, select active window based on direction during keyboard nav

leolost2605:
Currently when no window has visible keyboard focus it could happen that either nothing happened when pressing arrow keys or that a window you wouldn't expect is highlighted.

Now when no window has visible keyboard focus select the top left window when UP or LEFT is pressed and the bottom right window when DOWN or RIGHT is pressed.


src/Widgets/WindowCloneContainer.vala
```diff
@@ -254,8 +253,30 @@ namespace Gala {
...
+                            current_x = window_rect.x;
+                            current_y = window_rect.y;
+                            closest = (WindowClone) child;
+                        }
+                    } else {
+                        if (window_rect.x >= current_x && window_rect.y >= current_y) {
+                            current_x = window_rect.x;
+                            current_y = window_rect.y;
+                            closest = (WindowClone) child;
+                        }
```
lenemter, line 277: This code actually selects the right bottom window instead of second from top/left.

leolost2605: Yeah that was kinda my intention. Like when you have a menu and press down the top most is selected and when you press up the bottom most is selected. But I guess that's a @elementary/ux question. Also maybe it should be swapped so that right or down selects top left and up or left selects bottom right?

danirabbit:
Hm so my inclination would be to think that keyboard focus should travel from the previously focused window, which seems to be what main does already.

However, I agree that if your previously selected window is in a corner and you press a key that leads towards that corner, we should still indicate keyboard focus.

So, for example if your previously focused window is in the top right and you press up arrow or right arrow, we should probably show the keyboard focus ring on that window (and maybe throw a Gdk.Beep?).

But I think this branch makes keyboard navigation possibly less predictable since it seems like it now always navigates as if keyboard focus is starting in the top left (I think?)

leolost2605:
@danirabbit ok i updated it like that!

lenemter (approved):

PR #1851, opened 2024-02-11
SystemBackground: cleanup

lenemter:
SystemBackground doesn't use wallpaper since https://github.com/elementary/gala/pull/1582

### DIFF
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -155,7 +155,6 @@ namespace Gala {
             unowned Meta.Display display = get_display ();
             display.gl_video_memory_purged.connect (() => {
                 Meta.Background.refresh_all ();
-                SystemBackground.refresh ();
             });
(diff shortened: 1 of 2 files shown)


src/WindowManager.vala
```diff
@@ -155,7 +155,6 @@ namespace Gala {
             unowned Meta.Display display = get_display ();
             display.gl_video_memory_purged.connect (() => {
                 Meta.Background.refresh_all ();
-                SystemBackground.refresh ();
```
danirabbit, line 158: Do we still need this to apply the background color again?

lenemter: `Meta.Background.refresh_all ()` should do it. We had a separate `refresh ()` function because previously there was a noisy texture used in the system background.

danirabbit (approved):

PR #1854, opened 2024-02-13
Fix documentation

leolost2605:
- Add missing packages
- Fix syntax errors


danirabbit (approved):
:rocket: thanks for fixing this!

PR #1855, opened 2024-02-13
Add Gala.Barrier

leolost2605:
Takes the barrier from the hot corners, makes it an actual public class and deals with pressure revealing there.

Will be useful to allow pressure reveal for the dock.

I'm not sure which of the constants should be actual "global" constants and which shouldn't. I think at least trigger_pressure_threshold can't be global because current dock pressure is way higher than hot corner pressure.


danirabbit:
@leolost2605 this makes sense to me. Can you resolve conflicts?

leolost2605:
@danirabbit done!

danirabbit (approved):

Issue #1839, opened 2024-01-22
Remove the usage of GTK inside the compositor


tintou:
GTK can't be used inside the compositor and it tried to connect to it before the compositor is available.
As we are only using it for Cairo, the solution is to use Cairo directly here (or use shaders etc. and get rid of Cairo too)

leolost2605:
AFAIK all usages of GTK have been removed :partying_face:

PR #1857, opened 2024-02-17
Add ColorManager for theme accent color

### LINKED ISSUE #1674: Get accent from settings portal

leolost2605:
- Gets the theme accent color from settings portal
- Add a ColorManager because we are going to need it for more things like background color light and dark, tooltip color etc. since we currently use gtk.stylecontext for these colors and this is problematic (see #1839)

Fixes #1674 

First step towards #1839


tintou:
While I agree with the implementation, I don't think that we should use the portal as the portal will be available after Gala. For this I think that we should use the elementary AccountService extension

leolost2605:
I updated it to use the AccountsService :)

PR #1859, opened 2024-02-18
Expose BackgroundManager to the library

lenemter:
Allows changing wingpanel color immediately without the need to wait for the wallpaper transition. 

This change was requested in https://github.com/elementary/greeter/pull/698#pullrequestreview-1883840770


danirabbit (approved):
This feels so much faster and more fluid. Great work :rocket:

PR #1860, opened 2024-02-18
WindowManager: Allow making modal without a grab

leolost2605:
- This allows making the WindowManager modal without a stage grab
- Useful if you just want to block keybindings
- Fixes being able to switch workspaces while a window menu is active which was brought up by @lenemter in #1844 
- It's an API break (I think?) but I'm not sure anybody except for us currently uses gala api?
- Will also be useful if we want to introduce a shell modal hint for windows


lenemter (changes requested):
The menu is not clickable at all in this branch.

leolost2605:
Yeah and it doesn't even do what it was supposed to do so closing

PR #1861, opened 2024-02-18
Daemon: Implement monitor labels

### LINKED ISSUE #1823: API for display labels

leolost2605:
Taken from #1840 (was easier than rebase :))

Fixes #1823 

For the switchboard side see elementary/switchboard-plug-display#372


lenemter:
I'm not sure if window title is a good way to do this. I remember https://github.com/elementary/gala/pull/1427 was rejected for this reason. @tintou how should gala identify gala-daemon windows?

leolost2605:
Yeah I wasn't sure either but I think the difference to #1427 is that we already know for sure that it's a daemon window and only differentiate between different types of daemon windows 🤷

tintou:
Note that `handle_daemon_window` won't be called on X11 so we need another wrapper for X11, maybe by just ensuring that `("%" + uint64.FORMAT).format (window.get_pid ()) == subprocess.get_identifier ()`.
I think that this is good enough for a first iteration, ideally the daemon should communicate each type of window with a dedicated DBus signal, but that would probably be better to iterate on it so that we can move faster.

leolost2605:
Since the daemon is still GTK3 I've just used the GTK3 methods for positioning etc. for X11

tintou (approved):

PR #1862, opened 2024-02-21
MonitorLabel: clean up styles, gtk4 prep

danirabbit:
* In latest Gtk4 the only non-deprecated way to add styles from a string is to the whole display, so prepare for that
* Set label margin in CSS
* Remove some unused styles that were for the display plug


leolost2605 (approved):
LGTM

PR #1863, opened 2024-02-21
Wayland notifications

leolost2605:
Launch the notification server as a wayland client

TODO:
- [ ] Cleanup
- [ ] Check whether the notifications server is actually installed
- [ ] Make window type notifcation

### DIFF
--- a/src/ManagedClient.vala
+++ b/src/ManagedClient.vala
@@ -9,7 +9,7 @@ public class Gala.ManagedClient : Object {
     public Meta.Display display { get; construct; }
     public string[] args { get; construct; }
 
-    private Meta.WaylandClient? daemon_client;
+    public Meta.WaylandClient? wayland_client { get; private set; }
 
     public ManagedClient (Meta.Display display, string[] args) {
         Object (display: display, args: args);
@@ -20,7 +20,7 @@ public class Gala.ManagedClient : Object {
             start_wayland.begin ();
 
             display.window_created.connect ((window) => {
-                if (daemon_client != null && daemon_client.owns_window (window)) {
+                if (wayland_client != null && wayland_client.owns_window (window)) {
                     window_created (window);
                 }
             });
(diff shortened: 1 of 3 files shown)


leolost2605:
@tintou since your proposal to make meta_window_set_window_type public was rejected what do you think would be the best way to go here? 
With current meta api we can't mimic notification behavior because AFAIK at least preventing focus stealing will be impossible. Should I propose something like meta_wayland_client_set_window_type? This would be useful for monitor labels as well to make them tooltips because currently they can be moved with super + mouse drag.

tintou:
I think that for the super+mouse drag, one has to implement https://wayland.app/protocols/keyboard-shortcuts-inhibit-unstable-v1. At least that's the only way I currently see (here https://gitlab.gnome.org/GNOME/mutter/-/blob/main/src/core/window.c#L7404 )

tintou:
Another way might be to deal with https://wayland.app/protocols/wayland#wl_surface:request:set_input_region and set it to a zero region?

leolost2605:
Hmm it seems `has_move_func` being false might also prevent moving (https://github.com/GNOME/mutter/blob/11691b107c1659ea1c254792c8ed00680705e43e/src/core/window.c#L7465l). At least that's what's being set by override redirect windows.

Also I think we still need something to prevent a notification window from taking focus. I just kinda don't wanna go to mutter and propose making all kinds of API public so I guess here is what we need:

- A way to prevent a window from being moved
- A way to prevent a window from being influenced by struts
- A way to prevent a window from taking focus

The easiest and least hacky solution would probably still be to use existing type hints either directly or via make_dock, make_notification, etc. so maybe we should propose these methods for wayland clients and see how they are received? I guess I would start by trying and testing @tintou's make_dock branch?

danirabbit:
I'm going to convert this to a draft so it's clear you're not currently waiting on a review here

PR #1867, opened 2024-02-24
ShadowEffect: Dont use Gtk to draw shadows

leolost2605:
Basically reverts #250 

Currently it's a regression since we can't do fancy shadow styles using CSS now. I'll try looking into fully supporting CSS box shadow properties but that might take a while


danirabbit:
We're using CSS to draw borders too, so need to deal with those. First branch for that:

https://github.com/elementary/gala/pull/1918

danirabbit:
@leolost2605 Okay I think this is probably good enough.

Instead of setting shadow spread, we just take in "css class" as a construction argument

Shadows are drawn with a rounded rectangle now and you can set border radius so that works much better for things like the workspace switcher that casts a very small shadow

The more shadow spread there is, the further we offset shadows on the y axis. So in other words if an object is high up on the Z axis, the shadow is cast further away from its origin

It's not 1:1 to shadows in the stylesheet but I don't think anyone will really notice

danirabbit (approved):
I'm happy if you're happy :)

leolost2605:
Let's go!

PR #1869, opened 2024-02-25
Introduce Gala.Text

lenemter:
Clutter.Text object that automatically changes font-name to the system one.

Currently gets incorrect default values from GLib.Settings :(


bilelmoussaoui:
I am planning to move ClutterText to St inside of gnome shell so we can drop cairo/pango dependencies from libmutter. I guess this could be an opportunity to have an implementation inside gala itself.

PR #1871, opened 2024-02-28
Handle data manually

leolost2605:
This fixes the global creation and makes it actually work. I haven't tested the actual use of the protocol yet but it shows up in the global registry.

It's not very elegant since we now have an unused parameter but I don't really know much about writing VAPI so if anybody knows how to actually fix it that would be nice :)


leolost2605:
Aready fixed :)

PR #1873, opened 2024-02-29
Fix dim effect removal on wayland

### LINKED ISSUE #1800: Dim effect isn't removed from parent window in Wayland
### What Happened?

when a modal dialog has a parent, gala applies a dim effect over it window, and remove it after the dialog is closed. when running as a wayland session, the dim effect is kept, however.

### Steps to Reproduce

1. Launch gala as a wayland session.
2. Open terminal with two tabs and runs a TUI application in one.
3. Try to close the tab with the process application running (a modal dialog will appear making terminal be dimmed).
4. Cancel or confirm the dialog choice.

### Expected Behavior

That the dim effect is removed after the dialog be closed.

### OS Version

Early Access

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

leolost2605:
It seems that currently on wayland the window type changes to normal on dialog windows when they are being destroyed so that we don't know they were modal. Therefore instead of checking window type and undimming then we connect to the unmanaged signal and remove the effect when it is emitted. This also cleans up the workarounds we did regarding knowing the transient parent.

Fixes #1800

### DIFF
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -1658,7 +1658,6 @@ namespace Gala {
 
                     break;
                 default:
-                    warning ("DESTROY DEFAULT");
                     destroy_completed (actor);
                     break;
             }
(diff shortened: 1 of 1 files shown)


tintou (approved):
looks good to me

PR #1877, opened 2024-03-13
Move EndSessionDialog here from wingpanel-indicator-session

leolost2605:
This has the advantage that we can make it modal i.e. we dim everything and prevent all clicks behind it.
It should also be wayland compatible.


leolost2605:
Unfortunately we still have to go the route through gala since otherwise gnome-session for some reason doesn't detect that we exist.
I think this is still worth pursuing though since making it modal is IMO pretty cool and this is also wayland compatible (at least I think so, haven't tested it yet :))

danirabbit (changes requested):
I don't seem to have wallpaper with this branch :sweat_smile: 

Also, the shutdown dialog doesn't gain keyboard focus. Keyboard focus remains in whatever background application behind the dim window bit

leolost2605:
Superseded by #1879

PR #1878, opened 2024-03-18
Set `Super + Space` to switch keyboard layout by default

lenemter:
Super + Space is a default shortcut to switch keyboard layouts in almost every other OS. It would be nice if users could use it right away

@danirabbit do we want to do this or leave the manual configuration up to the user?


danirabbit:
Yeah let's go for it and see what kind of feedback we get. I think we probably want to change the "overlay key" shortcut to be the applications menu instead of shortcuts in this case as well

danirabbit (approved):

PR #1879, opened 2024-03-22
Move EndSessionDialog here

leolost2605:
This has the advantage that we can make it modal i.e. we dim everything and prevent all clicks behind it.
It also is wayland compatible. Unfortunately currently the modal part will only work on wayland. It can be made to work on X but that would require some additional effort that - if really wanted - will be left to other PRs. X behavior is therefore unchanged compared to the current end session dialog.

Closes #1877 since the approach taken there had quite a few issues that couldn't be worked around. This approach now uses a clutter actor to block all interaction. This is also more future proof since we can more easily use this for other dialogs that aren't spawned by the daemon, e.g. portal dialogs, etc.


leolost2605:
A modal system dialog currently blocks all shortcuts not sure whether we should (or even can lol) allow things like screenshot, etc.?

leolost2605:
Obsolete since we can now just use the wayland protocol :)

PR #1880, opened 2024-03-22
Introduce a StyleManager

### LINKED ISSUE #1674: Get accent from settings portal

leolost2605:
This is pretty much #1857 and therefore closes #1857

Only a bit refactored and it will now also be used to get the color scheme. This is because currently granite is used to get the color scheme which in turn uses the portal. This caused start time on wayland to be > 20 seconds since we are waiting for the portal. This PR reduces start time to the same as on X so about 1 - 2 seconds.

Fixes #1674

Continues towards #1839 (AFAIK only shadow effect is now still using GTK)


leolost2605:
My laptop just ran out of battery lol 🥲
I still need to do a push then this should be ready

leolost2605:
We could now also drop Granite as a dependency for the gala lib and main gala executable however that would be an API break since Drawing.Color implements Granite SettingsSerializable. Not sure how we handle gala API? Is it supposed to be stable?

danirabbit:
I think the only thing using Gala API is other Pantheon components so should be okay to break imo 🤷🏻‍♀️

leolost2605:
Ok I dropped the granite dependency in libgala and gala but I left the methods that are implementing settingsserializable in Drawing.Color so that getting it working again is only a line change and recompile away. Dropping these can probably be done in follow up PRs IMHO.

danirabbit (approved):
Works as expected and we have tintou's ack on the other PR with his suggestion to use accountsservice so I'm happy with that :rocket:

PR #1881, opened 2024-03-22
WindowSwitcher: Remove Gtk.Settings

leolost2605:
We don't need gtk settings here since we don't use the gtk theme directly anymore.


danirabbit (approved):
Makes sense to me!

PR #1882, opened 2024-03-22
wayland: Launch notifications server as a wayland client

### LINKED ISSUE #1693: Notifications appear in the center of the screen on Wayland

leolost2605:
Introduce a `ShellClientManager` that currently only creates the notificationsclient. However we will need something like this anyways for the wayland protocol since all clients that want to use it (or at least the panel/dock parts of it) have to be launched as wayland clients.

I'll also try to migrate the daemonmanager to `ManagedClient` as well in a future PR.

This also helps to keep the code added to WindowManager as little as possible by taking responsibility from it compared to #1863 since it's already a huge class anyways.

Closes #1863
Fixes #1693 

Going the route of the `NOTIFICATION_DATA_KEY` is a bit of a workaround and once I have some more time i'll try to get something like make_notification merged in upstream mutter however since AFAIK we now are too late for 46 anyways that would be OS 9 material and I think the few lines of code are good enough for now.

Any feedback regarding this approach especially with consideration of the upcoming wayland protocol etc. is much appreciated :)


leolost2605:
Without mutter <46 there is still focus stealing going on however that will be fixed in 46

tintou (approved):
Looks good to me

PR #1884, opened 2024-04-02
DesktopIntegration: Take stacking order into account

leolost2605:
We sort windows by stacking order ~and emit `windows_changed` when the stacking order changes~


danirabbit:
I think what I'm seeing with this branch is if I have 3 windows I sometimes get a case where scrolling will go like:
1 → 2 → 1 → 2 → 3

instead of just
1 → 2 → 3

Issue #275, opened 2018-06-19
Only move windows when switching workspaces


harveycabaguio:
Conceptually, if you think of wingpanel, dock, wallpaper, and windows as separate layers, the only one that needs to move when switching workspaces is the windows layer. Wingpanel + dock are omnipresent and stay the same across workspaces, and the wallpaper doesn't change across workspaces so it feels redundant to see it move as well.

danrabbit:
The wallpaper should move because it’s consistent with multitasking view and how we show that each workspace is represented by a separate wallpaper.

Agree about dock and panel though since we show those as being pulled away in multitasking view

janxkoci:
This would also make sense for current behaviour of wingpanel when changing workspaces where some windows are maximized (I was goint to open a new bug report to suggest a change that fits more with the current state).

Right now if you switch from workspace with maximized window (i.e. black wingpanel) into one without maximized window (i.e. transparent / translucent wingpanel), the wingpanel slides in with the new workspace, while changing it's state - this make no sense as the new workspace should just have the final wingpanel (i.e. transparent / translucent, no fading or change of state).

If you don't move wingpanel with the workspace, then fading to change it's state would make more sense.

PR #1887, opened 2024-04-02
ShellClients: Add PanelClient

leolost2605:
This PR implement everything necessary for the positioning of Panels (currently top and bottom are supported) and their autohiding.

The API that is available (and in the future can probably be accessed via the wayland protocol) is:
`set_anchor (Meta.Window window, Meta.Side side)`
Pins the given window to the given side on the primary monitor and keeps it there centered. The side can be changed on the fly (at least in theory).

`set_hide_mode (Meta.Window window, HideMode hide_mode)`
with hide_mode being one of the following:
```
NEVER // exclusive, i.e. with struts
MAXIMIZED_FOCUS_WINDOW // hide if the maximized focus window overlaps the dock
OVERLAP_FOCUS_WINDOW // hide if the focus window overlaps the dock
OVERLAP_WINDOW // hide if any normal or dialog window overlaps the dock
ALWAYS // always hide
```
Gala will then handle autohiding taking into account hover and any open transient windows (e.g. popovers, menus, dialogs, etc.). It will also show the panel when the barrier at the appropriate side is hit.
This can also be changed on the fly. 

This is enough to replicate the current X experience. It even supports autohiding wingpanel in theory (the example that's currently included here but will be removed sets up wingpanel to always hide and dock to dodge windows (`OVERLAP_WINDOW`).

It doesn't fully work/has some issues if run on mutter < 46

Also this of course only works on wayland


leolost2605:
@elementary/ux what's the opinion on having dock and wingpanel "float" above workspaces when switching via gestures or shortcut without opening multitasking view. I.e. they don't show like they do currently on both workspaces but instead just stay at their position while the workspace switching animates below them. IMHO that might look pretty good especially because then we could do some fancy gesture stuff e.g. when switching from a workspace where dock should be hidden because it's covered to a workspace where it should be shown we could animate it in sync with the gesture. Also with how it's currently we have some inconsistent states where plank shows even though it's overlapped because we are switching etc.
It's not a deal breaker but it would be pretty cool IMHO

danirabbit:
@leolost2605 yes! That actually is an open issue report: https://github.com/elementary/gala/issues/275

leolost2605:
Regarding focus management I'd like some opinions on how we want to do it.
Mutters behavior currently is like this:
https://github.com/GNOME/mutter/blob/8a862b102c2d059ed498eeab5e4c7abb94bc5f1f/doc/mutter-focus.txt#L109
I.e. dock windows only receive keyboard focus if they specifically request it via NET_ACTIVE_WINDOW on X so they won't get any focus on wayland.
Should we replicate this behavior? I.e. add a focus method to the wayland protocol? This would also allow a focus grab without user interaction which might come in handy in some special cases?

FWIW wlr_layer_shell does it like this:
https://wayland.app/protocols/wlr-layer-shell-unstable-v1#zwlr_layer_surface_v1:enum:keyboard_interactivity
(1:1 replication of this might need some upstream work and I'm not 100% sure this will get accepted given the things said in the document linked above)

Or maybe there are suggestions for alternative ways to handle this?

@tintou @Marukesu might have some opinions here?

tintou:
Wingpanel actually requires keyboard focus for the application menu (?)

leolost2605:
Yeah and it also might want to grab it when using super key to open it. Currently it just can't be interacted with on wayland

PR #1894, opened 2024-04-06
WindowManager: Keep docks static during workspace switch

### LINKED ISSUE #275: Only move windows when switching workspaces

leolost2605:
Works both on X (also with plank) and on wayland with #1887.

Fixes #275

### DIFF
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -1966,7 +1966,6 @@ namespace Gala {
 
             var to_has_fullscreened = false;
             var from_has_fullscreened = false;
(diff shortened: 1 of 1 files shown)


danirabbit (approved):
Works as expected. Nice job!

PR #1897, opened 2024-04-28
WindowClone: avoid using ClickAction

lenemter:


tintou:
What is the motivation behind this change?

lenemter:
I think there is no need in it. `DragDropAction` already handles click events, we can just use it instead, no?

PR #1900, opened 2024-04-29
Multitasking view: Display modal dialogs on top of parent window.

### LINKED ISSUE #198: Multitasking should be disabled when an application sends an important message.
If, for example, I have a command running in the terminal and I decide to close the terminal using multitasking, the terminal sends me an important message, this message can not be seen since it is not the focus.

gala error

We could improve the user experience if multitasking is disabled and this message takes focus, obviously only if the message is important.

lenemter:
Closes #198 
Removes a hack to automatically close multitasking view when window refuses to close. This branch displays new modal dialogs on top of a parent window.

Before:
---
Screenshot from 2024-04-29 23 34 19

After:
---
Screenshot from 2024-04-29 23 34 46

To do:
- [x] Fix dnd opacity transition
- [x] Fix duplicate windows in GTK4 demo
- [x] Test with multiple monitors
- [x] Simplify modal checks
- [x] Fix Files "Select Application" position


lenemter:
@leolost2605 Thanks you! I managed to get it working after revering latest commit. And I can confirm that the branch fixes the crash.

danirabbit:
I hope this comes back as a new branch. I really liked this change 🥲

PR #1906, opened 2024-05-24
wayland: Implement shell

### LINKED ISSUE #68: Wayland: design window communication API

### LINKED ISSUE #1365: Replacement for Gtk.Window.keep_above ()
### Problem

According to [this post](https://discourse.gnome.org/t/gtk-4-how-to-replace-gtk-window-set-keep-above-and-gtk-window-set-keep-below/3550) there is no replacement for Gtk.Window.keep_above in GTK4 since not all window managers implement this behavior

### Proposal

Implement an API in Gala or see if there's something in Mutter so that system components can be set as `keep_above`

### Prior Art (Optional)

_No response_

### LINKED ISSUE #1694: Panel and Dock are treated as normal windows on Wayland

leolost2605:
Big PR with #1705, #1887 and hooking both up with some modifications.
Therefore can/should be rebased merged IMHO.

This is all we need for a fully functional wayland shell.
The API we have for the panel is:
```
set_anchor
focus
set_size // If this is not called the size of the surface will be used but this is necessary for wingpanel
set_hide_mode
```
everything else API wise is left untouched from #1705 

Corresponding dock PR: elementary/dock#240

Closes #1892 
Closes #1887 
Closes #1888 
Closes #1705 
Closes #1820 

Fixes #1694 
Fixes #1365 
Fixes #68


src/ShellClients/ShellClientsManager.vala
```diff
@@ -6,15 +6,86 @@
...
-        Object (display: display);
+    private ShellClientsManager (WindowManager wm) {
+        Object (wm: wm);
     }
 
     construct {
-        notifications_client = new NotificationsClient (display);
+        notifications_client = new NotificationsClient (wm.get_display ());
+
+        var clients = shell_settings.get_value ("trusted-clients");
```
tintou, line 39: no need to keep `shell_settings` in memory forever
```
var shell_settings = new Settings ("io.elementary.desktop.wm.shell");
var clients = shell_settings.get_value ("trusted-clients");
```

tintou:
This looks really promising

tintou (approved):

danirabbit:
Congrats and thank you for your hard work on this! 🎉 🍾 🍻 🎈 🥳

PR #1908, opened 2024-05-25
Fix barriers

leolost2605:
Barriers require the backend since a newer mutter version. Also switch to composition instead of inheritance to prevent this from happening again as well as catching errors if barriers aren't supported.


danirabbit (approved):
Nice job :rocket:

PR #1912, opened 2024-05-26
MultitaskingView: Fix dock animation

leolost2605:
Currently on wayland the dock isn't animated out and in


danirabbit (approved):
Works for me!

PR #1913, opened 2024-05-26
InternalUtils: Fix window placement

### LINKED ISSUE #1911: MultiTaskingView: Windows are all over the place sometimes
### What Happened?

Ignore the white borders that's because I disable shadow effects by returning a null texture since that's still using gtk :)

Screenshot from 2024-05-26 14 22 42
At the bottom is vscode

### Steps to Reproduce

AFAIK it's wayland only but that's as far as I got to find a common cause.
It seems to happen more frequently if a window is maximized?

### Expected Behavior

As on X

### OS Version

8.x (Early Access)

### Software Version

Compiled from git

### Log Output

```shell
Nothing relevant AFAICT
```

### Hardware Info

_No response_

leolost2605:
Copying the list would sometimes lead to absolutly crazy numbers for width and height of some TilableWindow rects. I'm lacking the background knowledge as to why this just started happening now or what was actually going on but in my  testing this seems to fix it. Also it seems as if the algorithm is more stable now and windows don't just randomly change place but maybe I'm just imagining things at this point lol :)

Thanks @bobby285271 for your investigations otherwise I would've been looking into wayland stuff for far too long :)

Fixes #1911

### DIFF
--- a/src/InternalUtils.vala
+++ b/src/InternalUtils.vala
@@ -172,9 +172,8 @@ namespace Gala {
             }
 
             // Assign each window to the closest available slot
-            var tmplist = windows.copy ();
-            while (tmplist.length () > 0) {
-                unowned List<unowned TilableWindow?> link = tmplist.nth (0);
+            while (windows.length () > 0) {
+                unowned List<TilableWindow?> link = windows.nth (0);
                 var window = link.data;
                 var rect = window.rect;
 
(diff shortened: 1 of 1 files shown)


danirabbit (approved):
I've seen weird placement issues sometimes in X too. Glad you were able to find a fix. Nice work!

PR #1915, opened 2024-05-26
 ShellClients: Keep clone only destroy HideTracker 

leolost2605:
Preparation to use PanelClone for smoother MultitaskingView animations.

Regarding the memory leak: https://gitlab.gnome.org/GNOME/vala/-/issues/1548


src/ShellClients/HideTracker.vala
```diff
@@ -21,14 +21,17 @@ public class Gala.HideTracker : Object {
...
+
     private uint update_timeout_id = 0;
 
     public HideTracker (Meta.Display display, PanelWindow panel) {
         Object (display: display, panel: panel);
     }
 
     construct {
-        var current_focus_window = display.focus_window;
+        // Can't be local otherwise we get a reference cycle :(
```
danirabbit, line 33: Can you add a link to the vala issue report here?

leolost2605: Done :)

danirabbit (approved):
Great job cleaning up here. Seems like it makes doing the right thing a lot easier!

PR #1917, opened 2024-05-27
wayland: Don't hide under systemd

leolost2605:
Works around the broken session start for now. The proper fix would probably be to get the systemd units working (for some reason the X one works? or is gala launched by something else there?) but I don't know how :(


leolost2605:
See elementary/session-settings#75

danirabbit:
Can be closed now?

PR #1918, opened 2024-05-27
WindowSwitcher: draw our own borders

danirabbit:
We need to remove GTK style context for Wayland, but it's being used for both shadows and borders here. Draw our own border here so we can handle drawing our own shadow in a future branch without worrying about the border too

Screenshot from 2024-05-27 13 52 44

Remove some dead css while we're here

### DIFF
--- a/src/Widgets/WindowSwitcher/WindowSwitcher.vala
+++ b/src/Widgets/WindowSwitcher/WindowSwitcher.vala
@@ -142,10 +142,16 @@ public class Gala.WindowSwitcher : CanvasActor {
     }
 
     protected override void draw (Cairo.Context ctx, int width, int height) {
+        var background_color = Drawing.Color.LIGHT_BACKGROUND;
+        var border_color = Drawing.Color.LIGHT_BORDER;
         var caption_color = "#2e2e31";
+        var highlight_color = Drawing.Color.LIGHT_HIGHLIGHT;
 
         if (style_manager.prefers_color_scheme == Drawing.StyleManager.ColorScheme.DARK) {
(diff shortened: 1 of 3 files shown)


src/Widgets/WindowSwitcher/WindowSwitcher.vala
```diff
@@ -156,20 +162,38 @@ public class Gala.WindowSwitcher : CanvasActor {
...
         ctx.reset_clip ();
 
-        var background_color = Drawing.Color.LIGHT_BACKGROUND;
-        if (style_manager.prefers_color_scheme == Drawing.StyleManager.ColorScheme.DARK) {
-            background_color = Drawing.Color.DARK_BACKGROUND;
-        }
-
         ctx.set_operator (Cairo.Operator.SOURCE);
+
+        Drawing.Utilities.cairo_rounded_rectangle (ctx, 0.5, 0.5, width - 1, height - 1, InternalUtils.scale_to_int (9, scaling_factor));
```
leolost2605, line 167: I'm not sure whether the 0.5, - 1 here or -3 further below should be scaled as well but I guess with numbers this small it's not so important anyways?

leolost2605 (approved):
Looks good, thank you!

src/Widgets/WindowSwitcher/WindowSwitcher.vala
danirabbit: Ah yeah you’re right this should probably be scaled so we don’t get tiny borders on hidpi. Will fix!

PR #1919, opened 2024-05-27
Avoid ending our own process

leolost2605:
If a window created by a WaylandClient becomes unresponsive mutter will spawn a close dialog and if the user decides to force quit the window mutter will kill itself since windows created by waylandclients have mutters pid (see https://gitlab.gnome.org/GNOME/mutter/-/issues/3506). Instead safeguard closedialogs with windows with our own pid and instead handle is alive for waylandclients manually. Also with panels don't give the user the option of waiting.


src/Dialogs.vala
```diff
@@ -172,6 +172,18 @@ namespace Gala {
                 return;
             }
 
+            try {
+                var our_pid = new Credentials ().get_unix_pid ();
+                if (our_pid == window.get_pid ()) {
+                    critical ("We have an unresponsive window somewhere. Mutter wants to commit suicide. Don't let it.");
```
danirabbit, line 178: Might be kind of triggering language here :sweat_smile: 

```suggestion
                    critical ("We have an unresponsive window somewhere. Mutter wants to end its own process. Don't let it.");
```

src/ShellClients/ManagedClient.vala
```diff
@@ -29,6 +31,15 @@ public class Gala.ManagedClient : Object {
             display.window_created.connect ((window) => {
                 if (wayland_client != null && wayland_client.owns_window (window)) {
                     window_created (window);
+
+                    // We have to manage is alive manually since windows created by WaylandClients have our pid
+                    // and we don't want to commit suicide
```
danirabbit, line 36:
```suggestion
                    // and we don't want to end our own process
```

danirabbit (changes requested):
This makes sense, I think probably we should avoid using the word "suicide" here though

leolost2605:
Ah yeah that's probably not too good I was just really annoyed because it took me far too long to realize that mutter ends itself sorry

danirabbit (approved):

PR #1927, opened 2024-05-28
PanelClone: Don't animate while in fullscreen

leolost2605:
While in fullscreen docks are hidden so if we animate we just get the clone flying in and immediately disappearing

### DIFF
--- a/src/ShellClients/PanelClone.vala
+++ b/src/ShellClients/PanelClone.vala
@@ -89,6 +89,12 @@ public class Gala.PanelClone : Object {
         }
     }
 
+    private int get_animation_duration () {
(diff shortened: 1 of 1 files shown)


danirabbit:
Should this instead be any time the dock is hidden we don't animate? I've seen similar glitches with maximized apps for example

leolost2605:
Hmm yeah I know what you mean but I think it's a different problem. Because if windows are maximized and we hit the barrier we actually want the dock to show up but for some reason it won't causing the clone to just hide again. Whereas in fullscreen it's wanted behavior that the dock doesn't show so we just don't animate.

danirabbit (approved):

PR #1930, opened 2024-05-28
PanelClone: Animate in

leolost2605:
On start animate panels in.


danirabbit:
Should this be paired with a branch that removes the revealer stuff from the panel in a wayland session? https://github.com/elementary/wingpanel/blob/c1edcf44c41f08378d160bc246c61e04a9c0141f/src/PanelWindow.vala#L111

leolost2605:
That's probably a good idea, see elementary/wingpanel#538

danirabbit (approved):

PR #1935, opened 2024-05-29
PanelWindow: Make struts more reliable

leolost2605:
- Update struts on workspace change
- Update struts also on position change not only size

Fixes an issue that would cause the strut to be on the wrong display in some multi monitor setups on first start


danirabbit (approved):
This is working well for me, I haven't been able to break it :)

PR #1936, opened 2024-05-29
Fix icons

### LINKED ISSUE #1933: No app icons in Multitasking View, Window Switcher under Wayland
### What Happened?

There are no app icons in the alt+tab window switcher or the multitasking view when running under wayland

### Steps to Reproduce

Screenshot from 2024-05-29 08 21 39
Screenshot from 2024-05-29 08 21 54

### Expected Behavior

App icons :)

### OS Version

8.x (Early Access)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

leolost2605:
We don't have any connection to a screen so we can't get the default icon theme (which uses the default screen internally). Instead we create a new one which seems to use the right theme? And performance wise I don't notice anything strange either? AFAICT we just have to make sure that if we use it somewhere else we use the same one.

Fixes #1933

### DIFF
--- a/lib/Utils.vala
+++ b/lib/Utils.vala
@@ -31,9 +31,11 @@ namespace Gala {
 
         private static AppCache app_cache;
 
-        private static Gtk.IconTheme icon_theme = new Gtk.IconTheme ();
+        private static Gtk.IconTheme icon_theme;
 
         static construct {
+            icon_theme = new Gtk.IconTheme ();
+            icon_theme.set_custom_theme ("elementary");
             icon_cache = new Gee.HashMultiMap<DesktopAppInfo, CachedIcon?> ();
             window_to_desktop_cache = new Gee.HashMap<Meta.Window, DesktopAppInfo> ();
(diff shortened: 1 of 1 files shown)


lib/Utils.vala
```diff
@@ -31,6 +31,8 @@ namespace Gala {
 
         private static AppCache app_cache;
 
+        private static Gtk.IconTheme icon_theme = new Gtk.IconTheme ();
+
         static construct {
```
danirabbit, line 36: This works for apps that ship their own icon, but doesn't work for things that use system icons like Files. We can hardcode though to fix it:
```suggestion
        private static Gtk.IconTheme icon_theme;

        static construct {
            icon_theme = new Gtk.IconTheme ();
            icon_theme.set_custom_theme ("elementary");
```

leolost2605: Thank you :)

danirabbit (approved):

PR #1938, opened 2024-05-29
WindowManager: Add keybindings to launch dock items

leolost2605:
1 through 9 set by default to Super + number

Goes with elementary/dock#245


lenemter:
I think you should use `org.gnome.Shell.GrabAccelerators` and move these shortcuts to dock itself. You can look at https://github.com/elementary/wingpanel-indicator-sound/pull/270 as an example

leolost2605:
Oh nice I wasn't aware that that existed thank you :)

PR #1939, opened 2024-05-29
DesktopIntegration: Switch to correct workspace when focusing window

### LINKED ISSUE #1937: Launching from the dock doesn't switch workspaces to focus
### What Happened?

When clicking a launcher for an app that is already opened, nothing appears to happen

### Steps to Reproduce

in a wayland session (haven't verified if it happens in X)

1. Open an app
2. Switch to another workspace
3. Click that app's launcher in the dock
4. See nothing happen

### Expected Behavior

Clicking an app icon should bring me to the desktop where that app is instead of doing nothing

### OS Version

8.x (Early Access)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

leolost2605:
Fixes #1937


danirabbit (approved):
Love a one-liner

PR #1941, opened 2024-05-30
Dialogs: Fix Mutter bindings around dialog creation

tintou:
The dialog creation methods have to return a full reference.

Also we can assume that the dialog de-duplication work will already be done.


tintou:
(Upstream MR sent https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/3790 )

leolost2605 (approved):

PR #1942, opened 2024-05-30
wayland: Implement make_center

### LINKED ISSUE #1928: Add CENTER anchor to shell API
### Problem

There are several windows we want to always be centered on the screen like:
* Onboarding
* Installer
* Initial Setup
* Session dialogs

### Proposal

Add a "Center" anchor option

### Prior Art (Optional)

There used to be https://valadoc.org/gtk+-3.0/Gtk.WindowPosition.CENTER_ALWAYS.html which was cool because you could resize something like installer and it would remain centered on the screen when you did it

leolost2605:
Add a make_center method. These surfaces are always centered and can be used for e.g. shortcut overlay, onboarding, initial setup, various dialogs (like polkit, etc.).

I plan to add a make_modal method here as well that acts like #1879 

~Regarding naming stuff we can of course change it to anything we want. I just used the widget interface because I thought it was meant for that before reading @tintou's comment about it. If somebody wants to call it something different and has a good name for it I'll be happy to change it :)~

Fixes #1928 (IMHO it makes more sense to do it like this because a lot of stuff from the panel interface doesn't apply on centered "panels" like hide modes, barriers, struts, etc.)


leolost2605:
E.g. used with polkit agent: elementary/pantheon-agent-polkit#81

tintou:
Yeah widget is meant for Desktop widgets (such as [Nimbus](https://appcenter.elementary.io/io.github.danirabbit.nimbus/))

leolost2605:
Any alternative naming suggestions? Dialog? Centered? Or should I use the extended behavior?

tintou:
I guess that it fits the extended behavior, maybe we should have one function for centering and another for "focusing" with the rest of the desktop dimmed ? Or should we have everything in one as "ShellModal" ?

tintou:
I believe that @danirabbit want to give some input on if we have different use cases here

leolost2605:
Ooops I saw these comments to late I renamed it to system dialogs meanwhile but let's wait for @danirabbit

danirabbit:
The use cases I can think of are:
* Polkit
* Onboarding
* Installer + Initial Setup
* session dialog

I think the only one of those where some kind of modal styling might be weird is Onboarding because it can launch system settings for expanded options for example

leolost2605:
Ok so I guess we should keep modal optional since it won't be too hard anyways. Would you prefer it being in the extended behavior though? IMHO having a separate interface for this kinda makes sense to keep the methods more closely related but idk

tintou:
The "extended behavior" interface is exactly made for such cases :wink:

leolost2605:
Ok I moved it to extended behavior and introduced a make_centered there. I would leave the modal part to another PR

danirabbit:
@leolost2605 whats the best way to test this? I tried installing this branch and the polkit agent branch but I didn't notice any difference 🤔

leolost2605:
@danirabbit since the modal behavior and dimming is left to another PR the most visible thing would probably be that it always stays in the middle and also above other windows. In theory it shouldn't be movable (e.g. via shortcuts or super + drag) but I think it has big windowhandle on it so it manually requests moves which are still honored.

danirabbit:
@leolost2605 okay so I installed this branch and also https://github.com/elementary/pantheon-agent-polkit/pull/81 and logged out and back in and I can't confirm that polkit dialogs are appearing in the center of the display, and they aren't staying always on top either, but they do seem to appear on every workspace

leolost2605:
Then there's definitely something wrong here, I'll take another look at it :)

leolost2605:
Ok so turns out that 1. I forgot to set a quark and 2. I completely forgot you have to add the polkit agent to be launched as a trusted client via the dconf key `/io/elementary/wm/desktop/trusted-clients` as `/usr/libexec/policykit-1-pantheon/io.elementary.desktop.agent-polkit`

danirabbit:
Ahhh, maybe instead of (or in addition to) a dconf key we should check if the id is prefixed with `io.elementary`? Since a dconf key can't be updated so there would be no way to add new trusted clients to an existing install

leolost2605:
Maybe the name trusted client is a bit misleading but this actually doesn't only check whether it's allowed but actually launches it so some explicit setting is pretty much a must (we don't want to launch all elementary apps at galas start). I didn't actually consider that we can't update dconf keys (oops) but maybe we can do postinstalls? Or alternatively don't use dconf but hardcode (and then allow adding customs ones/disabling hardcoded ones)?

danirabbit:
@leolost2605 yeah dconf is considered user configuration so if we want like admin/maintainer configuration probably you want a keyfile in /etc

leolost2605:
@danirabbit would something like #1956 work?

danirabbit:
@tintou does this look okay to you? :)

tintou (approved):
looks good

PR #1948, opened 2024-06-16
Dialogs: Implement InhibitShortcutsDialog

leolost2605:
Fixes elementary/switchboard-plug-keyboard#512


danirabbit (approved):
No crash here! Looks good :)

PR #1950, opened 2024-06-22
DragDropAction: Fix memory leak

leolost2605:
We've had a ref cycle here leading to pretty big memory consumption if many workspaces were opened and closed again, since all the backgrounds were kept around.

### DIFF
--- a/lib/DragDropAction.vala
+++ b/lib/DragDropAction.vala
@@ -26,7 +26,7 @@ namespace Gala {
 
     public class DragDropAction : Clutter.Action {
         // DO NOT keep a reference otherwise we get a ref cycle
-        private static Gee.HashMap<string, Gee.LinkedList<unowned Actor>>? sources = null;
+        private static Gee.HashMap<string,Gee.LinkedList<unowned Actor>>? sources = null;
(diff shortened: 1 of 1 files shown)


lib/DragDropAction.vala
```diff
@@ -25,8 +25,9 @@ namespace Gala {
     }
 
     public class DragDropAction : Clutter.Action {
-        private static Gee.HashMap<string,Gee.LinkedList<Actor>>? sources = null;
-        private static Gee.HashMap<string,Gee.LinkedList<Actor>>? destinations = null;
+        // DO NOT keep a reference otherwise we get a ref cycle
+        private static Gee.HashMap<string, Gee.LinkedList<unowned Actor>>? sources = null;
+        private static Gee.HashMap<string,Gee.LinkedList<unowned Actor>>? destinations = null;
```
lenemter, line 30: A nitpick, but can you remove the space after the comma here please, just to be consistent :)

leolost2605: Done :)

lenemter (approved):
:+1:

Issue #1945, opened 2024-06-03
Zoom turns screen grey on Wayland


danirabbit:
### What Happened?

In the Wayland session zooming with Super + plus/minus turns the screen grey

### Steps to Reproduce

zoom with Super + plus/minus

### Expected Behavior

zoomed screen :)

### OS Version

8.x (Early Access)

### Software Version

Compiled from git

leolost2605:
Some additional info: this is only happening if the dock is visible for some reason

leolost2605:
Ok the reason for this is that we sometimes get some wild results when querying the ui_group for width and height (ranging from 0 to some random number) in `compute_new_pivot_point` in the zoom class. This leads to pivot points outside the 0 to 1 range causing the grey screen.

PR #1953, opened 2024-06-23
Manually set ui_group size

lenemter:
Fixes https://github.com/elementary/gala/issues/1945


leolost2605 (approved):
Can confirm it works (also with hotplugging multiple monitors)! Very nice :fire:

PR #1956, opened 2024-06-25
Shell: Move trusted clients config to own config file

### LINKED ISSUE #1946: Dock appears in the window switcher and multitasking view on X session
### What Happened?

Dock appears in the Alt + Tab window switcher and multitasking view if X session. This does not occur on Wayland session.

Screenshot from 2024-06-05 20-27-11

screen

### Steps to Reproduce

1. Log in to X session
2. Press Alt + Tab. Alternatively, open multitasking view

### Expected Behavior

Dock should not appear in the window switcher and multitasking view.

### OS Version

8.x (Early Access)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

leolost2605:
While here fix #1946 

See https://github.com/elementary/gala/pull/1942#issuecomment-2186839264

I'm not sure how apt handles files in /etc but this should be overwritten by updates no?

### DIFF
--- a/src/ShellClients/ShellClientsManager.vala
+++ b/src/ShellClients/ShellClientsManager.vala
@@ -38,7 +38,28 @@ public class Gala.ShellClientsManager : Object {
     }
 
     private async void start_clients () {
-        var file = File.new_build_filename (Config.SYSCONFDIR, "io.elementary.desktop.wm.shell");
+        // Prioritize user config over system
+        (unowned string)[] config_dirs = { Environment.get_user_config_dir () };
+        foreach (unowned var dir in Environment.get_system_config_dirs ()) {
+            config_dirs += dir;
+        }
+
+        string? path = null;
+        foreach (unowned var dir in config_dirs) {
+            var file_path = Path.build_filename (dir, "io.elementary.desktop.wm.shell");
+            warning (file_path);
+            if (FileUtils.test (file_path, EXISTS)) {
+                path = file_path;
+                break;
+            }
+        }
+
+        if (path == null) {
+            warning ("No shell config file found.");
+            return;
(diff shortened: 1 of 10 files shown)


src/ShellClients/ShellClientsManager.vala
```diff
@@ -34,10 +34,50 @@ public class Gala.ShellClientsManager : Object {
...
 
-        var shell_settings = new Settings ("io.elementary.desktop.wm.shell");
-        var clients = shell_settings.get_value ("trusted-clients");
-        foreach (var client in clients) {
-            protocol_clients += new ManagedClient (wm.get_display (), client.get_strv ());
+        start_clients.begin ();
+    }
+
+    private async void start_clients () {
+        var file = File.new_build_filename (Config.SYSCONFDIR, "io.elementary.desktop.wm.shell");
```
danirabbit, line 41: Here I think you want to get system config dirs from glib: https://valadoc.org/glib-2.0/GLib.Environment.get_system_config_dirs.html

here’s an example of looping over the list so we prioritize user overrides https://github.com/elementary/switchboard-plug-parental-controls/blob/49100787dcfd354f6a249c9f8307182d3e659226/src/daemon/Server.vala#L297

danirabbit:
I think this is right but let’s double check with @tintou :)

src/ShellClients/ShellClientsManager.vala
leolost2605: Done :)

danirabbit (approved):
I'd like to get this in for now since at least it gets us a step towards making sure both the Wayland and X11 sessions are usable and I think it's better than dconf at least

Issue #1960, opened 2024-06-29
Multitasking view doesn't respond to touchscreen events


danirabbit:
### What Happened?

You can use gestures to enter and exit the multitasking view, but elements inside the view like window clones, the workspace switcher, etc don't respond to touch events

### Steps to Reproduce

1. Using a touch screen device
2. Open the multitasking view
3. Attempt to select a window or a workspace via the touch screen
4. See that nothing responds to input

### Expected Behavior

Elements in the multitasking view should be tap-able so that we can multitask completely via touch in tablet formats

### OS Version

8.x (Early Access)

### Software Version

Latest release (I have run all updates)

### Hardware Info

Starlite Mk V

PR #1962, opened 2024-06-30
DragDropAction: support touch

lenemter:
I can't test these changes but they should fix https://github.com/elementary/gala/issues/1960


danirabbit:
This makes Drag and Drop operations work which is great! but it doesn't fix tap operations. Happy to merge this and address that in a separate branch if that makes more sense to you :)

danirabbit (approved):

leolost2605:
FWIW mutter will introduce a reworked gesture api (https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/2857) which should make things a bit easier

PR #1963, opened 2024-06-30
wayland/NotificationStack: Don't flicker at the beginning of the animation

leolost2605:
On wayland the notification shows completely for a frame before the animation starts. Prevent this by explicitly setting opacity and rotation to their initial values.


lenemter (approved):
Fixes the issue. Good job :+1:

PR #1964, opened 2024-07-02
Daemon: Port to GTK4

### LINKED ISSUE #1695: No context menu on wallpaper in Wayland

### LINKED ISSUE #1696: Window context menu doesn't automatically close on Wayland

leolost2605:
Fixes #1696
Fixes #1695


daemon/WindowMenu.vala
```diff
@@ -3,11 +3,13 @@
...
  */
 
-public class Gala.Daemon.WindowMenu : Gtk.Menu {
+public class Gala.Daemon.WindowMenu : Gtk.Popover {
     private static GLib.Settings gala_keybind_settings = new GLib.Settings ("org.pantheon.desktop.gala.keybindings");
     private static GLib.Settings keybind_settings = new GLib.Settings ("org.gnome.desktop.wm.keybindings");
 
-    public signal void perform_action (Gala.ActionType type);
+    public signal void perform_action (Gala.ActionType type) {
+        popdown ();
```
danirabbit, line 11: I didn't even know you could have a function attached to a signal. That's dope

meson.build
```diff
@@ -89,10 +87,11 @@ gio_dep = dependency('gio-2.0', version: '>= @0@'.format(glib_version_required))
 gio_unix_dep = dependency('gio-unix-2.0', version: '>= @0@'.format(glib_version_required))
 gmodule_dep = dependency('gmodule-2.0')
 gtk_dep = [dependency('gtk+-3.0', version: '>= @0@'.format(gtk_version_required)), dependency('gdk-x11-3.0')]
+gtk4_dep = dependency('gtk4')
 gee_dep = dependency('gee-0.8')
-granite_dep = dependency('granite', version: '>= 5.4.0')
+granite_dep = dependency('granite-7')
 gnome_desktop_dep = dependency('gnome-desktop-3.0')
-hdy_dep = dependency('libhandy-1')
+adw_dep = dependency('libadwaita-1')
```
danirabbit, line 94: Are you actually using Adw here? I think we were only using Hdy for the window

leolost2605: Ah right :)

leolost2605:
Oh right X is a thing oops :sweat_smile: 
I think monitor labels won't work at all on X :(
Maybe we should keep the gtk3 version as legacy and depending on wayland or x launch the right one? Shouldn't be a maintenance burden since we probably don't have to fix anything there? And we always can just drop it :shrug:

danirabbit:
@leolost2605 that makes sense to me! If that's not too much trouble for you

leolost2605:
@danirabbit done!

Issue #613, opened 2019-09-21
Slow-loading wallpapers set incorrect panel color on login


cassidyjames:
It looks like a slow-loading wallpaper (i.e. a very large resolution on a not super powerful system) causes Gala to think the wallpaper is whatever the color is before it's loaded (i.e. if there's a color set, it uses that instead—black by default). For light wallpapers, this means the background ends up being black for a second, which then renders white text on the light wallpaper once it loads in.

cassidyjames:
As a note, it only seems to do this on login; when changing the wallpaper in-session, the correct panel color is set.

I've also confirmed it can be worked around by setting the wallpaper color to white, _then_ choosing the light wallpaper. Then during that moment on login when the wallpaper is loading, the background is white and the correct panel color is set.

cassidyjames:
I'm able to reproduce this on two machines with the new `Canazei Granite Ridges.jpg` wallpaper.

cassidyjames:
To reproduce:

1. In _System Settings → Desktop_ ensure the wallpaper color is set to black
2. Set [Canazei Granite Ridges.jpg](https://github.com/elementary/wallpapers/blob/master/Canazei%20Granite%20Ridges.jpg) as your wallpaper
3. Log out and back in

If your machine is similar to mine, it'll take a moment to load the wallpaper when logging back in, and the panel color will be incorrect. If you repeat it but set the color to white in step one, the panel is set to the correct color.

cassidyjames:
@donadigo had mentioned being able to reproduce this, but I'm not sure if there's been any progress.

cassidyjames:
This is more easily reproducible for me on the Pinebook Pro with any of our light-topped wallpapers. @davidmhewitt had mentioned possibly taking a look at this? :crossed_fingers::sweat_smile:

PR #1968, opened 2024-07-07
BackgroundManager: Send changed signal early

lenemter:
Fixes https://github.com/elementary/gala/issues/613


danirabbit (approved):
Makes sense to me. Nice use of `requires`

PR #1972, opened 2024-07-07
WindowSwitcher: Fix memory leak

lenemter:
Avoid using local variable in signal handler.


leolost2605 (approved):
I can confirm this fixes WindowSwitcherIcons not being freed. Thank you!

Issue #514, opened 2019-04-04
Authentication window looses focus


marmeladapk:
When opening an app that requires root from application menu (checked with synaptic and grub-customizer) authentication dialog appears and looses focus after ~0.5s. Last window that has focus regains it. This can result in typing and sending password in undesired places.

I have two 4k screens with doubled pixels (I read in greeter issues that focus issues may be related to hi-dpi). I have latest version from repository `0.3.2~r690+pkg45~ubuntu5.0.1 amd64`

Is there any additional info I can provide?

ed: On 1080p and 4k, both with doubling pixels and with lodpi option this issue persists. Removing one monitor also didn't fix it.

lenemter:
Closing as stale. If you still encounter this issue, please reply here and I'll reopen it

PR #1974, opened 2024-07-07
wayland: Add make_modal protocol method

leolost2605:
Takes a bunch of stuff from #1879 and makes it a wayland protocol method. 

This will open up a bunch of useful possibilities like making end session, polkit, etc. dialogs modal with a dimmed background or like using GTK for the window switcher in an external process (that's also why dimming is optional).

### DIFF
--- a/protocol/pantheon-desktop-shell-v1.xml
+++ b/protocol/pantheon-desktop-shell-v1.xml
@@ -115,6 +115,8 @@
       <description summary="requests to make a surface system modal">
         This will block all user input outside the surface and most system shortcuts.
       </description>
+
+      <arg name="dim" type="uint" summary="1 to dim, 0 to not dim"/>
     </request>
   </interface>
 </protocol>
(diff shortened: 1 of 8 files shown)


zeebok (dismissed):
There are a bunch of git artifacts that got in and need to be cleaned up

danirabbit:
Maybe fixes #514?

danirabbit:
@leolost2605 can you resolve conflicts here? Would love to get this feature in :)

leolost2605:
@danirabbit do we want to dim everything or leave shell elements like wingpanel and dock undimmed?

danirabbit:
@leolost2605 probably everything

leolost2605:
This will keep the same behavior as main on older mutter which currently is only opensuse leap 15 afaik. Leap 16 has been released so this should be fine IMO.

danirabbit:
@lenemter can we have a re-review here? I'd love to get this in for OS 8.1 :)

danirabbit:
@leolost2605 can you address this final review comment? :)

PR #1975, opened 2024-07-08
WindowCloneContainer: Don't hold an extra reference to WindowClone

lenemter:
Fixes the issue when if you close window in the multitasking view (using the close button and without dragging it first) WindowClone will be freed only after reopening multitasking view. It's not a memory leak but still would be better to free objects early.


leolost2605:
Idk but I usually prefer setting things to `null` instead of making them `unowned`. Because `unowned` always has a kind of unsafe feeling for me...

PR #1976, opened 2024-07-08
DragDropAction: Don't hold any references to the actors

lenemter:
Fixes an issue with WindowClone where it can not be freed after dragging.


tintou:
I've looked at this MR, wouldn't https://github.com/elementary/gala/pull/1979 fix the issue you were having?

PR #1979, opened 2024-07-10
DragDropAction: Release hovered reference when finishing the action

tintou:
Make sure that we are not holding a reference to the object after handling the dnd action.


lenemter (changes requested):
I think we should release `handle` reference too, but we should do this in `cleanup` since the drag may not succeed

tintou:
@lenemter indeed, applied your suggestion!

lenemter (approved):
Fixes the memory leak in multitasking view. Good job!

PR #1983, opened 2024-07-12
Shell Config: add polkit agent

danirabbit:
Goes with https://github.com/elementary/pantheon-agent-polkit/pull/81/


leolost2605 (approved):

tintou:
I don't think that this is required at all

PR #1995, opened 2024-07-20
ShellClients: Don't make centered clients docks but allow their shortcut inhibition without dialog

### LINKED ISSUE #1994: Redesign center windows
For the centered windows, we don't want them to be docks, just centered, we also want to use the [zwp_keyboard_shortcuts_inhibit_manager_v1](https://wayland.app/protocols/keyboard-shortcuts-inhibit-unstable-v1#zwp_keyboard_shortcuts_inhibit_manager_v1) protocol (with the proper filtering in create_inhibit_shortcuts_dialog to allow it directly)

leolost2605:
Fixes #1994 

We don't really need the focus method now anymore, should we remove it too? Or might it come in handy at some other point in the future?


src/Dialogs.vala
```diff
@@ -229,8 +229,9 @@ namespace Gala {
                 return;
             }
 
-            // Naive check to always allow inhibiting by our settings app. This is needed for setting custom shortcuts
-            if (app.id == "io.elementary.settings.desktop") {
+            if (app.id == "io.elementary.settings.desktop" || // Naive check to always allow inhibiting by our settings app. This is needed for setting custom shortcuts
```
danirabbit, line 232: While we're here should we just do prefix of `io.elementary`?

tintou, line 232: I wouldn't put the check here but in the center window settings, only allow centered windows for the desktop utility, deny it for the rest, then it will always fail for the `is_positioned_window` check

src/ShellClients/ShellClientsManager.vala
```diff
@@ -152,8 +152,10 @@ public class Gala.ShellClientsManager : Object {
             return;
         }
 
-        make_dock (window);
-
         centered_windows[window] = new CenteredWindow (wm, window);
     }
+
+    public bool is_positioned_window (Meta.Window window) {
+        return (window in centered_windows) || (window in windows);
```
tintou, line 159: if this is a window of a positioned windows, it probably counts as a positioned window too?

src/Dialogs.vala
leolost2605: Uhmm I'm not quite sure I understand what you mean 😬

src/ShellClients/ShellClientsManager.vala
leolost2605: Added that :)

tintou (approved):
Great!

src/Dialogs.vala
tintou: you did it ;)

PR #1997, opened 2024-07-25
Implement PantheonShellX11 protocol

### LINKED ISSUE #1965: Launch Dock on X11
### What Happened?

We need a dock for the X11 session since the new dock only works with Wayland

### Steps to Reproduce

Start X11 session and see there's no dock since 956125b488ba3da2ea1b1d93ffa8c0521e14a3ff

### Expected Behavior

Make sure we launch Plank under X11

### OS Version

8.x (Early Access)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

lenemter:
Closes #1965 

Makes Gtk4 Dock (and other Gtk4 ports in the future) usable in X11 session.

This is achieved using 3 "hacks":

1. Adds `supports-id` as an attribute in shell list. If it's true, then window is launched with a unique id by passing it as `--id <id here>`. Then Gala can identify new windows by checking the window titles.
2. Implements `PantheonShellX11` protocol for positioning windows on X11. It requires an id from the first step. Other than that it's a simplified copy of our Wayland protocol.
3. Since mutter can't change window_type on X11, Gala does it manually via `X.Display`.

You can use https://github.com/elementary/dock/pull/260 to test it with the dock.

Unblocks https://github.com/elementary/notifications/pull/157

### DIFF
--- a/data/io.elementary.desktop.wm.shell
+++ b/data/io.elementary.desktop.wm.shell
@@ -1,11 +1,14 @@
 [io.elementary.wingpanel]
 launch-on-x=true
 args=io.elementary.wingpanel
+supports-id=false
 
 [io.elementary.desktop.agent-polkit]
 launch-on-x=true
 args=/usr/libexec/policykit-1-pantheon/io.elementary.desktop.agent-polkit
+supports-id=false
 
 [io.elementary.dock]
-launch-on-x=false
+launch-on-x=true
 args=io.elementary.dock
+supports-id=true
--- a/po/uk.po
+++ b/po/uk.po
@@ -8,7 +8,7 @@ msgstr ""
 "Project-Id-Version: beat-box\n"
 "Report-Msgid-Bugs-To: https://github.com/elementary/gala/issues\n"
 "POT-Creation-Date: 2024-07-18 17:28+0000\n"
-"PO-Revision-Date: 2024-07-10 16:16+0000\n"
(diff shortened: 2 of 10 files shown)


danirabbit:
I think this is a bit beyond my technical knowledge, but afaict everything works as expected and I'd love to not have to ship Plank for X11 :heart:

leolost2605:
This would be really cool :)

Since X11 has pretty much no security anyways what if we just allow windows to set the x property _MUTTER_HINTS (or one of our own if preferred) to something like `anchor=0:hide_mode=never` or `centered=1` and then act accordingly e.g. in map.
This would save us the id and dbus stuff while also allowing the same process to have multiple windows that act differently. 
Or maybe I've missed something this method wouldn't allow for?

lenemter:
@leolost2605 I think using x properties is double-edged sword. While, yes, it would save us ids and dbus stuff, but we would have to deal with x properties (which I assume is a bit more complicated than simple dbus interface) and also listen to properties changes in Gala, since user can change dock hide mode.

leolost2605:
IMHO it's not that bad at least when using the _MUTTER_HINTS property because there mutter already handles everything for us. I did a quick implementation at #2002 which works with elementary/dock#263. What would you say about that? Or if you really prefer we could also do it this way, although we kinda want to move away from launching our clients in general see https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/3844)

PR #2001, opened 2024-07-27
DBus: cleanup

lenemter:
* Inline namespace
* Remove methods that were moved into the wingpanel plugin 9 years ago


leolost2605 (approved):
LGTM

PR #2002, opened 2024-07-27
ShellClients: Allow X11 clients to use pantheon protocol via mutter hints

### LINKED ISSUE #1965: Launch Dock on X11

leolost2605:
Fixes #1965 
Closes #1997 

Has the same functionality as #1997 but uses the x property _MUTTER_HINTS


src/ShellClients/ShellClientsManager.vala
```diff
@@ -36,6 +36,12 @@ public class Gala.ShellClientsManager : Object {
         notifications_client = new NotificationsClient (wm.get_display ());
 
         start_clients.begin ();
+
+        //X11 only
+        wm.get_display ().window_created.connect ((window) => {
+            window.notify["mutter-hints"].connect ((obj, pspec) => parse_mutter_hints ((Meta.Window) obj));
+            parse_mutter_hints (window);
+        });
```
lenemter, line 44:
```suggestion
        if (!Meta.Util.is_wayland_compositor ()) {
            wm.get_display ().window_created.connect ((window) => {
                window.notify["mutter-hints"].connect ((obj, pspec) => parse_mutter_hints ((Meta.Window) obj));
                parse_mutter_hints (window);
            });
        }
```

src/ShellClients/ShellClientsManager.vala
```diff
@@ -98,7 +104,15 @@ public class Gala.ShellClientsManager : Object {
...
-    private void make_dock (Meta.Window window) {
+    public void make_dock (Meta.Window window) {
+        if (Meta.Util.is_wayland_compositor ()) {
+            make_dock_wayland (window);
+        } else {
+            make_dock_x11 (window);
+        }
+    }
+
+    private void make_dock_wayland (Meta.Window window) {
```
lenemter, line 115:
```suggestion
    private void make_dock_wayland (Meta.Window window) requires (Meta.Util.is_wayland_compositor ()) {
```

src/ShellClients/ShellClientsManager.vala
```diff
@@ -156,4 +189,50 @@ public class Gala.ShellClientsManager : Object {
 
         centered_windows[window] = new CenteredWindow (wm, window);
     }
+
+    //X11 only
+    private void parse_mutter_hints (Meta.Window window) {
```
lenemter, line 194:
```suggestion
    private void parse_mutter_hints (Meta.Window window) requires  (!Meta.Util.is_wayland_compositor ()) {
```

lenemter (commented):
That's way cleaner then using dbus. Thank you. Still waiting for @tintou review.

src/ShellClients/ShellClientsManager.vala
```diff
@@ -156,4 +190,50 @@ public class Gala.ShellClientsManager : Object {
...
+            if (split.length != 2) {
+                continue;
+            }
+
+            var key = split[0];
+            var val = split[1];
+
+            switch (key) {
+                case "anchor":
+                    set_anchor (window, int.parse (val));
```
tintou, line 213: I would use `try_parse` just to be on the safe side

leolost2605: Done :)

lenemter (changes requested):

lenemter:
Also, I noticed that this PR doesn't allow using Gtk4 notifications server in X11. If we will use this approach with mutter hints, we will have to add a new parameter or something like that to identify the windows. But we can deal with that later.

src/ShellClients/ShellClientsManager.vala
```diff
@@ -156,4 +190,63 @@ public class Gala.ShellClientsManager : Object {
...
+
+            switch (key) {
+                case "anchor":
+                    int parsed;
+                    if (int.try_parse (val, out parsed) && 0 <= parsed && parsed <= 15) {
+                        set_anchor (window, parsed);
+                    } else {
+                        warning ("Failed to parse %s as anchor", val);
+                    }
+                    break;
```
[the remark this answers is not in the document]
leolost2605: Added a comment :)

set_anchor uses Meta.Side which is a 4 value bitfield so it can go up to 15

leolost2605:
I think regarding the notifications server we can just use the x library to set our window type as notification just like the gtk3 version does

src/ShellClients/ShellClientsManager.vala
```diff
@@ -156,4 +190,63 @@ public class Gala.ShellClientsManager : Object {
...
+
+                case "hide-mode":
+                    int parsed;
+                    if (int.try_parse (val, out parsed) && 0 <= parsed && parsed <= 4) {
+                        set_hide_mode (window, parsed);
+                    } else {
+                        warning ("Failed to parse %s as hide mode", val);
+                    }
+                    break;
```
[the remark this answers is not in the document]
leolost2605: Done :)

lenemter (approved):
Code looks good :tada:

Issue #1617, opened 2023-04-04
Style PiP as a window


danirabbit:
### Problem

The current PiP has no window borders or shadow

![Screenshot from 2023-04-04 12 22 22](https://user-images.githubusercontent.com/7277719/229898458-0cb9350b-b96a-4f94-9946-4c8a5664d379.png)

### Proposal

We should use the border radius, box shadow, etc from the system style for a regular window so that it's styled

ToDo:
- [x] Add shadow
- [ ] More subtle shadow style
- [x] Rounded corners
- [ ] Border

PR #2005, opened 2024-08-03
PiP: Fix shadow

lenemter:
Fixes missing shadow from https://github.com/elementary/gala/issues/1617


danirabbit (approved):
Can confirm that indeed this fixes the shadow. Nice work

PR #2006, opened 2024-08-04
 DesktopIntegration/focus_window: Shake window and bell if already focused 

### LINKED ISSUE #1978: Bounce window when focusing an already focused window
If I click on an app on the dock that is already focused (it is sometimes on my other monitor and I don't remember it) it would be great to have a visual response to my action (like the window bouncing?)

leolost2605:
Sounds a bell and shakes the window if trying to focus an already focused window.

The shake is taken from the polkit dialog. However we could alternatively do something like a more relaxed bounce or similar.

Fixes #1978


danirabbit (approved):
Yo this is so good!

src/DesktopIntegration.vala
```diff
@@ -107,12 +107,55 @@ public class Gala.DesktopIntegration : GLib.Object {
...
 
+    private bool notifying = false;
+    private void notify_already_focused (Meta.Window window) {
+        if (notifying) {
+            return;
+        }
+
+        notifying = true;
+
+        wm.get_display ().get_sound_player ().play_from_theme ("bell", _("Window has already focus"), null);
```
danirabbit, line 128: Can we not use Gdk.Beep in Gala?

leolost2605: Unfortunately not, we get a glib critical along the lines of `assertion Gdk.Display != null failed` :(

tintou: yeah no that's better to use Mutter there

src/DesktopIntegration.vala
```diff
@@ -107,12 +107,55 @@ public class Gala.DesktopIntegration : GLib.Object {
...
+            var should_continue = n <= 10;
+
+            if (!should_continue) {
+                notifying = false;
+                window.move_frame (false, rect.x, rect.y);
+            }
+
+            return should_continue;
+        });
+    }
```
tintou, line 157: It feels to me that we should use Mutter integrated way to handle animations (which relies on frames instead of GLib time)

There is a similar utility in GNOME Shell for instance https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/js/misc/animationUtils.js#L80

tintou: FYI this is how the "ease" function is defined https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/js/ui/environment.js#L92

leolost2605: Hmm I get some weird flickering when doing this:

[Kooha-2024-08-07-15-39-04.webm](https://github.com/user-attachments/assets/12553b33-4512-4899-a096-b100ef1aeeb9)

Any idea why?

Code:
```vala
        var transition = new Clutter.KeyframeTransition ("translation-x") {
            repeat_count = 5,
            duration = 200,
            remove_on_complete = true
        };
        transition.set_from_value (0);
        transition.set_to_value (0);
        transition.set_values ( { -20, 20 });

        transition.stopped.connect (() => {
            notifying = false;
            wm.get_display ().enable_unredirect ();
        });

        wm.get_display ().disable_unredirect ();

        ((Meta.WindowActor) window.get_compositor_private ()).add_transition ("notify-already-focused", transition);
```

lenemter: You forgot to add key frames:

`transition.set_key_frames ( { 0.5, -0.5 } );`

leolost2605: @lenemter thank you that fixed it!

leolost2605:
I'm currently scaling the offset by which the window is shaken is that right?

And we can easily change duration, easing, etc. so if anybody has suggestions go ahead because I dont know what looks good :)

tintou:
Really glad you did it, it can be tweaked afterwards!

tintou (approved):

PR #2013, opened 2024-08-10
Sort windows by stacking order before workspace switch

lenemter:
Fixes an issue when maximized windows could've had an incorrect stacking order on workspace switch


leolost2605 (approved):
Tbh haven't noticed the issue yet but the PR LGTM :)

PR #2015, opened 2024-08-10
PanelClone: avoid using Timeout

lenemter:


leolost2605 (commented):
This breaks when disabling animation :(

leolost2605 (commented):
If you want we can do it like this but wouldn't it maybe then just be easier to stay with the timeout?

leolost2605 (approved):
This actually fixes a rare crash so let's got with it
(Can be removed with #2074 anyways)

Issue #2014, opened 2024-08-10
Taking screenshots is completely broken


lenemter:
### What Happened?

1. Taking fullscreen screenshots produces black screen
2. On wayland taking screenshots to a clipboard is broken

### Steps to Reproduce

???

### Expected Behavior

Working screenshot feature

### OS Version

8.x (Early Access)

### Software Version

Compiled from git

leolost2605:
> Taking fullscreen screenshots produces black screen

Hmm I can't reproduce this. Neither via shortcut, nor via the portal :shrug:

lenemter:
I can't reproduce it anymore :shrug:

PR #2017, opened 2024-08-12
ScreenshotManager: Fix saving to clipboard

lenemter:
Addresses https://github.com/elementary/gala/issues/2014

### DIFF
--- a/src/ScreenshotManager.vala
+++ b/src/ScreenshotManager.vala
@@ -300,7 +300,7 @@ namespace Gala {
             return Environment.get_home_dir ();
         }
 
-        private static async bool save_image (Cairo.ImageSurface image, string filename, float scale, out string used_filename) {
+        private async bool save_image (Cairo.ImageSurface image, string filename, float scale, out string used_filename) {
             return (filename != "")
                 ? yield save_image_to_file (image, filename, scale, out used_filename)
                 : save_image_to_clipboard (image, filename, out used_filename);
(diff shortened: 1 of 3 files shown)


vapi/libmutter.vapi
```diff
@@ -793,6 +793,7 @@ namespace Meta {
 	public sealed class SelectionSourceMemory : Meta.SelectionSource {
 		[CCode (has_construct_function = false, type = "MetaSelectionSource*")]
 		public SelectionSourceMemory (string mimetype, GLib.Bytes content);
```
leolost2605, line 795: AFAIK you can just do something like this:
```suggestion
		public SelectionSourceMemory (string mimetype, GLib.Bytes content) throws GLib.Error;
```
which avoids having to use the static function

src/ScreenshotManager.vala
```diff
@@ -342,19 +342,40 @@ namespace Gala {
...
-            unowned Gtk.Clipboard clipboard = Gtk.Clipboard.get_default (display);
-
             var screenshot = Gdk.pixbuf_get_from_surface (image, 0, 0, image.get_width (), image.get_height ());
             if (screenshot == null) {
-                warning ("could not save screenshot to clipboard: null pixbuf");
+                warning ("Could not save screenshot to clipboard: null pixbuf");
+                return false;
+            }
+
+            var stream = new GLib.MemoryOutputStream.resizable ();
```
leolost2605, line 354: I'm really not to knowledgeable in this area so what's the advantage of using a memory stream here? Wouldn't `save_to_buffer` together with `new Bytes.take ()` do the same? Or is saving to a memory stream here somehow faster?

leolost2605 (changes requested):

vapi/libmutter.vapi
lenemter: It didn't work for me when I tried it, but now for some reason it just works

src/ScreenshotManager.vala
lenemter: I used it because gnome-shell uses it, `GLib.Bytes.take` works just fine. Thank you.

leolost2605 (approved):
Very cool :)

PR #2022, opened 2024-08-13
Zoom: avoid using `Gdk.beep`

lenemter:


danirabbit (approved):
This fixes missing beep on Wayland for me. nice work!

PR #2028, opened 2024-08-16
ShellClients: Fix memory leak and crash

leolost2605:
If a shellclient window became unmanaged the corresponding panelwindow wasn't actually destroyed due to a ref cycle. This could cause a crash when a client was killed and then launched again and then something about the monitors (e.g. scale) changed. Also it was a memory leak.


danirabbit (approved):
Confirmed the crash and confirmed that this fixes it. Good catch! Kind of a niche crash

PR #1951, opened 2024-06-22
WindowSwitcher: set accessible properties

danirabbit:
Fixes #1301 

Not familiar enough with Clutter to know why this isn't working, but theoretically should add accessible labels, role, and state

PR #2030, opened 2024-08-19
WindowSwitcher: set accessible properties

lenemter:
Fixes https://github.com/elementary/gala/issues/1301
Based on https://github.com/elementary/gala/pull/1951


danirabbit (approved):
Can confirm this works as expected. Thanks for fixing it!

Issue #271, opened 2018-06-11
RDNN GSettings


codygarver:
No more `org.pantheon`, only `io.elementary`.

IMPORTANT: must also update gala path in Wingpanel!!! Or it crashes.

danrabbit:
The RDNN for gala should probably be `io.elementary.desktop.wm` to be consistent with the precedent set by `org.gnome.desktop.wm`

lenemter:
Fixed in #2034

PR #2032, opened 2024-08-19
RDNN keybindings gsettings schema

lenemter:
I think we should RDNN gsettings before OS 8 release because the next opportunity will be OS 9, and it's like a long time from now.

Addresses https://github.com/elementary/gala/issues/271


danirabbit (approved):

danirabbit:
Yeah let's do it. Just causes a couple re-releases but that's okay :)

Issue #1052, opened 2021-02-15
Workspace doesn't switch automatically when invoke app from different workspace


shantanu-csedu:
**How to regenerate:**
- open file browser in workspace 1
- open a web browser (chrome or others) in workspace 2
- Go to the downloads page in web browser and click show in folder

**Current behavior:**
File browser will open with the selected file but doesn't switch to the workspace where the file browser is. No way users know something happens.

**Expected behavior:**
Switch to file browser's workspace with file browser window focused

**Platform information:**
Elementary OS 5.1

PR #2041, opened 2024-08-21
Focus window when it demands attention

lenemter:
Fixes https://github.com/elementary/gala/issues/1052


leolost2605 (approved):
LGTM, I guess we could do these few lines of code also in WindowManager although I get that you want to avoid making it bigger than it already is so u decide :)

PR #2046, opened 2024-08-24
RDNN dbus

lenemter:


danirabbit:
I think we need to hold on this because we just release 8.0.0 and downstreams are already packaging etc and this would be a breaking change

PR #2048, opened 2024-09-01
MultiTaskingView: Fix notifications sliding in and out

leolost2605:
Currently if you open the multitaskingview while a notification is open they are being animated in and out like a normal dock although they should just stay where they normally are.


danirabbit (approved):
Nice simple fix!

PR #2050, opened 2024-09-13
Unify notifications check

lenemter:
Unifies a notification check to avoid mistakes and confusion in the future.


danirabbit (approved):
Makes sense to me :+1:

PR #2056, opened 2024-09-17
Move notifications out of the way for overlapping dock windows

leolost2605:
Same objective as #1594 but does it in a more general way and only requires changes in gala.

### DIFF
--- a/src/NotificationStack.vala
+++ b/src/NotificationStack.vala
@@ -28,8 +28,9 @@ public class Gala.NotificationStack : Object {
 
     private const int WIDTH = 300;
 
-    private int stack_y;
-    private int stack_width;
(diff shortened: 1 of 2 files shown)


leolost2605:
Actually I just realized this might require the GTK 4 port of wingpanel since popovers aren't handled as separate windows in GTK 3 at all it seems (or at least if it can be avoided idk)

teamcons:
> Actually I just realized this might require the GTK 4 port of wingpanel since popovers aren't handled as separate windows in GTK 3 at all it seems (or at least if it can be avoided idk)

Aaaah too bad :( this is a PR im hyped with
any idea how long thatd take ?

PR #2061, opened 2024-09-19
ShadowEffect: Delay cache drop for switching shadow sizes performance improvements

### LINKED ISSUE #2060: Gala.ShadowEffect has performance issues
### What Happened?

Using Gala ShadowEffect causes noticeable performance degradation. Especially noticeable during a workspace switch while not in multitasking view which stutters currently but is super smooth if you remove the shadow effects added to the wallpapers.

### Steps to Reproduce

Workspace switch stutters (at least on less performant laptops)

### Expected Behavior

Don't stutter

### OS Version

8.x (Early Access)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

leolost2605:
Idk if this is a good and/or common way to do this but it should work :shrug:  

The reasoning here is that during the workspace switch animation the actor width and/or height often changes by one pixel back and forth. This would cause the shadow to be dropped and created again multiple times during a switch. By keeping them in the cache for 5 seconds before dropping them we can avoid this. While a second or so would be enough for a single switch the whole feel is a lot smoother when trying to switch over multiple workspaces in one go.


lib/ShadowEffect.vala
```diff
@@ -184,4 +175,39 @@ public class Gala.ShadowEffect : Clutter.Effect {
...
+            return;
+        }
+
+        if (--shadow.users == 0) {
+            queue_shadow_drop (key);
+        }
+    }
+
+    private static void queue_shadow_drop (string key) {
+        shadows_marked_for_dropping[key] = Timeout.add_seconds (5, () => {
```
lenemter, line 207:
```suggestion
        shadows_marked_for_dropping[key] = Timeout.add_seconds_once (5, () => {
```

leolost2605: The *_once functions unfortunately do not ref the data and since vala unrefs it and it's also not used anywhere else, it gets immediately freed. This causes a crash when timeout runs.

lenemter (approved):
This PR doesn't fix a bigger performance problem with our ShadowEffect (e.g. when resizing a PiP) but it fixes the workspace switch performance. I think it makes sense to merge this.

leolost2605:
Yeah I agree it's less than perfect since it doesn't fix the root cause but it makes this specific animation a lot smoother so thanks :)

PR #2074, opened 2024-09-26
Introduce a GesturePropertyTransition

leolost2605:
Introduces a GesturePropertyTransition. This allows us to remove a lot of boilerplate code. For now it is a simple API that pretty much is used as a drop in replacement for connecting own handlers. In the future it can be easily expanded upon for more powerful usage. The nudge animations (both in multitasking view and while on the desktop) are a bit different now and IMHO a bit better (they can be changed back or to something different though).


src/Gestures/PropertyGestureTransition.vala
```diff
@@ -0,0 +1,98 @@
+
```
danirabbit, line 1: Missing copyright header :)

leolost2605:
Yeah that review request was a bit optimistic this still needs a few minor refinements:

- [ ] Window clones start from outdated positions
- [ ] Fix windowclone tooltips (also present in main but can be fixed in the process here)

leolost2605:
Ok this should now finally be ready :)

src/Gestures/GesturePropertyTransition.vala
```diff
@@ -0,0 +1,189 @@
...
+ * with easing without a gesture. Respects the enable animation setting.
+ */
+public class Gala.GesturePropertyTransition : Object {
+    public delegate void DoneCallback ();
+
+    /**
+     * Emitted when all animations are finished that is when the property has reached the target value
+     * either via gesture or via easing or combined.
+     */
+    public signal void done ();
```
lenemter, line 19: As far as I can see this signal is used only inside the class, the users of the class seem to use `done_callback`. Should this signal be private instead?

src/Gestures/GesturePropertyTransition.vala
```diff
@@ -0,0 +1,179 @@
...
+    public Value to_value { get; construct set; }
+
+    /**
+     * If not null this can be used to have an intermediate step before animating back to the origin.
+     * Therefore using this makes mostly sense if {@link to_value} equals {@link from_value}.
+     * This is mostly used for the nudge animations when trying to switch workspaces where there isn't one anymore.
+     */
+    public Value? intermediate_value { get; construct; }
+
+    private Value actual_from_value;
```
lenemter, line 46: I think this requires a comment on how this differs from `from_value`

src/Widgets/WindowClone.vala
```diff
@@ -290,68 +286,40 @@ public class Gala.WindowClone : Clutter.Actor {
...
-
-            set_size (width, height);
-            set_position (x, y);
 
-            window_icon.opacity = (uint) opacity;
             set_window_icon_position (width, height, scale, false);
             place_widgets ((int)width, (int)height, scale);
-
-            shadow_opacity = (uint8) opacity;
         };
```
lenemter, line 306: Is it possible to move this to one of the done callbacks?

src/WindowManager.vala
lenemter, line 1: The nudge animation is broken here :/

lenemter (changes requested):
Code looks good (apart from WindowClone but https://github.com/elementary/gala/pull/2082 should improve it a lot). There are some issues I mentioned in the review comments

src/Gestures/GesturePropertyTransition.vala
leolost2605: Ah right I wanted to add one but forgot it :sweat_smile:

src/Widgets/WindowClone.vala
leolost2605: Unfortunately not since it has to update during the animation

src/WindowManager.vala
leolost2605: Should be fixed now

lenemter (approved):
Good job :+1:

Issue #2080, opened 2024-09-20
Dock flickering


Letterus:
### What Happened?

Dock sometimes gets in "flickering" mode when using an app in full screen. When moving the cursor to the bottom of the screen it appears shortly and immediately disappears so you can't click an icon anymore.

### Steps to Reproduce

Hard to reproduce. It appears with apps in full screen mode (Firefox f.e.), but not always.

1. Open an app in full screen mode (f.e. Firefox).
2. Try to use the dock (it may be in "flickering mode")
3. You can stop the flickering mode by enabling the application menu once, f.e.

### Expected Behavior

Dock should always appear smoothly and stay to await my clicks.

### OS Version

8.x (Early Access)

### Software Version

Latest release (I have run all updates)

### Hardware Info

Thinkpad X280
Wayland session, 1.25x scale, i915 graphics driver.

leolost2605:
If we are talking about maximized here (which I assume we are since with fullscreen the dock shouldn't show at all) two ways I can currently reproduce this is the one fixed in #2081 and right after gala was started if the first launched window is maximized (although that's not 100% of the time but quite often)

leolost2605:
> if there is no window on the workspace if the first launched window is maximized (although that's not 100% of the time but quite often)

This should be fixed by #2070 although that's purely coincidental :)

Letterus:
It is! This really improves UX instantly! Thanks a lot!

PR #2081, opened 2024-10-07
Fix stacking order after workspace switch

leolost2605:
Starts addressing #2080 

On main a 100% way to reproduce #2080 is having a maximized window so that the dock is hidden then trying to switch workspace so that a nudge plays and then trying to show the dock again. That is fixed by this PR. However there are more ways to reproduce #2080 so it's not entirely fixed.

#2080 will be fixed by #2094 however this can/should still be merged because the current behavior is just wrong

### DIFF
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -2058,8 +2058,8 @@ namespace Gala {
                         continue;
                     }
 
-                    windows.prepend (actor);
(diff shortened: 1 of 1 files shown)


lenemter (approved):
Yep, using prepend here breaks the stack order. Makes sense. Good job :+1:

PR #2072, opened 2024-09-25
WindowClone: Only load after shown

leolost2605:
Fixes #2053 
Fixes #1898

Issue #2053, opened 2024-09-16
Window renders black


Letterus:
### What Happened?

I wanted to start Files, but switched to workspace view immediately. Files showed up, but was rendered black. It returns to graphics when doing a right-click on the app icon for example.

### Steps to Reproduce

1. Click the "Files" icon.
2. Immediately click the working space view icon before Files may show up.
3. Close the working space view.
4. Files is rendered black: 
![Bildschirmfoto von 2024-09-16 15 45 35@1x](https://github.com/user-attachments/assets/dc7414b8-a90e-424a-9c6e-b5a52ebfd758)
5. Right click on Files app icon returns fully rendered view: 
![Bildschirmfoto von 2024-09-16 15 45 42@1x](https://github.com/user-attachments/assets/59990891-6091-40a4-ba60-0697c87a9263)

### Expected Behavior

Files app should be rendered normally.

### OS Version

8.x (Early Access)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell
2024-09-16T15:48:35.746217+02:00 XinkPad280 gala[1833]: gala_drawing_canvas_set_size: assertion 'new_width >= -1 && new_height >= -1' failed
2024-09-16T15:48:35.746452+02:00 XinkPad280 gala[1833]: clutter_actor_allocate: assertion '!isnan (real_allocation.x1) && !isnan (real_allocation.x2) && !isnan (real_allocation.y1) && !isnan (real_allocation.y2)' failed
2024-09-16T15:48:35.746903+02:00 XinkPad280 gala[1833]: clutter_actor_allocate: assertion '!isnan (real_allocation.x1) && !isnan (real_allocation.x2) && !isnan (real_allocation.y1) && !isnan (real_allocation.y2)' failed
2024-09-16T15:48:35.747031+02:00 XinkPad280 gala[1833]: gala_drawing_canvas_set_size: assertion 'new_width >= -1 && new_height >= -1' failed
2024-09-16T15:48:35.747160+02:00 XinkPad280 gala[1833]: clutter_actor_allocate: assertion '!isnan (real_allocation.x1) && !isnan (real_allocation.x2) && !isnan (real_allocation.y1) && !isnan (real_allocation.y2)' failed
2024-09-16T15:48:35.747284+02:00 XinkPad280 gala[1833]: Can't update stage views actor unnamed [GalaMultitaskingView] is on because it needs an allocation.
[...]

Issue #1898, opened 2024-04-28
wayland: opening windows when overview is opened breaks a lot


leolost2605:
### What Happened?

Similar to what happened with wingpanel before the fix so might be related

### Steps to Reproduce

1. Open terminal
2. open a man page
3. go to overview and try closing terminal from there 
4. it will break because a dialog is opened by terminal about a running process

### Expected Behavior

Don't break

### OS Version

7.x (Horus)

### Software Version

Compiled from git

PR #2082, opened 2024-10-10
WindowClone: Layout in allocate virtual

### LINKED ISSUE #2088: Invisible window clones
Since #2072 some WindowClones are not visible (for me it's especially xwayland clients e.g. vs code). 
The issue temporarily fixed by #2072 is probably more severe (it can lead to everything breaking enough so that you have to restart gala and thus the session from a different tty) so I wouldn't revert for now.
But the root cause for that will be fixed in #2082 and thus that also reverts #2072 and fixes this issue.

### LINKED ISSUE #2128: Multitasking view does not show the application windows, only the icon
### What Happened?

In **classic session**, the multitasking view does not show the application windows, only the icon.

Classic session
Captura de tela de 2024-11-29 09 22 27

the same does not happen in **safe session**
Captura de tela de 2024-11-29 09 25 11

### Steps to Reproduce

1. in the dock, click on multitasking view
2. the multitasking view will show open applications
3. only the icons will appear

### Expected Behavior

Multitasking view should show the window image and the application icon.

### OS Version

8.x (Circe)

### Session Type

### Software Version

### Log Output

_No response_

### Hardware Info

notebook Asus i5 8gb ram Intel HD Graphics 4000
monitor resolution 1366x768

leolost2605:
Currently we are doing most of the layout management by hardcoding x, y, width and height of our child actors. However we can make the whole thing simpler and more robust by doing the layout management in the allocate virtual method as recommended by clutter docs.

Since the root cause for the following issues was that we didn't allocate some children on rare occasions which was fixed in this PR this
fixes https://github.com/elementary/gala/issues/2053 and
fixes https://github.com/elementary/gala/issues/1898 and therefore
closes #2072 

There are still some issues when windows spawn while multitasking view is open (like for example they reserve space but don't show) that I would like to tackle in follow up PRs but at least it doesn't completely break anymore as on current main :)

fixes #2088 by reverting #2072

Fixes #2128


danirabbit (changes requested):
I can confirm that this fixes the linked issues, but it looks like the textures (window clones, close buttons, app icons) are all rendered half a pixel offset so they're blurry (this is on LoDPI and Wayland)

danirabbit:
Tooltips also appear to be cut off and text looks pretty gnarly. Although when I went back to get a screenshot the other textures all look aligned correctly here :thinking: 

This branch:
Screenshot from 2024-10-11 10 47 41

main:
Screenshot from 2024-10-11 10 48 45

leolost2605:
@danirabbit the allocations are clamped to pixels now could you try again with that? I'm not so good at noticing those differences :sweat_smile: 
(and the tooltip thing I don't seem to be able to reproduce, probably requires a very specific window size and name :shrug:)

leolost2605:
So fedora is failing because for some reason it compiles with stricter rules (?) about implicit function declaration and `clutter_actor_get_relative_transformation_matrix` is defined in `clutter-mutter.h` and not `clutter.h` which we don't include anywhere. At least that's what I figured maybe I'm wrong somewhere :)

leolost2605:
Nvm I found a utility method that's much better to use in this case and properly exported :)

lenemter:
@leolost2605 Is this ready for review? I really want to include this in 8.0.4 release.

One thing I noticed is that clone no longer has magnifying filters? `clone.set_content_scaling_filters (Clutter.ScalingFilter.TRILINEAR, Clutter.ScalingFilter.TRILINEAR);` seems to fix this.

leolost2605:
@lenemter yes it's ready.

I've added the filters although TBH i don't really know what they do :sweat_smile:

tintou:
Opened https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/4156 for your missing includes…

leolost2605:
Thank you!

lenemter (changes requested):
Looks great, one small issue in the code, after that should be ready to go

lenemter (approved):

lenemter:
@danirabbit Can you review this branch please?

danirabbit (changes requested):
Everything looks sharp but it looks like this re-introduces the issue where windows that spawn while multitasking view is open aren't added. To reproduce:

1. Open screenshot and set a timed screenshot
2. Open multitasking view
3. when the timer runs out the new window will be transparent with no icon tooltip etc

leolost2605:
> There are still some issues when windows spawn while multitasking view is open (like for example they reserve space but don't show) that I would like to tackle in follow up PRs but at least it doesn't completely break anymore as on current main :)

From the description (the breaking refers to the version before the workaround was merged)

If it's a must I can do it here though :shrug:

danirabbit (approved):
Ah sorry I missed that. In that case 👍 😀

PR #2084, opened 2024-10-12
Introduce a wayland/x11 mutter hints protocol for monitor labels

leolost2605:
Currently we do monitor labels quite awkardly. First the display plug has to go via dbus to the daemon having to send color info, monitor index, label, x, y then the daemon spawns the labels and then gala looks at the window title and positions the label.

This can be made properly now by just using a wayland protocol/ setting the mutter hints on x11.

Goes with elementary/pantheon-wayland#3 and elementary/switchboard-plug-display#401

### DIFF
--- a/protocol/pantheon-desktop-shell-v1.xml
+++ b/protocol/pantheon-desktop-shell-v1.xml
@@ -159,5 +159,14 @@
 
       <arg name="dim" type="uint" summary="1 to dim, 0 to not dim"/>
     </request>
+
+    <request name="make_monitor_label">
+      <description summary="makes the surface a monitor label for the given monitor">
+        Request to make the surface a monitor label for the given monitor. The surface will be placed
+        in the top left corner of the monitor and will be kept above other surfaces.
+      </description>
+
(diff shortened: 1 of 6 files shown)


danirabbit (dismissed):
Since we're in the stable release phase we can add new API but we can't break old API until the next major version bump

leolost2605:
Ah right that is a thing oops :sweat_smile: 
This isn't very important just a better and slightly improved way to do it so it's obviously fine to wait
I think I'll factor out the WindowPositioner and the corresponding fix for centered windows though and prepare a separate PR for that (which, TBH, it probably should be anyway)

danirabbit:
@leolost2605 is it possible to add the new API without breaking the old API so we can move forward?

leolost2605:
@danirabbit finally got around to updating this so it should be ready without breaking the current API :)

zeebok (approved):
I dunno what was here before that was breaking API but _this_ code looks good to me

lenemter (approved):

Issue #2083, opened 2024-10-10
Autohide is too sensitive


dar5hak:
### What Happened?

When set to automatically hide (the default) the dock shows up even when the mouse pointer is ~40px above the bottom of the screen. This makes it difficult to click links at the bottom of a web page when it is maximized.

### Steps to Reproduce

1. Open a page with links at the bottom of the page, for example, https://proseandconst.xyz
2. Try to click the links. The dock shows up too early and obstructs the web page.

[Peek 2024-10-10 20-25.webm](https://github.com/user-attachments/assets/1cf2edb8-ebe0-4663-bbed-f426f0aaa92c)

### Expected Behavior

The dock should only show up when the pointer hits the bottom or near-bottom of the screen.

### OS Version

8.x (Early Access)

### Session Type

Classic Session (X11, This is the default)

### Software Version

Latest release (I have run all updates)

PR #2085, opened 2024-10-13
PanelWindow: Remove offset

lenemter:
Fixes https://github.com/elementary/gala/issues/2083

Not sure why it was added, hot corners work fine without the offset.


lenemter:
Oops. I got the offset purpose wrong. Sorry

PR #2086, opened 2024-10-13
Fix autohide

lenemter:
Fixes https://github.com/elementary/gala/issues/2083

HideTracker doesn't take into account that the dock window can be hidden and therefore cannot be hovered. This PR fixes it.

### DIFF
--- a/src/ShellClients/HideTracker.vala
+++ b/src/ShellClients/HideTracker.vala
@@ -175,7 +175,7 @@ public class Gala.HideTracker : Object {
     }
 
     private void toggle_display (bool should_hide) {
-        unowned var window_actor = (Meta.WindowActor ) panel.window.get_compositor_private ();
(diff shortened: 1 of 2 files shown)


leolost2605 (changes requested):
Makes sense but can we maybe add a comment about that being required for X11 since you wouldn't know what it's for if you're testing only on wayland.

src/ShellClients/HideTracker.vala
```diff
@@ -175,10 +175,12 @@ public class Gala.HideTracker : Object {
     }
 
     private void toggle_display (bool should_hide) {
+        unowned var window_actor = (Meta.WindowActor ) panel.window.get_compositor_private ();
```
leolost2605, line 178:
```suggestion
        unowned var window_actor = (Meta.WindowActor) panel.window.get_compositor_private ();
```

leolost2605 (approved):

PR #2070, opened 2024-09-23
GTK Background

leolost2605:
Fixes #1672 (while I can't reproduce that and it might already be fixed the backtrace there points to the background cache and source which is removed here so IMO it's pretty safe to assume it's fixed at the latest here)

TODO:

- [ ] Spanned mode
- [ ] Lots of testing (multiple displays, scale factors, etc.)
- [x] X11 support

PR #2087, opened 2024-10-13
Introduce a WindowPositioner

leolost2605:
Taken from #2084 and only prepares for more different positioned windows and fixes an issue that caused centered windows to sometimes not be centered and some small memory leaks (the Meta.Windows weren't freed due to the signal lambda using a local variable)

This will now also make the windows unmovable via super + drag.


src/ShellClients/ShellClientsManager.vala
```diff
@@ -166,37 +166,45 @@ public class Gala.ShellClientsManager : Object {
...
-        windows[window].set_hide_mode (hide_mode);
+        panel_windows[window].set_hide_mode (hide_mode);
     }
 
-    public void make_centered (Meta.Window window) {
-        if (window in centered_windows) {
-            return;
-        }
+    public void make_centered (Meta.Window window) requires (!is_itself_positioned (window)) {
+        positioned_windows[window] = new WindowPositioner (window, wm, (ref x, ref y) => {
```
danirabbit, line 187: I wonder is having a whole position func here kind of overkill? Do we need things other than like corners and center?

leolost2605: Corners, center, panel (so top and bottom) and maybe monitor corner without the padding of the monitor labels if we want to go with something like #2070.
I've got no strong preference here, we can move all types of "positioning funcs" into the positioner. It's not quite that straight forward because e.g. the monitor labels need additional info (the monitor index) but that can be done quite easily with a Variant or something (to keep it flexibel for the future :)).

leolost2605: I've moved the positioning logic to the positioner now to distribute the code more evenly :sweat_smile:

src/ShellClients/WindowPositioner.vala
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    construct {
+        window.stick ();
+
+        window.size_changed.connect (position_window);
+        window.position_changed.connect (position_window);
+        window.shown.connect (position_window);
+
+        var monitor_manager = wm.get_display ().get_context ().get_backend ().get_monitor_manager ();
```
lenemter, line 29:
```suggestion
        unowned var monitor_manager = wm.get_display ().get_context ().get_backend ().get_monitor_manager ();
```

lenemter (approved):
LGTM Makes sense and works as expected

PR #2091, opened 2024-10-15
HideTracker: Reveal panel on swipe

### LINKED ISSUE #1959: 1:1 Bottom edge swipe to reveal the dock
### Problem

On a tablet there's currently no way to reveal the dock when there are maximized or fullscreened windows

### Proposal

an edge swipe with one finger from the bottom edge of the display should reveal the dock. This should be a 1:1 gesture like our other multitouch gestures

### Prior Art (Optional)

iPad OS uses a bottom swipe to reveal the dock

leolost2605:
Fixes #1959 
Fixes elementary/dock#78

The dock gains focus and will stay until it is unfocused, e.g. by tapping another window.

### DIFF
--- a/src/ShellClients/HideTracker.vala
+++ b/src/ShellClients/HideTracker.vala
@@ -15,6 +15,8 @@ public class Gala.HideTracker : Object {
     public unowned PanelWindow panel { get; construct; }
     public Pantheon.Desktop.HideMode hide_mode { get; set; default = NEVER; }
 
+    private Clutter.PanAction pan_action;
+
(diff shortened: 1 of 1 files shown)


leolost2605:
I've started work on a bigger PR which lays groundwork for retiring touchegg by using the Clutter Gestures and also allows this to be one to one. It pretty much doesn't have anything in common with this PR but if we want to land the feature we can still merge it for now :shrug:

danirabbit (approved):
Yay, Definitely a huge improvement for tablet mode

PR #2094, opened 2024-10-16
Fix panel sometimes being behind windows

### LINKED ISSUE #2080: Dock flickering

leolost2605:
Fixes #2080

### DIFF
--- a/src/ShellClients/PanelClone.vala
+++ b/src/ShellClients/PanelClone.vala
@@ -63,6 +63,10 @@ public class Gala.PanelClone : Object {
             }
         });
 
+        // Make sure the actor is visible once it's focused FIXME: better event not only focused
+        // https://github.com/elementary/gala/issues/2080
+        panel.window.focused.connect (update_visible);
+
         update_visible ();
         update_clone_position ();
 
--- a/src/ShellClients/PanelClone.vala
+++ b/src/ShellClients/PanelClone.vala
@@ -77,6 +77,12 @@ public class Gala.PanelClone : Object {
 
     private void update_visible () {
         actor.visible = !panel_hidden;
+
+        if (actor.visible && !wm.get_display ().get_monitor_in_fullscreen (panel.window.get_monitor ())) {
+            // The actor has just been revealed, make sure it's at the top
+            // https://github.com/elementary/gala/issues/2080
+            actor.get_parent ().set_child_above_sibling (actor, null);
+        }
     }
 
     private void update_clone_position () {


lenemter (changes requested):
This fixes the issue with the Dock, but not if you log in -> open maximized window using Super + number -> press Super to open Applications menu, this results in application menu being below maximized window. I guess we need to apply the same fix somewhere else in the code?

leolost2605:
Yeah probably I'm just not sure where would be best. I guess we should connect to some event but which? We could do focus on the window but that only works because we know the wingpanel implementation and that it requests focus every time it goes out of it's reserved space :shrug:
Any ideas?

lenemter:
I have no idea either. I think we should add // FIXME here and just merge it. We can fix wingpanel issue later as it's not critical.

TomiOhl:
Note that even on "stock" Horus (and nothing compiled from source) the application menu is behind maximized windows, so it's not a new issue and not related to new changes

leolost2605:
> I have no idea either. I think we should add // FIXME here and just merge it. We can fix wingpanel issue later as it's not critical.

I did that now so it updates visible on focused which should fix the wingpanel issue for now

TomiOhl:
> > Note that even on "stock" Horus (and nothing compiled from source) the application menu is behind maximized windows, so it's not a new issue and not related to new changes
> 
> Hmm I don't remember that. Maybe you mean fullscreen windows so that wingpanel isn't visible at all?

Oh yes I mean fullscreen, when wingpanel isn't visible. Then if I press Super + space to open the application menu, it opens behind the window (I can search and launch apps via the keyboard). Doesn't happen if the app is just maximized and wingpanel is still visible, in that case the application menu appears above the window. Got a little confused by the terms, sorry

lenemter (approved):

PR #2103, opened 2024-10-31
Add panel hide delay

lenemter:
Fixes https://github.com/elementary/dock/issues/300

This required a small code refactor. All show/hide trigger logic (including barriers) now located in HideTracker. This ensures that all the hide/show signals don't interrupt each other.


teamcons:
Amazing ! Thank you :) I look forward to this !

alainm23 (approved):

src/ShellClients/HideTracker.vala
```diff
@@ -236,9 +276,75 @@ public class Gala.HideTracker : Object {
...
         if (delta_y < 0) { // Only allow swipes upwards
             panel.window.focus (pan_action.get_last_event (0).get_time ());
-            show ();
+            trigger_show ();
         }
 
         return false;
     }
+
+    private void destroy_barrier () {
```
leolost2605, line 285: I'm not sure haven't looked too much at it and can't test since I don't have a lot of time currently but this might need a destroy_barrier () call in the destructor

leolost2605 (changes requested):

src/ShellClients/HideTracker.vala
lenemter: Used prints to test if there is a memory leak and as far as I can see there is none, Gala.Barrier's destructor gets called by itself and Meta.Barrier gets destroyed.

leolost2605: Ahh right I forgot we are using Gala Barrier and not Meta Barrier here :)

src/ShellClients/HideTracker.vala
```diff
@@ -6,14 +6,30 @@
...
+    private Pantheon.Desktop.HideMode _hide_mode = NEVER;
+    public Pantheon.Desktop.HideMode hide_mode {
+        get {
+            return _hide_mode;
+        }
+        set {
+            _hide_mode = value;
+
+            destroy_barrier ();
+            if (hide_mode != NEVER) {
```
leolost2605, line 28: HideTracker will actually never exist with a NEVER hide mode. So maybe we can just remove all this here as well as the destroy_barrier method and connect setup_barrier to a notify["hide-mode"]? Would remove a lot of code and make it IMHO cleaner.

src/ShellClients/HideTracker.vala
```diff
@@ -196,23 +215,44 @@ public class Gala.HideTracker : Object {
...
+
+        hide_timeout_id = Timeout.add_once (HIDE_DELAY, () => {
             hide ();
-        } else {
-            show ();
+            hide_timeout_id = 0;
+        });
+    }
+
+    private void reset_hide_timeout () {
```
leolost2605, line 244: This should also be called in the destructor of this class or otherwise we might get a crash when e.g. revealing the dock to open settings and then switching to NEVER quick enough.

leolost2605 (changes requested):
A few things but apart from that looks good and thanks for the cleanup wanted to do it since the original got merged but never found time :)

src/ShellClients/HideTracker.vala
lenemter: I prefer using setters instead of notify in cases where we connect notify to itself. Using notify just moves the code to the constructor but it doesn't make it any cleaner IMO.

leolost2605: Well we would go from

```vala
    private Pantheon.Desktop.HideMode _hide_mode = NEVER;
    public Pantheon.Desktop.HideMode hide_mode {
        get {
            return _hide_mode;
        }
        set {
            _hide_mode = value;

            setup_barrier ();
        }
    }
```
to
```vala
public Pantheon.Desktop.HideMode hide_mode { get; set; }
…
notify["hide-mode"].connect (setup_barrier);
```
which is in the very least easier to read because you don't have to skip a lot of boilerplate code everytime you skim over the class IMHO :shrug:

src/ShellClients/PanelWindow.vala
```diff
@@ -152,16 +146,13 @@ public class Gala.PanelWindow : Object {
...
-        destroy_barrier ();
-
         if (hide_mode == NEVER) {
             make_exclusive ();
         } else {
             unmake_exclusive ();
-            setup_barrier ();
         }
+
+        clone.hide_mode = hide_mode;
```
leolost2605, line 155: Another thing I just noticed: Any reason this was moved down? That causes that struts don't update when changing to NEVER.

lenemter: Oops, I forgot to push the fix

src/ShellClients/HideTracker.vala
lenemter: The biggest issue I see in this approach is `...`. The property definition and what happens when you change it will be from 50 to 100 lines apart. For me that's counter intuitive and hard to read.

leolost2605: While this definitely makes sense and it isn't a no go for me I still prefer the notify (which is also used e.g. in PanelClone - admittedly also written by me but consistency etc. :shrug: ) so maybe we should just a let a third party decide? :)

teamcons:
Hello !
Is this one still in the works ?

leolost2605:
Yeah I think it's pretty much ready but we have a small disagreement on a style choice so we are kinda looking for a third party to decide :)

src/ShellClients/HideTracker.vala
stsdc: I understand the urge to remove the boilerplate code. However for me, the setter/getter approach is more readable right away. The notify approach is new for me and I need to read about it.

danirabbit: Another vote for the setter approach, if the property name ever changes there’s no compile time errors with notify so the setter way is safer imo

leolost2605: Then this is good to go from my side :)

leolost2605 (approved):

teamcons:
thank you <3

Issue #2064, opened 2024-09-22
eOS 8: Multitasking view on wayland gets stuck grabbing window


teamcons:
### What Happened?

Sometimes, when on the multitasking view (wayland) and grabbing a window to move it somewhere
grab gets stuck, and you cannot release without pressing escape

There does not seem to be a regularity to how often the issue occur, but sometimes it is quasi systematic

The bug can occur when clicking a window. Gala converts it into grabbing the window and refusing to release it.

### Steps to Reproduce

install eOS 8
go on wayland session
open a few windows
activate multitasking view
grab-release a few windows randomly
pointer will be stuck in grabbing mode at some point

### Expected Behavior

when releasing mouse, window gets released
when clicking on a window, the view switches to the window

### OS Version

8.x (Early Access)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell
Uncertain what to run. Cannot seem to replace a running gala like on x11
```

### Hardware Info

Intel® Core(TM) i3-7100U CPU @ 2.40GHz
Intel® HD Graphics 620

desiderantes:
I can reproduce this on the X11 session

PR #2104, opened 2024-11-01
DragDropAction: properly handle button release event

lenemter:
Fixes https://github.com/elementary/gala/issues/2064


leolost2605 (commented):
Seems to work can't really confirm the fix since it happens only irregularly for me.
An explanation would be nice though why this should work ~(because I don't really get it :grimacing:)~ ah nvm I guess it's because of grabbing an actor that might be below us in hierarchy when the drag actually starts

leolost2605 (approved):

lenemter:
@leolost2605 TBH I have no idea. I tried it and it just worked.

teamcons:
thank you so much

PR #2105, opened 2024-11-01
Introduce AnimationsSettings

lenemter:
Moves WindowManager.enable_animations to a separate AnimationsSettings namespace. There are a lot of objects that use wm only to get that value, in this case AnimationsSettings will simplify that. Also will help avoid passing window manager object all over the code.

Motivated by a duplicate enable_animations value in https://github.com/elementary/gala/pull/2074


leolost2605 (approved):
LGTM and very useful :)

PR #2106, opened 2024-11-02
AnimationSettings: Introduce get_animation_duration

leolost2605:
We very often use `AnimationSettings.get_enable_animations ? some duration : 0;`. This is now done by `AnimationSettings.get_animation_duration`. The method can also be expanded upon in the future e.g. by moving all hardcoded times to an enum and take that (e.g. expanding the AnimationDuration enum from constants and using that).


lib/AnimationsSettings.vala
```diff
@@ -22,4 +22,11 @@ namespace AnimationsSettings {
 
         return enable_animations;
     }
+
+    /**
+     * Utility that returns the given duration or 0 if animations are disabled.
+     */
+    public inline uint get_animation_duration (uint duration) {
+        return enable_animations ? duration : 0;
```
lenemter, line 30: We have to either use `get_enable_animations` here (because the value from gsettings hasn't been read yet) or create some `init` function that will initialize the gsettings read. In case we go with the second solution, we should then change all `AnimationsSettings.get_enable_animations ()` instances with `AnimationsSettings.enable_animations`.

lenemter (approved):

PR #2107, opened 2024-11-02
Pass Meta.Display instead of Gala.WindowManager where possible

lenemter:
A followup to https://github.com/elementary/gala/pull/2105


leolost2605 (approved):
Nice cleanup :)

PR #2109, opened 2024-11-03
Use WindowPositioner for panel positions

leolost2605:
This will also allow to do a bit better architecture in #2103 where instead of changing barriers on setting hide mode we change on changing anchor which is the actual reason we would need to resetup.


src/ShellClients/PanelWindow.vala
```diff
@@ -53,13 +41,24 @@ public class Gala.PanelWindow : Object {
...
+        var workspace_manager = display.get_workspace_manager ();
         workspace_manager.workspace_added.connect (update_strut);
         workspace_manager.workspace_removed.connect (update_strut);
+
+        window.size_changed.connect (update_strut);
+        window.position_changed.connect (update_strut);
+
+        window_positioner = new WindowPositioner (display, window, WindowPositioner.Position.from_anchor (anchor));
+
+        notify["anchor"].connect (() => {
```
leolost2605, line 58: Btw the notify here is also useful because we don't want the handler to run the first time it's set (i.e. with the constructor).

leolost2605:
I've been running this quite a bit now so it should be ready for review and testing.
Unfortunately we have to do a little workaround to avoid API breakage with wingpanel and dock but that can be removed in the next major release.

src/ShellClients/PanelWindow.vala
```diff
@@ -11,34 +11,22 @@ public class Gala.PanelWindow : Object {
     public WindowManager wm { get; construct; }
     public Meta.Window window { get; construct; }
```
lenemter, line 13:
```suggestion

```

src/ShellClients/PanelWindow.vala
```diff
@@ -49,13 +37,18 @@ public class Gala.PanelWindow : Object {
         delegate_actor = new DelegateActor ((Meta.WindowActor) window.get_compositor_private ());
         clone = new PanelClone (wm, this);
 
-        var monitor_manager = wm.get_display ().get_context ().get_backend ().get_monitor_manager ();
-        monitor_manager.monitors_changed.connect (() => update_anchor (anchor));
-        monitor_manager.monitors_changed_internal.connect (() => update_anchor (anchor));
+        var display = wm.get_display ();
```
lenemter, line 40:
```suggestion
        unowned var display = wm.get_display ();
```

src/ShellClients/PanelWindow.vala
```diff
@@ -49,13 +37,18 @@ public class Gala.PanelWindow : Object {
...
 
-        var workspace_manager = wm.get_display ().get_workspace_manager ();
+        var workspace_manager = display.get_workspace_manager ();
```
lenemter, line 42:
```suggestion
        unowned var workspace_manager = display.get_workspace_manager ();
```

src/ShellClients/ShellClientsManager.vala
```diff
@@ -226,8 +226,30 @@ public class Gala.ShellClientsManager : Object {
...
+                                parsed = RIGHT;
+                                break;
+
+                            default:
+                                break;
+                        }
+
+                        set_anchor (window, parsed);
+                        // We need to set a second time because the intention is to call this before the window is shown which it is on wayland
+                        // but on X the window was already shown when we get here so we have to call again to instantly apply it.
```
lenemter, line 252: If calling `set_anchor` twice is only required on X11 maybe wrap the second call in `if (!Meta.Util.is_wayland_compositor ())` ?

lenemter (changes requested):

src/ShellClients/WindowPositioner.vala
```diff
@@ -7,13 +7,28 @@
...
     public Meta.Display display { get; construct; }
     public Meta.Window window { get; construct; }
-    public Position position { get; private set; }
-    public Variant? position_data { get; private set; }
+    /**
+     * This may only be set after the window was shown.
+     * The initial position should only be given in the constructor.
+     */
+    public Position position { get; construct set; }
+    public Variant? position_data { get; construct set; }
```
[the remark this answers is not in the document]
leolost2605: I'm not sure what you mean here? The position passed to the constructor is used until a new one is set (which currently never happens, since we don't support changing panel positions)

src/ShellClients/ShellClientsManager.vala
leolost2605: The whole method is X only so I dont think it's necessary? 
But I can add it if you prefer :shrug:

lenemter: Oops I forgot about this

src/ShellClients/WindowPositioner.vala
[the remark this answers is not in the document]
lenemter: Ok sorry, I got confused because `position_window` is not called inside the constructor and I assumed it was only called when the position changed

lenemter (approved):

PR #2110, opened 2024-11-05
HideTracker: Take windows that are only vertically maximized into account

leolost2605:
Fixes elementary/dock#309


lenemter (approved):
Didn't know `in` works with flags. Fixes the issue.

PR #2120, opened 2024-11-22
Remove GTK usage and cleanup

leolost2605:
Towards dropping gtk3 dependency from the main gala


tintou:
I guess that we want more separate commits here :slightly_smiling_face:

Issue #2101, opened 2024-10-26
Dock interacts with the mouse even with the window overlapping


diogopessoa:
### What Happened?

Even with an application overlapping, the Dock reponds to the mouse. 

However, this only happens with the classic session (Xwindow). The issue doesn't occur in the secure session (Wayland).

Demo video:

https://github.com/user-attachments/assets/253b197c-b545-4314-ba8a-08fd44058e01

elementary OS 8 early access RC (classic session)

![Captura de tela de 2024-10-25 21 41 29](https://github.com/user-attachments/assets/be158102-751d-463f-846c-4979b8c70104)

### Steps to Reproduce

1. Maximize any application to overlap the dock;
2. The dock will hide;
3. Position the mouse in the dock area;
4. The name of the applications will apper and you can click to show or open them.

### Expected Behavior

When overlapping, the dock should not interact with the mouse.

### OS Version

8.x (Early Access)

### Session Type

Classic Session (X11, This is the default)

### Software Version

Latest release (I have run all updates)

### Hardware Info

Notebook Asus i5 
1280x720 screen

totchi-lagawi:
I had this issue in the demonstration version of elementary OS. It disappeared as soon as I installed it.

PR #2122, opened 2024-11-23
X11: move dock window out of bounds when hidden

lenemter:
Fixes https://github.com/elementary/gala/issues/2101

This is very very bad code but I just couldn't find other solution. So far I tried:
* Minimizing window
* Using `_NET_WM_STATE_HIDDEN` (which is basically minimizing AFAIK)
* Blocking X11 window events by changing its event mask
* Using XShape extension

Nothing worked. If you know any other solution please let me know.

### DIFF
--- a/data/gala.metainfo.xml.in
+++ b/data/gala.metainfo.xml.in
@@ -36,6 +36,7 @@
         </ul>
       </description>
       <issues>
+        <issue url="https://github.com/elementary/gala/issues/2101">Dock interacts with the mouse even with the window overlapping</issue>
       </issues>
     </release>
 
--- a/src/ShellClients/DelegateActor.vala
+++ b/src/ShellClients/DelegateActor.vala
@@ -4,7 +4,7 @@
  */
 
 /* This class is used to workaround https://github.com/elementary/gala/issues/2101 */
-public class Gala.DummyActor : GLib.Object {
+public class Gala.DelegateActor : GLib.Object {
     public const int OUT_OF_BOUNDS = 1000000;
 
     public Meta.WindowActor actor { get; construct; }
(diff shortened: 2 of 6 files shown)


leolost2605:
Tbh I think at some point we should just stop doing workarounds and instead go back to shipping plank for X since the whole infrastructure here (i.e. the new dock, the protocol, everything in /ShellClients) was never intended, tested or designed to work with X11 and just retrofitted into it with hopes it will work. 
And IMHO this whole thing just gets messier and messier and hard to maintain since always two scenarios have to be tested all at the risk of reducing quality for all sessions.
I'm not saying this point should be now but we should at least think about it.

lenemter:
@leolost2605 I don't agree with that. If we don't adapt gala's code to X11, the same bugs will happen with Wingpanel when we'll implement https://github.com/elementary/wingpanel/issues/53 for example. And suddenly we need to ship two versions of the same component

leolost2605:
Something like elementary/wingpanel#53 would be pretty much the only reason to support this and I think it's not worth polluting the codebase for that. Since there also doesn't seem to be a lot of movement regarding that I don't see it as a valid argument right now.
And even if something like elementary/wingpanel#53 would be agreed upon I'd argue that it can land in Wayland only because X11 can be considered deprecated and can hopefully be completely dropped very soon. So I don't think it's worth supporting it at the cost of a messy codebase especially since afaik apart from very specific edge cases Wayland provides a better experience as of today (tbh I also don't quite get why X is still the default for OS8).

alainm23:
Regardless of the fact that the solution at the code level may be poor, this issue is causing a lot of noise. Several users have already reported it, and it’s even worse because X11 is being sent by default. What do you think about this, @danirabbit ?

alainm23 (approved):

Issue #2119, opened 2024-11-21
Gala 8.0.3 fails to build with Mutter 45.5


pimvullers:
### What Happened?

Building gala on Gentoo with Mutter 45.5 failed with following error:
```
libmutter-13.vapi:1120.10-1120.14: error: The type name `pid_t' could not be found
 1120 | 		public pid_t get_pid ();
      | 		       ^~~~~            
../gala-8.0.3/lib/App.vala:161.23-161.28: error: The type name `pid_t' could not be found
  161 |     public GLib.SList<pid_t?> get_pids () {
      |                       ^~~~~~               
../gala-8.0.3/lib/App.vala:162.38-162.43: error: The type name `pid_t' could not be found
  162 |         var results = new GLib.SList<pid_t?> ();
      |                                      ^~~~~~     
Compilation failed: 3 error(s), 0 warning(s)
````

### Steps to Reproduce

Use Mutter 45.5, build gala

### Expected Behavior

Build passes

### OS Version

Other Linux

### Session Type

Classic Session (X11, This is the default)

### Software Version

Latest release (I have run all updates)

**Comment by pimvullers** [comment]
Comparing to Gala-8.0.2 the previous type was `int` instead of `pid_t`. Reverting this part with below command fixes build:
```
sed -i -e "s#pid_t#int#" vapi/libmutter.vapi lib/App.vala src/WindowTracker.vala
```

tintou:
I'm guessing it is because of your Vala version

tintou:
You need 0.56.17 to have the pid_t type

pimvullers:
Thanks, Gentoo is currently at 0.56.16 as stable. 0.56.17 is only available in unstable.
Might be good to have meson actually ensure proper vala version.

lenemter:
Fixed in https://github.com/elementary/gala/pull/2482

PR #2124, opened 2024-11-24
Smoother session start

leolost2605:
For easier review do the individual commits (or rather 87b7321 only since the other one is just #2109 squashed)

Rebase merge if merged before #2109

Ok this is now mostly for wayland because on X the window was already shown when we know it's a panel so we can't really say wait for anything.

On wayland the way it now works is we wait for all clients that have requested being a panel to show and then immediately reveal them in sync. As a fail safe the WM lets the ShellClientsManager know when the stage is ready which will start a 5 second timer which will reveal all clients that are ready. After the panels have been revealed the first time we won't do anything like that again until the next session start so e.g. when a panel crashes and restarts it will show as soon as it's ready.

Unfortunately I can't make a recording since it's at session start but if you try it you should definitely see it :)

### DIFF
--- a/src/PantheonShell.vala
+++ b/src/PantheonShell.vala
@@ -243,25 +243,7 @@ namespace Gala {
             return;
         }
 
-        Meta.Side side = TOP;
-        switch (anchor) {
-            case TOP:
-                break;
-
-            case BOTTOM:
-                side = BOTTOM;
-                break;
-
-            case LEFT:
-                side = LEFT;
-                break;
-
-            case RIGHT:
-                side = RIGHT;
-                break;
-        }
-
-        ShellClientsManager.get_instance ().set_anchor (window, side);
+        ShellClientsManager.get_instance ().set_anchor (window, anchor);
     }
 
     internal static void focus_panel (Wl.Client client, Wl.Resource resource) {
--- a/src/ShellClients/HideTracker.vala
+++ b/src/ShellClients/HideTracker.vala
@@ -152,7 +152,13 @@ public class Gala.HideTracker : Object {
         });
     }
 
-    private void update_overlap () {
+    /**
+     * Immediately updates the overlap. Usually {@link schedule_update} should be used instead which
+     * internally calls this but throttles it since it's a somewhat expensive operation.
+     * On very rare use cases however it is required to update immediately. In that case you can call
+     * this directly.
+     */
+    public void update_overlap () {
         overlap = false;
         focus_overlap = false;
         focus_maximized_overlap = false;
--- a/src/ShellClients/HideTracker.vala
+++ b/src/ShellClients/HideTracker.vala
@@ -16,17 +16,7 @@ public class Gala.HideTracker : Object {
     public Meta.Display display { get; construct; }
     public unowned PanelWindow panel { get; construct; }
 
-    private Pantheon.Desktop.HideMode _hide_mode = NEVER;
-    public Pantheon.Desktop.HideMode hide_mode {
-        get {
-            return _hide_mode;
-        }
-        set {
-            _hide_mode = value;
-
-            setup_barrier ();
-        }
-    }
+    public Pantheon.Desktop.HideMode hide_mode { get; set; }
 
     private Clutter.PanAction pan_action;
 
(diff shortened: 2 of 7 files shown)


leolost2605:
`sudo gala --replace` only works under X and not wayland. It should apply after logging out though.
How did you install it? Since AFAIK Gala doesn't build on current OS8

vjr:
I did the usual steps after cloning and checkout your branch:

`sudo apt build-dep gala`
`meson build --prefix=/usr`
`cd build`
`ninja`
`sudo ninja install`
`sudo gala --replace` (also did a relogin and also a full reboot)

Let me try this branch again with your latest changes.

vjr:
I grabbed the latest, looks like there's unresolved merge conflicts checked in but I just resolved them locally.

Why do you say it won't build on OS8? I am running the daily and it seems to build without errors.

Also, Wayland doesn't seem to work but when I choose X session this works and looks good to me (not a code review) so I am in favour of this PR - hopefully it can be made to work for Wayland too.

leolost2605:
> Why do you say it won't build on OS8? I am running the daily and it seems to build without errors.

Hmm for some reason I thought OS8 didn't have a recent enough vala, causing #2119 :shrug: 

> Wayland doesn't seem to work

Can you describe how it behaves? Do the panels just pop up without animation? Do they animate but not in sync? Something else?

leolost2605:
> I grabbed the latest, looks like there's unresolved merge conflicts checked in but I just resolved them locally.

Like on the latest commit?

vjr:
> > Wayland doesn't seem to work
> 
> Can you describe how it behaves? Do the panels just pop up without animation? Do they animate but not in sync? Something else?

Yep I meant it behaves as usual - both panel and dock pop up without delay and without animation.

vjr:
> > I grabbed the latest, looks like there's unresolved merge conflicts checked in but I just resolved them locally.
> 
> Like on the latest commit?

Yes, the latest commit at the time of my comment :-)

Are you not able to log in to Wayland session and observe yourself? I will only be able to come back to check this PR tomorrow though.

leolost2605:
> Yes, the latest commit at the time of my comment :-)

Hmm that's strange there shouldn't have been any and I don't remember any.. (I take it you mean an unresolved conflict commited in and not that it wasn't up to date with main?)

> Are you not able to log in to Wayland session and observe yourself? I will only be able to come back to check this PR tomorrow though.

I'm currently running it and just logged out and in again a few times but for me it works on wayland :shrug:

vjr:
> I'm currently running it and just logged out and in again a few times but for me it works on wayland 🤷

Oh my bad - it is working for Wayland too - it's just that my monitor goes blank for I guess more than the 1.5 seconds when I login so I was not able to see the new effect. I hardcoded the delay to 5 seconds in my local checkout and yes I can see it working on Wayland too. Login to X session also monitor goes blank but for shorter interval so I was able to see the new effect.

tintou:
Having an hard-coded empirical value that delays the startup is a no-go to me. Each application sends a "ready" signal to the session that we would have to acknowledge

stsdc:
What if my system has less RAM and components will load longer then 1,5s?

leolost2605:
> What if my system has less RAM and components will load longer then 1,5s?

> Each application sends a "ready" signal to the session that we would have to acknowledge

We are waiting for the clients to be ready anyways, the 1.5 was just because I saw somewhere a mockup with quite some delay and also to not stall if one of the clients is malfunctioning.
But fair enough I guess we can show immediately once clients are ready and introduce something like 5 seconds as backup to not stall if one client fails.

leolost2605:
Ok this is now mostly for wayland because on X the window was already shown when we know it's a panel so we can't really say wait for anything.

On wayland the way it now works is we wait for all clients that have requested being a panel to show and then immediately reveal them in sync. As a fail safe the WM lets the ShellClientsManager know when the stage is ready which will start a 5 second timer which will reveal all clients that are ready. After the panels have been revealed the first time we won't do anything like that again until the next session start so e.g. when a panel crashes and restarts it will show as soon as it's ready.

leolost2605:
Please don't merge main if it's intended to be rebase merged, instead rebase on main :)

leolost2605:
will be superseded by a new pr based on #2239

PR #2125, opened 2024-11-24
Shortcuts: Add an interactive screenshot shortcut

leolost2605:
Add a keyboard shortcut to launch the screenshot app.
Default is `Super` + `Print`

This is useful as an alternative to having to remember all of the shortcuts for screen, window, area and their clipboard variants or just in general if you prefer the interactive way.
As a bonus it also has the save dialog.
(As prior art: e.g. GNOME launches its interactive UI by default for the Print key, windows has snipping tool, etc.)

Needs to be added to the keyboard plug

### DIFF
--- a/data/gala.gschema.xml
+++ b/data/gala.gschema.xml
@@ -76,6 +76,11 @@
       <summary></summary>
       <description></description>
     </key>
+    <key type="s" name="interactive-screenshot-action">
+      <default>'flatpak run io.elementary.screenshot'</default>
+      <summary>Interactive screenshot action</summary>
+      <description>Sets the command to run when the interactive-screenshot keybinding is pressed.</description>
+    </key>
     <key type="b" name="move-maximized-workspace">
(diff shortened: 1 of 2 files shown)


stsdc:
obraz
I have this `super + Q`  to screenshot an area to clipboard.

stsdc (approved):

lenemter:
Please do not merge this. I'd really like to release 8.0.4 as a hotfux before the OS 8 release

lenemter (changes requested):
It doesn't work on OS 8. Screenshot is flatpak here.

1. You should change it to `flatpak run io.elementary.screenshot`
2. Maybe make command editable? Some user might want to use different screenshot tool and IMO we shouldn't hardcode it.

leolost2605:
Updated it :)

lenemter (approved):
Looks great and works as intended :tada:

PR #2133, opened 2024-12-03
HideTracker: Remove queued update in destructor

leolost2605:
Fixes a rare crash that occurred when a panel client was killed while an overlap update was queued.


lenemter (approved):
Makes sense

PR #2134, opened 2024-12-03
Cleanup `DelegateActor`

lenemter:
* `DelegateActor` -> `HidableWindow`, We might want to reuse the ability to hide windows on X11 to fix _ https://github.com/elementary/gala/issues/658, which got stalled because there was no way to hide window on X11.
* Expose methods instead of properties and reduce their visibility.
* Revert https://github.com/elementary/gala/pull/2086


leolost2605:
I'll continue to argue against committing to this approach because
1. #658 is a new feature which we can justify to bring to wayland only
2. I'm really, really looking forward to reverting this one day and currently it's just an easy git revert for a single commit
3. Judging from your commit history this might introduce a crash somewhere, we just haven't found it yet
4. For that reason if you really want it I'd argue for accepting some code duplication and do it in the pip plugin only so that if it brings a regression it's limited to pip and doesn't affect panels

Once again this is my opinion and i stand by it but if enough people think otherwise there is not much I can do :shrug:

lenemter:
1. I don't like justifying bringing a feature to Wayland only, we literally have code to do that, that works on both X11 and Wayland.
2. When we will be dropping X11 (I really hope this will happen very soon) the removal of this class will be very trivial and every method from this class we use can be replaced in 1-3 lines of code.
3. Solid point. I'm not good at that. I already got a number of crashes when developing this branch, I'll add more null checks just to be sure.
4. TBH I wanted to bring PiP plugin in-tree for a long time.

PR #2135, opened 2024-12-03
Use `panel-main-menu-action` to open applications menu

lenemter:


src/WindowManager.vala
```diff
@@ -410,10 +410,12 @@ namespace Gala {
         private void launch_action (string action_key) {
             try {
                 var action = behavior_settings.get_string (action_key);
-                if (action != null && action != "") {
```
stsdc, line 413: Is `action != null` the same as `action != ""`?

lenemter: No, I removed it because if `action` is empty it would produce a warning instead of failing silently.

PR #2141, opened 2024-12-05
WindowTracker: Search desktop file by process name

lenemter:
Fixes https://github.com/elementary/dock/issues/282
Fixes https://github.com/elementary/dock/issues/310 (TablePlus)

Gets process name from `/proc/%d/comm` and tries to get desktop file with that name. If it doesn't exist it tries to do the same with the parent process until we get to process 0.


src/WindowTracker.vala
```diff
@@ -63,7 +63,46 @@ public class Gala.WindowTracker : GLib.Object {
...
+
+            // this particular process doesn't have corresponding desktop file
+            // but maybe parent process has it
+
+            var stat_file = File.new_for_path ("/proc/%d/stat".printf ((int) pid));
+            var stat_dis = new DataInputStream (stat_file.read ());
+            var stats = stat_dis.read_line (null).split (" ");
+
+            if (stats.length < 3) {
+                warning ("WindowTracker.get_app_from_pid: stat file has wrong format");
```
leolost2605, line 89: Should there be a return here? (And in the next if)
Also you should probably be checking for < 4 here since you acces at index 3

leolost2605 (changes requested):

src/WindowTracker.vala
lenemter: Thanks for noticing that. Fixed!

lenemter:
This branch has a side effect. With https://github.com/elementary/dock/pull/330 if apps with no app_info launched from Terminal, gala thinks that the application belongs to the Terminal. Is this ok to proceed with this branch?

danirabbit:
I'm a bit worried this might also reintroduce the issue with duplicate launchers in the dock :/

lenemter:
@danirabbit our apps system (in both gala and dock) works on determining desktop file that belongs to an app. All this PR does is tries to match process name to a desktop file. If there is no duplicated desktop file, the bug will not appear.

I can't imagine an edge-case where two icons of the same app will appear

danirabbit:
Yeah I'm wondering in the case that there's a launcher but it can't be matched to the running window what will happen? Will it show a launcher for the panel or dock instead?

lenemter:
Well it fixes the issue for GParted and TablePlus. Not sure if there are other affected apps

leolost2605:
Ah ok then I guess it's kind of needed 🤷 
I don't have a strong opinion here then :)

PR #2142, opened 2024-12-06
X11: Use window shape to fix input for invisible windows

leolost2605:
The first commit just reverts #2122 

Rebase merge

Fixes the same issue as #2122 and is already prepared to be used as intended in #2134. Since it doesn't move anything we don't run into issues of having to solve conflicts between e.g. window positioner and the hidden window, as well as not having to have an additional class handle coordinates etc. which is IMHO quite a bit cleaner.

 Closes #2134


lib/Utils.vala
```diff
@@ -400,5 +400,50 @@ namespace Gala {
 
             return texture;
         }
+
+        private static HashTable<Meta.Window, X.XserverRegion?> regions = new HashTable<Meta.Window, X.XserverRegion?> (null, null);
+
+        public static void x11_set_window_pass_through (Meta.Display display, Meta.Window window) {
+            unowned var x11_display = display.get_x11_display ();
```
lenemter, line 407:
```suggestion
        public static void x11_set_window_pass_through (Meta.Window window) {
            unowned var x11_display = window.display.get_x11_display ();
```

lib/Utils.vala
```diff
@@ -400,5 +400,50 @@ namespace Gala {
...
+
+            var region = X.Fixes.create_region (xdisplay, {rect});
+
+            X.Fixes.set_window_shape_region (xdisplay, x_window, 2, 0, 0, region);
+
+            X.Fixes.destroy_region (xdisplay, region);
+        }
+
+        public static void x11_unset_window_pass_through (Meta.Display display, Meta.Window window) {
+            unowned var x11_display = display.get_x11_display ();
```
lenemter, line 428:
```suggestion
        public static void x11_unset_window_pass_through (Meta.Window window) {
            unowned var x11_display = window.display.get_x11_display ();
```

lenemter (changes requested):
Tried the same thing before https://github.com/elementary/gala/pull/2122 but it didn't work. Here it works flawlessly ¯\\\_(ツ)_/¯

lib/Utils.vala
leolost2605: Ah I always forget about that :)

lenemter (approved):
Thanks for looking into this <3

PR #2143, opened 2024-12-06
PiP: Mask window

### LINKED ISSUE #658: Make Picture-in-Picture "mask" the window instead of drawing a region

lenemter:
Fixes https://github.com/elementary/gala/issues/658
Based on https://github.com/elementary/gala/pull/1192


lenemter:
Ready for review

@danirabbit This requires a new icon for confirm button, I suck at inkscape so can you make one please :pray:

plugins/pip/SelectionArea.vala
```diff
@@ -1,117 +1,331 @@
...
+        var click_action = new Clutter.ClickAction ();
+#endif
+
+#if HAS_MUTTER49
+        click_action.recognize.connect (capture_selected_area);
+#else
+        click_action.clicked.connect (capture_selected_area);
+#endif
+
+        confirm_button = new Clutter.Actor () {
```
leolost2605: You should be able to use `Gala.Icon.from_resource` here. Should handle all of the pixbuf loading and displaying automatically

plugins/pip/SelectionArea.vala
```diff
@@ -120,52 +334,70 @@ public class Gala.Plugins.PIP.SelectionArea : CanvasActor {
...
 
-        int x, y, w, h;
-        get_selection_rectangle (out x, out y, out w, out h);
+        ctx.save ();
 
-        ctx.rectangle (x, y, w, h);
-        ctx.set_source_rgba (0.1, 0.1, 0.1, 0.2);
-        ctx.fill ();
+        var accent_color = Drawing.StyleManager.get_instance ().theme_accent_color;
+        ctx.set_source_rgba (accent_color.red, accent_color.green, accent_color.blue, 1.0);
```
leolost2605: This looks like the frame and handles should be in accent color? Unfortunately for me they are just white :(

leolost2605 (changes requested):
This looks really cool!
I left some comments on the code.

There also seems to be some miscalculation somewhere because when I don't select the whole window the area that is actually used as pip is offset (note that I'm using fractional scaling at 125%):

https://github.com/user-attachments/assets/6611e63b-53dd-4280-b457-1348f000ae52

Regarding design I personally feel like the confirm button is a bit too big but I'll leave that to UX :)

lenemter:
@leolost2605 Thanks for the review. Everything should be fixed now

leolost2605 (approved):
Code looks good to me and everything seems to be working great! This is really cool, nicely done!

PR #2096, opened 2024-10-16
Introduce a pan backend for touchscreen gestures

leolost2605:
The version of #2095 that will be prepared for merging.

- [ ] Cleanup commits

Rebase merge

PR #2150, opened 2024-12-08
ShellClients: Allow force hiding all panels

### LINKED ISSUE #2080: Dock flickering

leolost2605:
This is used by the multitasking view and supports hiding with gestures.
The gesture functionality also allows an easy way to implement #1959 once #2096 is merged.

This also has the advantage that we don't get inconsistencies when leaving the multitasking view by making sure that an autohide panel only reveals if it should be revealed.

Furthermore it prepares for an eventual shell group that we can reparent the panels to in order to fix #2080 for real, to be able to show the applications menu while a fullscreen app is open, keep dock and wingpanel interactive while in the overview (currently they show but cant be clicked on) and in the future keep the dock in the multitasking view, etc.


leolost2605:
See #2193

PR #2152, opened 2024-12-10
WindowClone: Fix invisible if shown while Multitasking view is open

### LINKED ISSUE #857: Toggling the active window's maximization state during multitasking view messes up the window preview size
## Prerequisites

## Describe the bug

I have my top-right corner screen corner set up to toggle the maximization state for the current window:

hot corners settings

This allows me to toggle the maximization state even while the multitasking view is active. This is not something I ever wish to do intentionally, but nevertheless it makes it possible to do so.

I noticed that when the active window's maximization state gets toggled this way the scale of the window in the multitasking view gets messed up, becoming either too zoomed in or too zoomed out. See the screen recording attached below for an example.

### To Reproduce
1. Set up a hot corner to toggle the active window's maximization state
2. Activate multitasking view
3. Activate the aforementioned hot corner
4. See a weirdly zoomed window

### Expected behavior
The window preview should still display correctly despite the window being resized during multitasking view.

An acceptable alternative would be to temporarily disable the "toggle maximization" hot corner while the multitasking view is active, as I don't think it ever makes sense to apply one operation while the other one is still ongoing, but that wouldn't cover the case of the window being resized without user action (perhaps the window decided to resize itself).

Maybe both things should happen.

### Screenshots or screen recordings

screen recording

### Logs

N/A

## Platform Information

- [x] I'm using the latest released stable version

Screenshot from 2020-06-22 23-48-42

## Additional context

N/A

leolost2605:
We now just connect to size changed signal and request reposition which we should be doing anyway. The stuff we did previously won't work on wayland because we don't have a window size and position yet when the actor becomes available so the given slot is nonsensical.

This brings no change for X neither compared to main nor to pre #2082 but fixes it for wayland.

Fixes #857

### DIFF
--- a/src/Widgets/WindowClone.vala
+++ b/src/Widgets/WindowClone.vala
@@ -112,6 +112,7 @@ public class Gala.WindowClone : Clutter.Actor {
         window.notify["fullscreen"].connect (check_shadow_requirements);
         window.notify["maximized-horizontally"].connect (check_shadow_requirements);
         window.notify["maximized-vertically"].connect (check_shadow_requirements);
+        window.size_changed.connect (() => request_reposition ());
 
         if (overview_mode) {
             var click_action = new Clutter.ClickAction ();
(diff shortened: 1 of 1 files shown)


lenemter (approved):
Makes sense and fixes the linked issue. Good job

PR #2155, opened 2024-12-11
WindowClone: Show small, transparent thumbnail during DND

### LINKED ISSUE #320: Icon animates in from wrong coordinate on HiDPI
To reproduce:

0. Use a HiDPI display
1. Maximize a window on an otherwise empty workspace
2. Open the Multitasking View (<kbd>⌘</kbd><kbd>▼</kbd>)

Expected: The icon animates in from the center

Actual: the icon animates in from the top-left, presumably loDPI coordinates (not multiplied by the scaling factor)

leolost2605:
Instead of just showing the application icon, we now shrink the whole clone to 30% size and make it a bit transparent when it is grabbed for a drag and drop. I couldn't figure out whether that existed at some point and was replaced with icon only for some reason but I find it more intuitive this way (also e.g. GNOME does it like this).

This also allows us to not special case the window icon position so we can drop a lot of code here.

Fixes #320

### DIFF
--- a/src/Widgets/WindowClone.vala
+++ b/src/Widgets/WindowClone.vala
@@ -159,8 +159,6 @@ public class Gala.WindowClone : Clutter.Actor {
     }
 
     private void reallocate () {
-        var window_frame_rect = window.get_frame_rect ();
-
         close_button = new Gala.CloseButton (monitor_scale_factor) {
             opacity = 0
         };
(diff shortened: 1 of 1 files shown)


leolost2605:
Ah that makes sense :sweat_smile: . I guess this is more of a step backwards then so I'll close it for now

PR #2158, opened 2024-12-11
WindowClone: Allocate window icon in allocate virtual

leolost2605:
This time with no major visual changes compared to main :)

### DIFF
--- a/src/Widgets/WindowClone.vala
+++ b/src/Widgets/WindowClone.vala
@@ -381,11 +381,13 @@ public class Gala.WindowClone : Clutter.Actor {
         var window_title_alloc = InternalUtils.actor_box_from_rect (window_title_x, window_title_y, window_title_width, window_title_height);
         window_title.allocate (window_title_alloc);
 
-        var window_icon_size = InternalUtils.scale_to_int (WINDOW_ICON_SIZE, monitor_scale);
-        var window_icon_x = (box.get_width () - window_icon_size) / 2;
-        var window_icon_y = box.get_height () - (window_icon_size * 0.75f);
+        float window_icon_width, window_icon_height;
+        window_icon.get_preferred_size (null, null, out window_icon_width, out window_icon_height);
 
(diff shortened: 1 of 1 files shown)


src/Widgets/WindowClone.vala
```diff
@@ -756,28 +701,12 @@ public class Gala.WindowClone : Clutter.Actor {
...
-            x = (int) Math.round (x);
-            y = (int) Math.round (y);
-        }
-
-        window_icon.set_size (size, size);
-        window_icon.set_position (x, y);
+        Timeout.add (duration, () => {
+            prev_parent.set_child_at_index (this, prev_index); // Set the correct index so that correct stacking order is kept
+            return Source.REMOVE;
+        });
```
lenemter, line 709: Can we use `clone.transitions_completed` here instead of timeout please?

src/Widgets/WindowClone.vala
```diff
@@ -448,6 +380,13 @@ public class Gala.WindowClone : Clutter.Actor {
 
         var window_title_alloc = InternalUtils.actor_box_from_rect (window_title_x, window_title_y, window_title_width, window_title_height);
         window_title.allocate (window_title_alloc);
+
+        var window_icon_size = InternalUtils.scale_to_int (WINDOW_ICON_SIZE, monitor_scale);
+        var window_icon_x = (box.get_width () - window_icon_size) / 2;
+        var window_icon_y = box.get_height () - (window_icon_size * 0.75f);
+
+        var window_icon_alloc = InternalUtils.actor_box_from_rect (window_icon_x, window_icon_y, window_icon_size, window_icon_size);
+        window_icon.allocate (window_icon_alloc);
```
lenemter, line 389: Is there a reason to calculate `window_icon_size` every time we allocate the clone? Can we use `window_icon.width` and `window_icon.height` instead?

leolost2605: I think the reason was that the scale factor might change e.g. if the window was on another monitor but is shown in a spread on the primary. But I guess since we only change size here it would look bad either way so let's just use the preferred size instead (we should maybe deal with the scale change somewhere else)

lenemter (approved):
LGTM :+1:

Issue #1202, opened 2021-08-11
Touchpad desktop switching gesture is sometimes unresponsive


validatedev:
## What Happened

Swiping three finger left-right behavior is buggy. Three finger gesture does not work when an animation was triggered before

## Expected Behavior

It should be interactible evertyime

## Steps to Reproduce

1. Swipe three finger right to left with touchpad
2. Immediately swipe left to right with touchpad
3. Gesture does not respond 2.

### Logs

Video for the issue:

https://user-images.githubusercontent.com/29679552/129089633-b05f25ab-f3ae-410d-8602-8a4cbeee8448.MOV

It works as expected with Fedora 34 + GNOME 40.x + X11 + touchegg + gnome-shell-extension-x11gestures:

https://user-images.githubusercontent.com/29679552/129089845-3104999a-61a6-4797-b63c-bf77e37f4dba.MOV

@JoseExposito you are the creator of that piece of system so supposing you know better than anyone else, hence mentioning you.

## Platform Information

elementary OS 6.0 10 August 2021 build.

validatedev:
Hi @JoseExposito 

First, I appreciate all of your work, not only in elementary OS, especially in terms of touchpads.

Actually, I did not fully understand the situation. Did you specifically make sure that the user does not make any gestures before the animation ends? If yes, why did you go this route? It's an incredibly uncomfortable way of using it, at least for me. Or that was a limitation in `gala`?

Thanks in advance.

JoseExposito:
Hi @validatedev, and thank you.

I faced some issues integrating multi-touch gestures in Gala and I couldn't make them work as you mentioned.

Reopening the issue as we can always improve them in the future :)

validatedev:
Thanks for the explanation! I hope that one day the issue is going to be fully resolved.

PR #2161, opened 2024-12-11
MultitaskingView: Allow workspace switch animation to be interrupted

leolost2605:
First step towards #1202 

Remaining are multitasking view open and close and while on the desktop workspace switch but those require the cancelling fixes and are left to follow up PRs.

TODO:

- [x] Some calculation corrections


lenemter (changes requested):
This branch is bugged after merging https://github.com/elementary/gala/pull/2164 :/

leolost2605:
@lenemter can you describe what happens? It seems to work fine for me :/

lenemter:
This bug happens when you interrupt workspace switch animation while switching to the last workspace

https://github.com/user-attachments/assets/9aa24538-5639-4ac5-a2e5-360b2c5afa37

leolost2605:
This should be ready now. 
Cancelling is now also much more predictably. We cancel if
- We haven't moved far enough and our velocity is to low
- We are going backwards with enough velocity

This fixes some unintuitive behavior like when switching to a new workspace and you're already there but you decide to go back and swipe fast backwards it won't cancel.

leolost2605:
I just realized this somehow cancels multitasking view open/close even though it shouldn't...
I'll look into it :)

leolost2605:
Ok it's fixed now so ready again :)

lenemter (approved):
This looks weird with touchegg backend but there's nothing we can do with it. Looking forward using this with touchpad backend

Issue #1140, opened 2021-05-18
Polishing change workspace gesture (eOS 6 Public Beta)


gabriel-ab:
## Problem

When trying to go to a left in workspace the first one you can't go to the right workspace by just swiping back.
Currently, if you try to do that, you must lift your fingers and do the other direction.

## Proposal

If in the first (or last) workspace when trying to change to an unavailable workspace, the gesture to the other direction is unlocked.

This is not totally essential, but would make the gesture of change workspace more polished.

## Prior Art

4 fingers gesture to change workspace in Windows 10

lenemter:
Fixed in https://github.com/elementary/gala/pull/2291 but requires new touchpad backend

Issue #1057, opened 2021-02-22
Animate workspace switching as one continuous 1:1 animation


hanaral:
## Problem

The 1:1 workspace switching only animates between the start workspace and the target workspace. The user has to lift their hand up and swipe again in the same direction if they want to see whether there are any more workspaces in that direction. This is also jarring for when you realise you are swiping to the wrong workspace and want to slide back and through the start workspace in to the next one over. 

## Proposal

The animation between and at the end of the workspace row should be a continuous 1:1 animation as if it was on a rail. MacOS and iOS do this exceptionally well.

## Prior Art

https://user-images.githubusercontent.com/58219504/108688815-3b40af00-74f0-11eb-9202-7550fb80f0eb.mp4

hanaral:
@JoseExposito Here is the issue I mentioned before - since the workspace overview has got multiple rendered at once shouldn't it be able to just track the fingers like a Hdy carousel?

JoseExposito:
Yes, it is technically possible, but it'll require some changes. Maybe something to implement in the future :+1:

lenemter:
Fixed in https://github.com/elementary/gala/pull/2291

PR #2164, opened 2024-12-13
Overshoot gestures

leolost2605:
This currently only works when using a non touchegg backend because the touchegg backend only sends progress between 0 and 1. This means that this only works when using two fingers to switch workspaces in the multitasking view.

This puts the required infrastructure in place to allow gestures to go as far as they want also in the opposite direction than they were started in. It's then used for the workspace switch in the multitasking view while clamping it to the respective number on workspaces on either side.

It also puts a system in place that when clamping to a fractional value this last bit will feel stretched to show that in this direction the gesture is bounded.

F*xes #1057 and #1140 but only in the multitaskingview due to the aforementioned issues with touchegg.

### DIFF
--- a/src/Gestures/GestureTracker.vala
+++ b/src/Gestures/GestureTracker.vala
@@ -187,15 +187,11 @@ public class Gala.GestureTracker : Object {
      * values not divisible by physical pixels.
      * @return The linear animation value at the specified percentage.
      */
-    public static float animation_value (float initial_value, float target_value, double percentage, bool rounded = false, double end_percentage = 0) {
+    public static float animation_value (float initial_value, float target_value, double percentage, bool rounded = false) {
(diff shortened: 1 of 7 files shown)


src/Gestures/GestureTracker.vala
```diff
@@ -88,12 +88,16 @@ public class Gala.GestureTracker : Object {
 
     /**
      * @param percentage Value between 0 and 1.
+     * @param completions How often a full cycle of the gesture was completed in this go. Can be
```
lenemter, line 91:
```suggestion
     * @param completions The number of times a full cycle of the gesture was completed in this go. Can be
```

lenemter (approved):
All the changes make sense to me and the overshoot does work with the scroll backend. Good job!

Issue #1178, opened 2021-07-10
Workspace Switching with Keyboard (meta) + Mouse scroll


kp2401075:
## Problem
Moving between workspaces when you're using mouse and keyboard to gather.
Currently the way to do this is wither open multi-tasking view(meta + s) and then move with mouse / scroll with mouse.

## Proposal
Direct switching of workspaces with holding down meta + scroll.
This will greatly improve workflow for those using mouse + keyboard.

## Prior Art
Not sure if any other system implements this.

I've searched through open issues could not find anything matching this.

If it is duplicate please close the issue.
Thanks:)

haeckle:
I (and presumably [others](https://elementaryos.stackexchange.com/q/17870)) totally second this. GNOME Shell has this built in ([Source](https://discourse.gnome.org/t/how-to-disable-super-mousewheel-switching-workspaces/8109/2)). However this will be conflicting #815

Issue #815, opened 2020-05-25
Zoom with Super + Scroll


techdev5521:
## Prerequisites
- [x] I have searched open and closed issues for duplicates.

## Feature
**Is your feature request related to a problem? Please describe.**

I'm visually impaired and would like to invoke the screen magnifier with <kbd>Ctrl</kbd> + Scroll like macOS so that I don't have to take my hand off my mouse to zoom in and zoom out as I change zoom levels several times a minute and this slows me down.

**Describe the solution you'd like**

macOS allows the magnifier to be invoked with <kbd>Ctrl</kbd> + Scroll. I'd like similar functionality.

**Existing work**

On Gnome Shell proper, this functionality was achieved with this project: https://github.com/tobiasquinn/gnome-shell-mousewheel-zoom 

**Describe alternatives you've considered**

- Continue using the keyboard shortcut

**Additional context**

I've been looking into tobiasquinn's project to understand how it works (for about 20 minutes) and am willing to contribute code when I understand both that project and Gala enough to do so. Any help is appreciated.

cassidyjames:
I think we'd do Super + Scroll since Super is window-manager-level stuff, and many apps have Ctrl + Scroll for zooming inside the app. But otherwise I'm :+1:

janxkoci:
This could also help avoid some locale-specific collisions: When my zoom works properly with the shortcut `Super`+`+`, then I can no longer use `Super`+`1` to switch to 1st workspace, because in Czech locale the symbols `1` and `+` are on the same key (the key above a letter `Q`). I currently experience a race condition after reboots when some sessions have that key set as `+` (so pressing the combo starts zooming) and other sessions have it set as `1` (so pressing the combo switches to 1st workspace).

cassidyjames:
For touchpads we do already support multi-touch zooming as well, through System Settings → Mouse & Touchpad → Gestures.

PR #2165, opened 2024-12-13
Add Super+Scroll actions

lenemter:
Fixes https://github.com/elementary/gala/issues/815
Fixes https://github.com/elementary/gala/issues/1178
Closes https://github.com/elementary/switchboard-plug-mouse-touchpad/issues/215

### DIFF
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -165,6 +165,11 @@ namespace Gala {
         }
 
         private void handle_super_scroll (uint32 timestamp, double dx, double dy) {
+            if (behavior_settings.get_enum ("super-scroll-action") != 1) {
+                super_scroll_triggered (timestamp, dx, dy);
+                return;
+            }
+
             var d =  dx.abs () > dy.abs () ? dx : dy;
 
             if (d > 0) {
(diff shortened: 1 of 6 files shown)


src/WindowManager.vala
```diff
@@ -157,6 +157,26 @@ namespace Gala {
                 });
             }
 #endif
+
+            var scroll_action = new SuperScrollAction (display);
+            scroll_action.triggered.connect (handle_super_scroll);
+
+            stage.add_action_full ("super-scroll-action", CAPTURE, scroll_action);
```
leolost2605, line 164: Currently all of the stage stuff and keybindings etc. is done in show_stage maybe we should do this there as well?

leolost2605 (changes requested):
I really really love this!

leolost2605 (commented):
On second thought I think it might be better to make the `void triggered ()` signal a `bool triggered ()`, return that for propagation and have everything that needs a super scroll action add their own. This would stay more in line with how events, etc. are generally handled and avoids having to add even more api to the windowmanager.
What do you think?

leolost2605:
> @leolost2605 we need to implement our own objects hierarchy for this to make sense

I'm not sure I understand what you mean here?

lenemter:
How are we gonna propagate the signal in this case? It's not an event, so we can't use Clutter... TBH I don't understand exactly what you mean in your original comment.

leolost2605:
I meant something like the last commit in #2182 does

PR #2166, opened 2024-12-13
Cleanup `InternalUtils.set_input_area`

lenemter:
Hot corners still work for me without this. I have no idea why we are doing this


leolost2605 (approved):
I can confirm hot corners still work.
However I don't know enough about X to know what's going on here so I can't say if this may break something we don't know about.
But as I may already have stated ( :grimacing: ) I don't really care about X so feel free to merge :shrug: 
Otherwise I guess we can just remove everything X in OS 9

PR #2168, opened 2024-12-15
Gestures: Add more details and introduce on_gesture_handled

leolost2605:
Add information about whether a gesture will be handled. This might be used to determine whether events are supposed to be consumed. Add information about the origin of a gesture. This might be used to determine whether a gesture should be handled (e.g. only allow swipe from screen edge). Add common GestureBackend interface that all backends have to implement. Add a on_gesture_handled signal to the GestureTracker which runs if and only if true was returned in on_gesture_detected. Before it is emitted prepare_gesture_handling will be called on the backend. This will be needed for the upcoming PanBackend to avoid conflicts in event consuming.
Add a utility get_action to the GestureSettings that gets the corresponding action for a gesture. This allows to remove some duplicated code.

Split out from #2096 
These changes are generic cleanup, get rid of some `META_CURRENT_TIME used to choose focus window` warnings and are required for a touchpad backend in gala as well.

Use rebase to merge

### DIFF
--- a/src/Gestures/Gesture.vala
+++ b/src/Gestures/Gesture.vala
@@ -35,9 +35,25 @@ namespace Gala {
     }
 
     public class Gesture {
+        public const float INVALID_COORD = float.MAX;
+
         public Clutter.EventType type;
         public GestureDirection direction;
         public int fingers;
         public Clutter.InputDeviceType performed_on_device_type;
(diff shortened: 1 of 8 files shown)


lenemter (approved):
LGTM

Issue #2175, opened 2024-12-19
Gala crashes when receiving a notification while switching workspaces


TomiOhl:
### What Happened?

Gala crashed when I got a notification while swiping from one workspace to another

### Steps to Reproduce

1. execute `sleep 2 & notify-send "test"` from terminal
2. start swiping on the touchpad to get to another workspace
3. when the notification arrives, gala crashes

### Expected Behavior

Gala should not crash, ideally the notification should be shown

### OS Version

8.x (Circe)

### Session Type

Secure Session (Wayland)

### Software Version

Latest release (I have run all updates)

### Log Output

Collapsed by default to save space, but here you go:

<details>
  <summary>Output</summary>

```shell
~$ coredumpctl debug gala
           PID: 2097 (gala)
           UID: 1000 (tomiohl)
           GID: 1000 (tomiohl)
        Signal: 11 (SEGV)
     Timestamp: Thu 2024-12-19 01:33:51 CET (14s ago)
  Command Line: /usr/bin/gala
    Executable: /usr/bin/gala
 Control Group: /user.slice/user-1000.slice/user@1000.service/session.slice/_io.elementary.gala@wayland.service
          Unit: user@1000.service
     User Unit: io.elementary.gala@wayland.service
         Slice: user-1000.slice
     Owner UID: 1000 (tomiohl)
       Boot ID: 3980b87110bd4752a47f767c46ae8a0e
    Machine ID: 7d97e773b25f1c6c771c218867546e27
      Hostname: ThinkPad-T16
       Storage: /var/lib/systemd/coredump/core.gala.1000.3980b87110bd4752a47f767c46ae8a0e.2097.1734568431000000.zst (present)
  Size on Disk: 5.8M
       Message: Process 2097 (gala) of user 1000 dumped core.
                
                Module libgcc_s.so.1 from deb gcc-14-14.2.0-4ubuntu2~24.04.amd64
                Module libstdc++.so.6 from deb gcc-14-14.2.0-4ubuntu2~24.04.amd64
                Module libzstd.so.1 from deb libzstd-1.5.5+dfsg2-2build1.1.amd64
                Module libudev.so.1 from deb systemd-255.4-1ubuntu8.4.amd64
                Module libsystemd.so.0 from deb systemd-255.4-1ubuntu8.4.amd64
                Stack trace of thread 2097:
                #0  0x00007000918ee482 n/a (libmutter-14.so.0 + 0xee482)
[...]

PR #2129, opened 2024-12-01
WindowManager: Only show notifications after their window was shown

leolost2605:
Fixes #2102 

The private window_created signal was recently introduced. @lenemter could you provide some background why that was done since it seems to work like this as well?

Issue #2102, opened 2024-10-27
Notifications sometimes result in segfault


djaeris:
### What Happened?

When printing, the segfault in the logs below happens. This happens sporadically — it happened to me twice while printing a PDF and also once in LibreOffice. I have no clue as to why it sometimes happens and sometimes it works normally. The document does print, though. But the session is over…

### Steps to Reproduce

1. Open PDF in Evince
2. Print it
3. You are back in the greeter

### Expected Behavior

Print. Don't crash.

### OS Version

8.x (Early Access)

### Session Type

Secure Session (Wayland)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell
okt 27 04:15:22 aerisbase gala[1841]: meta_window_set_stack_position_no_sync: assertion 'window->stack_position >= 0' failed
okt 27 04:15:22 aerisbase kernel: gala[1841]: segfault at 18 ip 000071ff53cee482 sp 00007ffc2e202090 error 4 in libmutter-14.so.0.0.0[71ff53c50000+165000] likely on CPU 3 (core 3, socket 0)
okt 27 04:15:22 aerisbase kernel: Code: f6 ff 66 0f 1f 44 00 00 e8 fb f0 ff ff 0f b6 83 9d 02 00 00 eb 86 66 90 f3 0f 1e fa 55 48 89 e5 41 56 41 55 41 54 41 89 f4 53 <48> 8b 47 18 48 89 fb 4c 8b 6f 30 4c 8b b0 a0 01 00 00 48 8b 07 ff
okt 27 04:15:22 aerisbase systemd[1]: Created slice system-systemd\x2dcoredump.slice - Slice /system/systemd-coredump.
okt 27 04:15:22 aerisbase systemd[1]: Started systemd-coredump@0-75504-0.service - Process Core Dump (PID 75504/UID 0).
okt 27 04:15:22 aerisbase systemd-coredump[75506]: Removed old coredump core.gala.1000.fb8656caabb840f3a48e5bf7576d552d.1695.1729591661000000.zst.
okt 27 04:15:22 aerisbase systemd-coredump[75506]: Removed old coredump core.eurotrucks2.1000.fb8656caabb840f3a48e5bf7576d552d.28075.1729602854000000.zst.
okt 27 04:15:25 aerisbase systemd-coredump[75506]: Process 1841 (gala) of user 1000 dumped core.
                                                   
[...]

PR #2183, opened 2024-12-19
WindowManager: Only show notifications after their window was shown

lenemter:
Fixes https://github.com/elementary/gala/issues/2175
Fixes https://github.com/elementary/gala/issues/2102

Closes https://github.com/elementary/gala/pull/2129

Now `InternalUtils.wait_for_window_actor` is basically the same as `Meta.Window.shown`

### DIFF
--- a/src/InternalUtils.vala
+++ b/src/InternalUtils.vala
@@ -355,5 +355,17 @@ namespace Gala {
                 return Source.REMOVE;
             });
         }
+
+        public static void wait_for_window_actor_to_show (Meta.Window window, owned WindowActorReadyCallback callback) {
+            unowned var window_actor = (Meta.WindowActor) window.get_compositor_private ();
+            if (window_actor.visible) {
+                callback (window_actor);
+                return;
+            }
+
+            InternalUtils.wait_for_window_actor (window, (actor) => {
+                actor.show.connect (() => callback (actor));
+            }); 
+        }
     }
 }
--- a/src/InternalUtils.vala
+++ b/src/InternalUtils.vala
@@ -365,7 +365,7 @@ namespace Gala {
 
             InternalUtils.wait_for_window_actor (window, (actor) => {
                 actor.show.connect (() => callback (actor));
-            }); 
+            });
         }
     }
 }
--- a/src/InternalUtils.vala
+++ b/src/InternalUtils.vala
@@ -358,7 +358,7 @@ namespace Gala {
                     });
                 }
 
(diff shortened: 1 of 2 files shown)


src/InternalUtils.vala
```diff
@@ -355,5 +355,17 @@ namespace Gala {
                 return Source.REMOVE;
             });
         }
+
+        public static void wait_for_window_actor_to_show (Meta.Window window, owned WindowActorReadyCallback callback) {
+            unowned var window_actor = (Meta.WindowActor) window.get_compositor_private ();
+            if (window_actor.visible) {
+                callback (window_actor);
+                return;
+            }
```
leolost2605, line 364: Shouldn't this part also be in the `wait_for_window_actor` callback?

src/InternalUtils.vala
```diff
@@ -355,5 +355,17 @@ namespace Gala {
...
+
+            InternalUtils.wait_for_window_actor (window, (actor) => {
+                actor.show.connect (() => callback (actor));
```
leolost2605, line 367: I think you have to disconnect this signal otherwise we might get a memory leak

leolost2605 (changes requested):

src/InternalUtils.vala
```diff
@@ -340,19 +340,25 @@ namespace Gala {
...
+                } else if (window_actor != null) {
+                    ulong show_handler = 0;
+                    show_handler = window_actor.show.connect (() => {
+                        window_actor.disconnect (show_handler);
+                        callback (window_actor);
+                    });
                 }
 
-                return Source.REMOVE;
+                return window_actor != null ? Source.REMOVE : Source.CONTINUE;
```
leolost2605, line 361: Is this required? The old version from the windowclone didn't do this :shrug: 
My concern is that we might get an infinite loop here if a window is created but never receives a window actor because it crashes before it can be assigned one (is that even possible)?

leolost2605 (changes requested):

src/InternalUtils.vala
lenemter: Looked into mutter's source code and seems like compositor private is set on`constructed`. Unfortunately this is not a signal so we cannot listen to it. But since it's `constructed` that means that there shouldn't be an infinite loop... but at the same time that means that `Source.CONTINUE` is not needed here because it's `constructed`, it can't take a long time to set compositor private...

leolost2605 (approved):
Fixes the issue. Good job and thanks for the in depth investigation!

PR #2192, opened 2024-12-24
ShellClients: Use translation y instead of clone and use gesturetransition for animation

leolost2605:
Split out from #2150 for easier review.

This introduces some inconsistencies when opening multitasking view and switching workspaces but these will be fixed properly by the follow ups from #2150.

The changes here include using the translation-y property of the window actor to hide it instead of making it invisible and animating a clone. Also included is the use of the GesturePropertyTransition for the actual animation and the related introduction of GestureTrackers. This is preparation for moving all panel hiding stuff to the shell clients (instead of the multitasking view doing its own thing with additional clones) as well as one to one touch gestures for showing the dock.

### DIFF
--- a/src/ShellClients/PanelClone.vala
+++ b/src/ShellClients/PanelClone.vala
@@ -32,8 +32,9 @@ public class Gala.PanelClone : Object {
 
     public bool panel_hidden { get; private set; default = true; }
 
-    private SafeWindowClone clone;
     private Meta.WindowActor actor;
+
+    private GestureTracker default_gesture_tracker;
 
     private HideTracker? hide_tracker;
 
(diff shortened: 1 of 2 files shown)


src/ShellClients/HideTracker.vala
lenemter, line 1: Can't we use `default_gesture_tracker` instead of passing gesture_tracker into update_overlap and update_hidden?

src/ShellClients/PanelClone.vala
```diff
@@ -136,39 +92,24 @@ public class Gala.PanelClone : Object {
...
-        clone.set_easing_duration (get_animation_duration ());
-        clone.y = calculate_clone_y (true);
-        clone.restore_easing_state ();
+        gesture_tracker.add_success_callback (with_gesture, () => panel_hidden = true);
     }
 
-    private void show () {
-        if (!panel_hidden) {
+    private void show (GestureTracker gesture_tracker, bool with_gesture) {
+        if (!panel_hidden || last_gesture_tracker.recognizing) {
```
lenemter, line 101: same here, PanelClone already has default_gesture_tracker, why do we pass it here again?

leolost2605:
@lenemter ok I've removed the gesturetracker passing from this commit so that it will make more sense in the commit history where we'll introduce it once we actually need it :)

src/ShellClients/PanelClone.vala
```diff
@@ -79,54 +64,22 @@ public class Gala.PanelClone : Object {
...
-            case TOP:
-            case BOTTOM:
-                return actor.x;
-            default:
-                return 0;
-        }
-    }
-
-    private float calculate_clone_y (bool hidden) {
+    private float calculate_y (bool hidden) {
```
lenemter, line 67: I think 'calculate_y' is ambiguous in this case because it calculates translation not coordinates, maybe 'calculate_translation_y' would be better?

leolost2605: Updated it :)

lenemter (changes requested):
I can partially reproduce https://github.com/elementary/gala/issues/2080. After starting the session open a new maximized window, the dock windows are placed behind said window. Probably requires same workaround as https://github.com/elementary/gala/pull/2094?

leolost2605:
@lenemter that will be completely fixed by #2195 
Would it be ok if we just merge this and then get #2195 in as well?

lenemter (approved):
:rocket:

PR #2193, opened 2024-12-24
WIP: Move panel hiding to shellclients from multitaskingview

### LINKED ISSUE #2130: Can't open the applications menu when on fullscreen
### What Happened?

If an app enters fullscreen mode, pressing the SUPER key will no longer open the applications menu.

### Steps to Reproduce

1. Open an application on fullscreen;
2. Press SUPER;
3. Bug happens.

### Expected Behavior

Applications menu shows up normally.

### OS Version

8.x (Circe)

### Session Type

Secure Session (Wayland)

### Software Version

### Log Output

_No response_

### Hardware Info

CPU: Dual Core Intel® Core(TM) i3-1005G1 CPU @ 1.20GHz
Graphics: Intel® UHD Graphics
RAM: 4 GB

leolost2605:
Includes #2192.

Same as #2150 but cleaned up for a rebase instead of squash for easier review

The additional changes here on top of #2192 are:

- Introduce a shell_group clutter actor that will be the parent of all shell UI WindowActors (e.g. panels, notifications, centered windows, etc.). This has the advantage that we have full control when and above what we show these. For example with this we can fix #2130, have wingpanel and dock interactable in the windowoverview (currently they show but you can't do anything with them) or eventually we can keep e.g. the dock visible and interactable in the multitasking view.
- Introduce a ShellClients.set_force_hide_panels to have all hiding/showing be done in the ShellClients instead of having the multitasking view do their own thing. This has to be done together with the shell group or else we get duplicate dock clones, etc.

Closes #2192 if merged before it.

### DIFF
--- a/src/Gestures/GestureTracker.vala
+++ b/src/Gestures/GestureTracker.vala
@@ -75,6 +75,8 @@ public class Gala.GestureTracker : Object {
      */
     public bool enabled { get; set; default = true; }
 
+    public bool recognizing { get; private set; }
+
     /**
      * Emitted when a new gesture is detected.
      * This should only be used to determine whether the gesture should be handled. This shouldn't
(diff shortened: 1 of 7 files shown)


leolost2605:
Converting to draft real quick because I think splitting the multitasking view change and the shell group change is probably worth it :)

leolost2605:
Superseded by #2239

PR #2146, opened 2024-12-07
Move window to old workspace when destroying the window

lenemter:
Fixes https://github.com/elementary/gala/issues/1986

Issue #551, opened 2019-06-13
Return to original workspace if a fullscreen window is closed and there is no more windows open


Blast-City:
Situation 1:
1- Open Firefox and go to Youtube.
2- Press Shift+Click on a video link to open a video in a new window.
3- Doube click the video to go fullscreen.
4- Press Alt+F4 to close the window.

**Result:** The workspace switches automatically to the original one after closing the window.

Situation 2:
1- Open Firefox and go to Youtube.
2- Press Shift+Click on a video link to open a video in a new window.
3- Press F11 to make the window fullscreen.
4- After a new workspace is created, double click on the video to make it fullscreen.
4- Press Alt+F4 to close the window.

**Result:** The workspace stays selected even thought is empty.

What behaviour is the normal one? Should the workspace switch back to the original workspace every time that a fullscreen window - that created a new workspace - is closed?
Is the first situation a Firefox bug or are this two situations not cohesive?

Since leaving the fullscreen state (F11) also returns the window to the original workspace, why not do the same for closing to maintain cohesion thought out all apps?

Edit: This behaviour can also be seen on Steam's trailers and video player. Closing fullscreen windows makes the workspace switch back to the original one.

PR #2194, opened 2024-12-24
Add fullscreen effect and move to old ws on close

### LINKED ISSUE #1986: Does not return to the first virtual desktop after exiting fullscreen mode.
### What Happened?

After last update system, something went wrong. Does not return to the first virtual desktop after exiting full screen mode. I use telegram, open some picture or video. It's opened in full screen mode, and in new virtual desktop. But after close pics or video in full screen mode.  I'm staying on the new virtual desktop, rather than returning to the main one where the application I'm using is located. because of this, it returns to the main desktop by pressing the win+'right arrow' key.

### Steps to Reproduce

1.use any application that uses the transition to full-screen mode (example telegram app).
2.open telegram. open picture or video, it's opens in full screen, on a new virtual desktop.
3.after close picture or video, there is no return to the main desktop.
4.to return to the main virtual desktop, you have to press the win+arrow key combination.

### Expected Behavior

Before the update, everything was ok. after closing the viewing of images or videos in telegram. the return to the main work took place by itself, without using hotkeys

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

 OS: elementary OS 7.1 Horus x86_64 
Host: Z390 AORUS MASTER 
 Kernel: 6.5.0-41-generic 
Uptime: 36 mins 
Packages: 1970 (dpkg), 40 (flatpak) 
Shell: bash 5.1.16 
  Resolution: 2560x1440 
 DE: Pantheon 
 WM: Mutter(Gala) 
 Theme: io.elementary.stylesheet.blueberry [GTK3] 
  Icons: elementary [GTK3] 
  Terminal: io.elementary.t 
 CPU: Intel i9-9900K (16) @ 5.000GHz 
 GPU: NVIDIA GeForce RTX 3080 Lite Hash Rate 
  Memory: 3231MiB / 32016MiB 

Снимок экрана от 2024-07-14 11 38 00

leolost2605:
Fullscreening/Unfullscreening will now play the same effect as maximizing and unmaximizing (since it's pretty much just an extended version of that action I think it fits)

Also we will now move to the old ws if a window that was moved to a new ws is closed.

Fixes #1986
Fixes https://github.com/elementary/gala/issues/551
Closes #2146

### DIFF
--- a/data/gala.metainfo.xml.in
+++ b/data/gala.metainfo.xml.in
@@ -39,8 +39,10 @@
       </description>
       <issues>
         <issue url="https://github.com/elementary/gala/issues/397">Prevent PIP overlapping wingpanel?</issue>
+        <issue url="https://github.com/elementary/gala/issues/551">Return to original workspace if a fullscreen window is closed and there is no more windows open</issue>
         <issue url="https://github.com/elementary/gala/issues/857">Toggling the active window's maximization state during multitasking view messes up the window preview size</issue>
         <issue url="https://github.com/elementary/gala/issues/1967">Some apps ignore HiDPI mode</issue>
+        <issue url="https://github.com/elementary/gala/issues/1986">Does not return to the first virtual desktop after exiting fullscreen mode.</issue>
         <issue url="https://github.com/elementary/gala/issues/2088">Invisible window clones</issue>
         <issue url="https://github.com/elementary/gala/issues/2113">gnome-session-x11-services-ready.target isn't started on Wayland session</issue>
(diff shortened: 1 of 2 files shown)


leolost2605:
This is now actually ready :)

The reason we connect to unmanaged and don't do it in destroy () anymore is that the window list only gets fully updated once a window is finished unmanaging. Otherwise we might try to remove a workspace but mutter thinks it has still windows on it which leads to a bail out.

lenemter (approved):
Makes sense and fixes the issue :+1:

PR #2195, opened 2024-12-25
Introduce a shell group

leolost2605:
Since moving the panel hiding from the multitasking view completely to the shell clients and introducing a shell group are bigger changes, I think they probably should be separated which is what this PR does.

It still includes the new version of #2192 but now only introduces the shell group which is (taken from the code comment):

> The group that contains all WindowActors that make shell elements, that is all windows reported as 
ShellClientsManager.is_positioned_window and notifications, as well as manually added elements.
         It will (eventually) never be hidden by other components and is always on top of everything. Therefore elements are
         responsible themselves for hiding depending on the state we are currently in (e.g. normal desktop, open multitasking view, fullscreen, etc.).

This will allow us to fix https://github.com/elementary/gala/issues/2130, have wingpanel and dock interactable in the windowoverview (currently they show but you can't do anything with them) or eventually we can keep e.g. the dock visible and interactable in the multitasking view.

There are no functional changes apart from the side effect that we get a nice hide animation for the panels when entering fullscreen.

This will now be required for moving all hiding logic to shell clients from the multitasking view (#2193) which will allow us to do some nice consistency improvements :)

### DIFF
--- a/src/InternalUtils.vala
+++ b/src/InternalUtils.vala
@@ -388,5 +388,15 @@ namespace Gala {
             Clutter.get_default_backend ().get_default_seat ().bell_notify ();
 #endif
         }
+
+        public static void update_transients_visible (Meta.Window window, bool visible) {
+            window.foreach_transient ((transient) => {
+                var actor = (Meta.WindowActor) transient.get_compositor_private ();
+
+                actor.visible = visible;
+
+                return true;
+            });
+        }
     }
 }
--- a/src/InternalUtils.vala
+++ b/src/InternalUtils.vala
@@ -391,7 +391,7 @@ namespace Gala {
 
         public static void update_transients_visible (Meta.Window window, bool visible) {
             window.foreach_transient ((transient) => {
-                var actor = (Meta.WindowActor) transient.get_compositor_private ();
+                unowned var actor = (Meta.WindowActor) transient.get_compositor_private ();
 
                 actor.visible = visible;
 
--- a/src/ShellClients/PanelClone.vala
+++ b/src/ShellClients/PanelClone.vala
@@ -8,7 +8,7 @@
 public class Gala.PanelClone : Object {
     private const int ANIMATION_DURATION = 250;
 
-    public WindowManager wm { get; construct; }
+    public WindowManagerGala wm { get; construct; }
     public unowned PanelWindow panel { get; construct; }
 
     public Pantheon.Desktop.HideMode hide_mode {
@@ -38,7 +38,7 @@ public class Gala.PanelClone : Object {
 
     private HideTracker? hide_tracker;
 
-    public PanelClone (WindowManager wm, PanelWindow panel) {
+    public PanelClone (WindowManagerGala wm, PanelWindow panel) {
         Object (wm: wm, panel: panel);
     }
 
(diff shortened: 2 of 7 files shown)


leolost2605:
There were some minor changes made (renaming) in #2192 that I'm to lazy to replicate here so merge that first and I'll just drop the commit here then :)

src/InternalUtils.vala
```diff
@@ -388,5 +388,15 @@ namespace Gala {
             Clutter.get_default_backend ().get_default_seat ().bell_notify ();
 #endif
         }
+
+        public static void update_transients_visible (Meta.Window window, bool visible) {
+            window.foreach_transient ((transient) => {
+                var actor = (Meta.WindowActor) transient.get_compositor_private ();
```
lenemter, line 394:
```suggestion
                unowned var actor = (Meta.WindowActor) transient.get_compositor_private ();
```

src/ShellClients/PanelClone.vala
```diff
@@ -38,14 +38,16 @@ public class Gala.PanelClone : Object {
...
+    public PanelClone (WindowManagerGala wm, PanelWindow panel) {
         Object (wm: wm, panel: panel);
     }
 
     construct {
         default_gesture_tracker = new GestureTracker (ANIMATION_DURATION, ANIMATION_DURATION);
 
         actor = (Meta.WindowActor) panel.window.get_compositor_private ();
+        actor.get_parent ().remove_child (actor);
+        wm.shell_group.add_child (actor);
```
lenemter, line 50: why not use `InternalUtils.clutter_actor_reparent`?

src/WindowManager.vala
```diff
@@ -1162,6 +1168,20 @@ namespace Gala {
             show_window_menu (window, menu, rect.x, rect.y);
         }
 
+        private void check_shell_window (Meta.WindowActor actor) {
+            unowned var window = actor.get_meta_window ();
+            if (ShellClientsManager.get_instance ().is_positioned_window (window)
+                || NotificationStack.is_notification (window)
+            ) {
+                actor.get_parent ().remove_child (actor);
+                shell_group.add_child (actor);
```
lenemter, line 1177: same here

src/WindowManager.vala
```diff
@@ -2222,10 +2226,14 @@ namespace Gala {
...
             animating_switch_workspace = false;
+
+            return true;
         }
 
         public override void kill_switch_workspace () {
-            end_switch_workspace ();
+            if (end_switch_workspace ()) {
+                switch_workspace_completed ();
+            }
```
lenemter, line 2236: why do we have that here? it looks unrelated to the topic of this PR

lenemter (changes requested):

src/WindowManager.vala
leolost2605: I'm sorry that change was actually a remnant from some testing I did. 
However the other change in end_switch_workspace () is required because now we can actually end up in a situation where windows and their parents is null. However just returning was wrong there (which wasn't noticed previously because we always had wingpanel and dock).

(FYI this will be completely removed in #2225 anyways)

lenemter (approved):
Code looks good and from my minimal testing works as expected

lenemter:
CI is failing because of ubuntu repos

PR #2202, opened 2025-01-01
DesktopIntegration: new method to close window used by dock

moises2657:
Implementation of close_window method on Gala.DesktopIntegration to solve the issue: https://github.com/elementary/dock/issues/278 
https://github.com/elementary/dock/pull/344

### DIFF
--- a/src/DesktopIntegration.vala
+++ b/src/DesktopIntegration.vala
@@ -183,4 +183,28 @@ public class Gala.DesktopIntegration : GLib.Object {
 
         wm.window_overview.open (hints);
     }
+
+    public void close_window (uint64 uid) throws GLib.DBusError, GLib.IOError {
+
+        var apps = Gala.AppSystem.get_default().get_running_apps();
+
+        foreach (unowned var app in apps) {
+
+            foreach (weak Meta.Window window in app.get_windows()) {
+
+                if (window.get_id() == uid) {
+
+                    window.delete(0);
+
+                    return;
+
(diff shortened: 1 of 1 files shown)


leolost2605 (changes requested):
Hey, thanks for this!
The code looks good only some stylistic nits :)
Most notably you have a lot of empty new lines here which usually only should be used to separate unrelated code, so I think you can remove most of them. Also the linter will probably throw some errors, you can check the workflow for the exact lines and what's wrong there.

teamcons:
https://github.com/elementary/dock/issues/278 is about closing an app in background, not just windows, though

src/DesktopIntegration.vala
```diff
@@ -244,4 +244,17 @@ public class Gala.DesktopIntegration : GLib.Object {
...
+        var apps = Gala.AppSystem.get_default ().get_running_apps ();
+        foreach (unowned var app in apps) {
+            foreach (weak Meta.Window window in app.get_windows ()) {
+                if (window.get_id () == uid) {
+                    window.delete (Meta.CURRENT_TIME);
+                    return;
+                }
+            }
+        }
+        throw new GLib.IOError.NOT_FOUND ("Window not found");
```
lenemter, line 258:
```suggestion
        }

        throw new GLib.IOError.NOT_FOUND ("Window not found");
```

lenemter (approved):
Looks good to me!

PR #2210, opened 2025-01-05
WindowClone: allow close window action on mice only

lenemter:
Follow up to https://github.com/elementary/gala/pull/2327


leolost2605:
I don't know whether we should do this. I personally use both gestures (three finger tap to close and three finger switch) every day without a problem (even if I start the swipe gesture while over a window in the multitasking view). So this really might be hardware/something else specific.
Just disabling it would be a big regression for me because i really like the three finger tap to close.
I think we should at least wait for input from @elementary/ux

TomiOhl:
Another workaround for the issue is to use four finger swipe for workspace switching, like on Windows. I have it set up like that so I have never encountered the issue. But I see that changing that default setting would also cause confusion so it's equally undesired.

(also noting that even if this gets merged, I guess two finger swipe will still work)

wpkelso:
[This poll on Mastodon](https://mastodon.online/@mirlo/113788492774143170) seems to imply that overwhelmingly, people use the close button over three-finger tap. @danirabbit also provided this data point in discord for context, taken from an earlier survey:
image

However, the replies to the fedi poll seem to imply that some people don't know about the three-finger tap close functionality, and when they learn of it they see it as a positive.  It's clear we shouldn't do away with the functionality.

I think for this solving this specific issue, a switch is probably appropriate. If a user finds it to be an issue, they can turn it off, but otherwise it's there if you want it, and for those that turn the tap functionality off it probably won't break their muscle memory like needing to change swipes to 4-finger would.

@TomiOhl does bring up an interesting point to me though: defaulting to using 4-fingers to control the workspace view, and 3-fingers to control applications within workspaces could be a good approach to explore. It would definitely break most muscle memory, but I could see a system like that working well, and also giving us the flexibility to do things like have 3-finger swipes move individual applications that are hovered over to the next or previous workspace, inside or outside the multitasking view. Perhaps a toggle to switch between which number of fingers controls what action? Given Windows uses 4-fingers for the workspace view, I think it's reasonable to expect that a wide number of touch control devices would be able to detect 4-finger inputs.

danirabbit:
Ah see I'm kind of inclined to think we should just remove it if only 26% of people close from the multitasking view and of those people only 13% do it via 3 finger tap, that's like 3%. That's an extremely small group compared to the potential for accidental loss here imo.

I do agree that some of the comments indicate that folks are interested in *a way* to close applications via gestures though. Maybe we could explore something a little harder to accidentally trigger like the swipe up gesture on iOS, iPad OS, and Phosh?

wpkelso:
> I do agree that some of the comments indicate that folks are interested in *a way* to close applications via gestures though. Maybe we could explore something a little harder to accidentally trigger like the swipe up gesture on iOS, iPad OS, and Phosh?

I think that would fit into the model I proposed pretty well. 

e.g. 
* 3-finger horizontal -> show next/previous workspace
* 3-finger vertical -> open/close view
* 4-finger horizontal -> swap window to next/previous workspace
*4-finger vertical -> close application

danirabbit:
I think for swiping application windows in the multitasking view folks would expect to use one finger on touch screen, which usually means two fingers on touchpad

danirabbit (approved):
Works as expected and makes sense in combination with swipe to close branch

PR #2213, opened 2025-01-05
Cleanup WindowManager

lenemter:
1. Move `clutter_actor_reparent` into InternalUtils
2. Move handling screenshot shortcuts into ScreenshotManager

Don't merge until 8.1.0 is released


leolost2605:
Maybe we can make ScreenshotManager not a singleton anymore? Like just have a field in WindowManager that keeps a reference on it and hand it to DBus.init

leolost2605:
Also since these are pretty unrelated can you clean up the commits to something like a clutter_actor_reparent and a screenshotmanager commit so that we can use rebase to merge? (Or alternatively propose two PRs)

PR #2218, opened 2025-01-08
Remove Clutter.get_default_backend usage

### LINKED ISSUE #2217: Stop using Clutter.get_default_backend
### What Happened?

Currently, Gala uses `Clutter.get_default_backend ()` in various places to either get a `Cogl.Context` or a `Clutter.Seat`. That helper will go away and compositors should go through the `Clutter.Context` associated to every `Clutter.Actor` and pass it around. See https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/3528 for an example.

The helper will probably not gonna be removed in next release and will just be deprecated but moving away from it already would help us getting there "faster".

### Steps to Reproduce

No steps

### Expected Behavior

See above.

### OS Version

8.x (Circe)

### Session Type

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

leolost2605:
Fixes #2217 

Use rebase to merge

### DIFF
--- a/lib/Drawing/Canvas.vala
+++ b/lib/Drawing/Canvas.vala
@@ -9,6 +9,8 @@ public class Gala.Drawing.Canvas : GLib.Object, Clutter.Content {
     private int height = -1;
     private float scale_factor = 1.0f;
 
+    private Cogl.Context? cogl_context;
+
     private Cogl.Texture? texture = null;
     private Cogl.Bitmap? bitmap = null;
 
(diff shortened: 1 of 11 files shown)


lib/Drawing/Canvas.vala
```diff
@@ -9,20 +9,33 @@ public class Gala.Drawing.Canvas : GLib.Object, Clutter.Content {
...
+#else
+        cogl_context = Clutter.get_default_backend ().get_cogl_context ();
+#endif
+    }
+
+    public override void detached (Clutter.Actor actor) {
+        cogl_context = null;
+    }
+
     private void emit_draw () requires (width > 0 && height > 0) {
```
tintou, line 33:
```suggestion
    private void emit_draw () requires (width > 0 && height > 0 && cogl_context != null) {
```

tintou (approved):
Looks good!

src/InternalUtils.vala
```diff
@@ -369,5 +369,9 @@ namespace Gala {
                 }
             });
         }
+
+        public static void bell_notify (Meta.Display display) {
+            display.get_sound_player ().play_from_theme ("bell-window-system", _("Bell Event"), null);
+        }
```
lenemter, line 375: This doesn't respect the flash screen option from sound settings, does it?

leolost2605: Should be fixed now. Not sure whether there's a better way to use the built in bell?

lenemter (approved):

PR #2225, opened 2025-01-12
Introduce a DesktopWorkspaceSwitcher

leolost2605:
Takes over the responsibility for animating the workspace switch from the window manager. It also prepares for switching over multiple workspaces at once with one gesture.


leolost2605:
This is ready for testing and review now. 
However it requires changes from #2195 and #2161 to fix some rare cases of unwanted behavior (e.g. the dock will always show up during switching, and one can stack the nudge infinitely)

PR #2229, opened 2025-01-15
Gestures: Replace passing with_gesture with GestureTracker.recognizing

leolost2605:
Currently we pass a `bool with_gesture` everywhere we use gesture animations for both one to one gestures and for other animations. However with some adjustment to the recognizing timing we can replace this by just checking whether the gesturetracker we're using is recognizing.

This also introduces a has_started that will only turn true once on_begin was emitted to still allow for the original function of recognizing (preventing other input while the gesture is going).

This is preparation for using the shellclients panel hiding instead of the multitasking views own stuff.


leolost2605:
Closing since this isn't needed anymore with the new gesture framework

PR #2231, opened 2025-01-16
Start the port to Mutter 48

tintou:


tintou:
There are probably many things broken right now but I would prefer to iterate in the main branch to not diverge much

lib/Image.vala
```diff
@@ -0,0 +1,75 @@
...
+        Object ();
+
+        Cogl.PixelFormat pixel_format = (pixbuf.get_has_alpha () ? Cogl.PixelFormat.RGBA_8888 : Cogl.PixelFormat.RGB_888);
+        try {
+            set_data (pixbuf.get_pixels (), pixel_format, pixbuf.width, pixbuf.height, pixbuf.rowstride);
+        } catch (Error e) {}
+    }
+}
+#else
+public class Gala.Image : GLib.Object, Clutter.Content {
```
lenemter, line 19: `Clutter.Content` already has `GLib.Object` as its parent class

lenemter (approved):
LGTM

PR #2236, opened 2025-01-18
Avoid making DBusAccelerator static

lenemter:


src/DBus.vala
```diff
@@ -33,7 +33,7 @@ public class Gala.DBus {
         Bus.own_name (BusType.SESSION, "org.gnome.Shell", BusNameOwnerFlags.NONE,
             (connection) => {
                 try {
-                    connection.register_object ("/org/gnome/Shell", DBusAccelerator.init (wm.get_display ()));
+                    connection.register_object ("/org/gnome/Shell", dbus_accelerator);
```
leolost2605, line 36: Are we using DBusAccelerator anywhere in the WindowManager? Can't we just do
```suggestion
                    connection.register_object ("/org/gnome/Shell", new DBusAccelerator (wm.get_display ()));
```

leolost2605 (changes requested):

leolost2605 (approved):
LGTM

PR #2237, opened 2025-01-18
Introduce NotificationsManager

lenemter:


src/NotificationsManager.vala
```diff
@@ -0,0 +1,121 @@
...
+ *                         2025 elementary, Inc. (https://elementary.io)
+ */
+
+/**
+ * Gala.NotificationsManager can be used to send notifications to org.freedesktop.Notifications
+ */
+ public class Gala.NotificationsManager : GLib.Object {
+    [DBus (name = "org.freedesktop.Notifications")]
+    private interface DBusNotifications : GLib.Object {
+        public abstract uint32 notify (string app_name, uint32 replaces_id, string app_icon, string summary,
```
leolost2605, line 13: I think you can just mark this as async then you don't have to go via the threadpool

lenemter: This doesn't work for me..?

```
public void send (owned NotificationData notification_data) {
    uint32? replaces_id = replaces_id_table.get (notification_data.component_name);
    if (replaces_id == null) {
        replaces_id = 0;
    }

    notifications.notify.begin (
        "gala-feedback",
        replaces_id,
        notification_data.icon,
        notification_data.summary,
        notification_data.body,
        {},
        notification_data.hints,
        EXPIRE_TIMEOUT,
        (obj, res) => {
            try {
                var notification_id = ((DBusNotifications) obj).notify.end (res);
                replaces_id_table.insert (notification_data.component_name, notification_id);
            } catch (Error e) {
                warning (e.message);
            }
        }
    );
}
    ```

Comment (leolost2605) [reply src/NotificationsManager.vala:13]: What doesn't work for you?

Here is a branch that works for me: https://github.com/elementary/gala/tree/leolost/notifications-manager
Feel free to propose it, alter it, push to it, etc :)

File: src/NotificationsManager.vala [hunk 2]
```diff
@@ -0,0 +1,101 @@
...
+                }
+            }
+        );
+    }
+
+    private void on_unwatch (DBusConnection conn) {
+        notifications = null;
+    }
+
+    public async void send (owned NotificationData notification_data) {
```
leolost2605, line 74: Do you want to keep the NotificationData class (do we need it in the future?) or can we just pass the arguments directly?

lenemter: Thanks for noticing, removed!

leolost2605 (approved):
LGTM

PR #2238, opened 2025-01-18
Send notification on screenshot

lenemter:
Merge after https://github.com/elementary/gala/pull/2237


src/ScreenshotManager.vala
```diff
@@ -23,6 +23,7 @@ namespace Gala {
     [DBus (name="org.gnome.Shell.Screenshot")]
     public class ScreenshotManager : Object {
         private WindowManager wm;
+        private NotificationsManager notifications_manager;
```
leolost2605, line 26: Do we use it here anywhere?

leolost2605 (changes requested):

src/ScreenshotManager.vala
lenemter: This is required for https://github.com/elementary/gala/pull/2241. But I removed it for now

leolost2605 (approved):
Works and code LGTM. Nice work :)

PR #2239, opened 2025-01-20
Move panel hiding to the shell clients

leolost2605:
~Requires #2229~ (not anymore, we pass with_gesture for now, won't be a big change)

We rename WindowPositioner to PositionedWindow and introduce a class ShellWindow that derives from PositionedWindow. 
The ShellWindow is the new thing that's used e.g. for centered windows. 
It's responsible for showing and hiding the window with an animation that depends on the position (e.g. slide in and out for TOP and BOTTOM, opacity for CENTERED).
We need this because with the new shell group every shell window is responsible itself for when to hide and show depending on the state (e.g. open multitaskingview).  Therefore we also introduce a ShellWindow.State flags enum which will cause the window to hide if at least one flag that indicates hide is set. 
Currently only the multitasking view notifies all shellwindows (via the manager) when it's active so that they can hide accordingly but we might add more cases here. There's also a CUSTOM_HIDDEN state that can be used by consumers of the shell window to force hide it (e.g. the panel window which derives from it force hides depending on the hide mode).


leolost2605:
This is now ready and the final part for an autarkic shell group :)

src/ShellClients/PanelWindow.vala
```diff
@@ -30,22 +52,34 @@ public class Gala.PanelWindow : Object {
...
+        unowned var workspace_manager = window.display.get_workspace_manager ();
         workspace_manager.workspace_added.connect (update_strut);
         workspace_manager.workspace_removed.connect (update_strut);
 
         window.size_changed.connect (update_strut);
         window.position_changed.connect (update_strut);
+
+        default_gesture_tracker = new GestureTracker (ANIMATION_DURATION, ANIMATION_DURATION);
+
+        notify["hidden"].connect (() => {
```
lenemter, line 66: what is `hidden`? I can't find it anywhere in the code

leolost2605: Ah right that's a copy paste remnant (the notify issue :sweat_smile:)

While looking whether we still need that I came across an issue in the hidetracker where if hide is triggered again while one was already scheduled the old hide id would be overwritten which in turn caused for it to not be reset by a call to show while it was still queued. It's a bit tricky to reproduce but can be seen if you have the dock set to autohide on overlap and move a window over it, move it while it overlaps the dock and then move it away from the dock all in the time before it hides. This would cause the dock to stay hidden even though it should show.
It happens in main as well but it's fixed now here :)

lenemter (changes requested):
There are a couple of bugs:

1. Open 2 workspaces: 1st with a window and the send without window. Using touchpad switch from 2nd to 1st but cancel it and immediately after that open multitasking view. If you timing it right you can get a dock partially (or fully) hidden when it should be visible.

2. After reproducing the first bug I can't to cancel desktop switch animation? Not sure what's going on but I get window clones from the 1st workspace on the 2nd one and I can interact with them..?

leolost2605:
> Open 2 workspaces: 1st with a window and the send without window. Using touchpad switch from 2nd to 1st but cancel it and immediately after that open multitasking view. If you timing it right you can get a dock partially (or fully) hidden when it should be visible.


I've been trying for 5 minutes now but can't reproduce it :(
Could you maybe provide a screen recording or similar?

lenemter:
I cannot reproduce this issue anymore because of the second issue :(

leolost2605:
Ok we don't ignore incoming state changes anymore while a gesture is ongoing so this should hopefully fix the issue you observed :)

lenemter:
https://github.com/user-attachments/assets/2cb4691e-4359-4210-969a-5c1906dad2f5

Not fixed :(

leolost2605:
Ah ok my guess what's happening is that it's the same symptoms we observed in #2161. Could it be that the timing is so that you start the open multitasking gesture while the hide animation of the dock is playing? For example if I increase the animation duration in panel window I can reproduce this much more easily (e.g. also by triggering a show via the barrier and then while it's animating opening the multitasking view).
To fix this we require the initial progress changes from #2161 so I guess I'll go fix up the math there (the only reason it's still drafting) which is long overdue anyways :)

leolost2605:
@lenemter ok this should be fixed now and I can't reproduce it anymore even with slower animation settings :)

lenemter:
Not sure if it is in scope of this PR but when you switch workspace there's a delay in hiding the dock, and then when you open the multitasking view before thee dock was hidden the dock doesn't hide when it should.


https://github.com/user-attachments/assets/5796736e-1b0e-49ff-bad4-a809659594b9

---

Also SafeWindowClone class is unused now, remove it please :)

leolost2605:
> Not sure if it is in scope of this PR but when you switch workspace there's a delay in hiding the dock, and then when you open the multitasking view before thee dock was hidden the dock doesn't hide when it should.

I'd say it's not in the scope of this PR since the underlying issue here is that due to the hide delay we only realize we should be hidden after a while and we can't do anything about it in the shell window because we're already in the one to one animation with the open multitasking gesture and if we would change direction or something there that would be really weird. It's pretty easy to solve in the hidetracker though which I would like to do in a follow up PR since it's not really related to the changes here :)

lenemter (approved):

PR #2241, opened 2025-01-24
Add button to the screenshot notification

lenemter:


danirabbit:
Why are we manually creating a dbus for sending notifications here instead of using https://valadoc.org/libportal/Xdp.Portal.add_notification.html ?

lenemter:
@danirabbit because it was written with fdo notifications a long time ago and I haven't looked into using portal yet. Rewriting to portals is trivial in this case because there are drop-in replacement for what the fdo dbus does so I'd like to do this later

leolost2605 (approved):

Issue #2242, opened 2025-01-24
Gala crashes when opening specific files in flatpak Inkscape 1.4


wpkelso:
### What Happened?

When opening a few specific `svg` files in flatpak Inkscape version 1.4, Gala crashes

### Steps to Reproduce

Using flatpak Inkscape V1.4 commit b329a7c4f5bd3d0ff6d9652901083f01ec0fbf75e4a4c7f92921aa5602bbd96e
- Attempt to open `icons:main /mimes/48/application-vnd.ms-access.svg` or `icons:main /mimes/48/text-x-changelog.svg` or `icons:main /mimes/48/text-x-copying.svg` (these are the only ones I've encountered so far) either from files, from within Inkscape's start dialog, or when using open from a terminal window

### Expected Behavior

File opens as normal or throws an error that doesn't crash the entirety of Gala

### OS Version

8.x (Circe)

### Session Type

Secure Session (Wayland)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell
PID: 3876 (gala)
           UID: 1000 (wkelso)
           GID: 1000 (wkelso)
        Signal: 11 (SEGV)
     Timestamp: Fri 2025-01-24 15:32:26 EST (6min ago)
  Command Line: /usr/bin/gala
    Executable: /usr/bin/gala
 Control Group: /user.slice/user-1000.slice/user@1000.service/session.slice/_io.elementary.gala@wayland.service
          Unit: user@1000.service
     User Unit: io.elementary.gala@wayland.service
         Slice: user-1000.slice
     Owner UID: 1000 (wkelso)
       Boot ID: a78abff95c2a43d1959c441cb53d2756
    Machine ID: d2cc6303c73549bea92f9a5a67468ac5
      Hostname: wonju
       Storage: /var/lib/systemd/coredump/core.gala.1000.a78abff95c2a43d1959c441cb53d2756.3876.1737750746000000.zst (present)
  Size on Disk: 15.4M
       Message: Process 3876 (gala) of user 1000 dumped core.
                
                Module libgcc_s.so.1 from deb gcc-14-14.2.0-4ubuntu2~24.04.amd64
                Module libstdc++.so.6 from deb gcc-14-14.2.0-4ubuntu2~24.04.amd64
                Module libzstd.so.1 from deb libzstd-1.5.5+dfsg2-2build1.1.amd64
                Module libudev.so.1 from deb systemd-255.4-1ubuntu8.4.amd64
[...]

Issue #2232, opened 2025-01-16
Gala crashes when leaving a workspace empty.


vvvvvvitor:
### What Happened?

Moving a window to a new workspace may sometimes crash Gala, causing the session to restart.

### Steps to Reproduce

1. Open any window.
2. Move the window to a new workspace, leaving the current workspace empty.
3. Bug happens.

### Expected Behavior

Not crashing.

### OS Version

8.x (Circe)

### Session Type

Secure Session (Wayland)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell

```

### Hardware Info

CPU: Dual Core Intel® Core(TM) i3-1005G1 CPU @ 1.20GHz
GPU: Intel® UHD Graphics
Ram: 4 GB
Mem: 256 GB

leolost2605:
What's your output of `gala --version`?

TomiOhl:
<details>
  <summary>Log output:</summary>

```sh
~$ coredumpctl debug gala
           PID: 1853 (gala)
           UID: 1000 (tomiohl)
           GID: 1000 (tomiohl)
        Signal: 11 (SEGV)
     Timestamp: Wed 2025-02-26 14:06:25 CET (25s ago)
  Command Line: /usr/bin/gala
    Executable: /usr/bin/gala
 Control Group: /user.slice/user-1000.slice/user@1000.service/session.slice/_io.elementary.gala@wayland.service
          Unit: user@1000.service
     User Unit: io.elementary.gala@wayland.service
         Slice: user-1000.slice
     Owner UID: 1000 (tomiohl)
       Boot ID: 0dbe03fdbcde4a7585ad8cb7ff853174
    Machine ID: 7d97e773b25f1c6c771c218867546e27
      Hostname: ThinkPad-T16
       Storage: /var/lib/systemd/coredump/core.gala.1000.0dbe03fdbcde4a7585ad8cb7ff853174.1853.1740575185000000.zst (present)
  Size on Disk: 6.5M
       Message: Process 1853 (gala) of user 1000 dumped core.
                
                Module libgcc_s.so.1 from deb gcc-14-14.2.0-4ubuntu2~24.04.amd64
                Module libstdc++.so.6 from deb gcc-14-14.2.0-4ubuntu2~24.04.amd64
                Module libzstd.so.1 from deb libzstd-1.5.5+dfsg2-2build1.1.amd64
                Module libudev.so.1 from deb systemd-255.4-1ubuntu8.5.amd64
[...]

Issue #2227, opened 2025-01-14
Sudden crash after waking up from sleep


MadEvening1:
### What Happened?

My laptop, with nothing open (as far as I remember), went to sleep as normal. When I woke it up from sleep, I suddenly got this screen:

![IMG_0891](https://github.com/user-attachments/assets/46f9f54f-d3b7-4af4-b156-ecb5d6ab9e53)

### Steps to Reproduce

1. Put laptop into sleep mode
2. Wake up from sleep mode
3. Look at screen

### Expected Behavior

elementary OS should have displayed the login screen.

### OS Version

8.x (Circe)

### Session Type

Classic Session (X11, This is the default)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell
PID: 1341 (gala)
           UID: 1000 (mad)
           GID: 1000 (mad)
        Signal: 11 (SEGV)
     Timestamp: Tue 2025-01-14 21:04:35 CET (15min ago)
  Command Line: /usr/bin/gala
    Executable: /usr/bin/gala
 Control Group: /user.slice/user-1000.slice/user@1000.service/session.slice/_io.elementary.gala@x11.service
          Unit: user@1000.service
     User Unit: io.elementary.gala@x11.service
         Slice: user-1000.slice
     Owner UID: 1000 (mad)
       Boot ID: 35b5f508666446b6a043d80a3fb65016
    Machine ID: a68e4fa6c52c3c301368df5567841a99
      Hostname: mad-latitude-3340
       Storage: /var/lib/systemd/coredump/core.gala.1000.35b5f508666446b6a043d80a3fb65016.1341.1736885075000000.zst (present)
  Size on Disk: 3.8M
       Message: Process 1341 (gala) of user 1000 dumped core.
                
                Module libgcc_s.so.1 from deb gcc-14-14.2.0-4ubuntu2~24.04.amd64
                Module libstdc++.so.6 from deb gcc-14-14.2.0-4ubuntu2~24.04.amd64
                Module libzstd.so.1 from deb libzstd-1.5.5+dfsg2-2build1.1.amd64
                Module libudev.so.1 from deb systemd-255.4-1ubuntu8.4.amd64
                Module libsystemd.so.0 from deb systemd-255.4-1ubuntu8.4.amd64
                Stack trace of thread 1341:
                #0  0x0000760b1a862cd8 g_type_check_instance_is_a (libgobject-2.0.so.0 + 0x3ccd8)
[...]

PR #2243, opened 2025-01-25
WorkspaceManager: avoid using Meta.WorkspaceManager.list_workspaces

lenemter:
I'm not sure if this is a true cause of the crashes but https://github.com/elementary/gala/issues/2242, https://github.com/elementary/gala/issues/2227, https://github.com/elementary/gala/issues/2232 indicate that the issue is somewhere here


leolost2605 (changes requested):
I've added some comments adjusting for the fact that we remove workspaces in the for loop which might lead to some out of bounds/skipping workspaces issues.

PR #2247, opened 2025-01-25
IconGroup: Inline namespace

leolost2605:


lenemter (commented):
I guess we should update copyright year here?

lenemter (approved):

PR #2248, opened 2025-01-25
IconGroupContainer: Inline namespace

leolost2605:


src/Widgets/IconGroupContainer.vala
```diff
@@ -1,175 +1,162 @@
...
-                return _scale_factor;
-            }
-            set {
-                if (value != _scale_factor) {
-                    _scale_factor = value;
-                    reallocate ();
-                }
+/*
+ * Copyright 2014 Tom Beckmann
+ * Copyright 2025 elementary, Inc. <https://elementary.io>
```
lenemter, line 3: You should use SPDX-FileCopyrightText: here

leolost2605: Updated it :)

lenemter (approved):

Issue #2126, opened 2024-11-26
Applications taking over


doobes:
### What Happened?

Was watching a youtube video embedded in a Reddit post.

Hit the expand to full screen selection

An image at that precise instant was presented, with the expanded youtube window behind it. I could not close or eliminate the image. All the other windows were displayed behind the image.

![Screenshot from 2024-11-26 10 05 15](https://github.com/user-attachments/assets/775621a2-7b2f-4e31-b4df-87b8287766d6)

The only recourse was to reboot EOS.

I had the same thing happen with XNView MP yesterday.

### Steps to Reproduce

Open Brave Browser

Navigate to a page with an embedded YouTube video

Hit the Expand to full screen selection

### Expected Behavior

Not to have a static image of a window displayed in front of everything else.

### OS Version

8.x (Early Access)

### Session Type

Classic Session (X11, This is the default)

### Software Version

Latest release (I have run all updates)

### Hardware Info

Dell XPS 13 9333

PR #2255, opened 2025-01-26
WindowManager: cleanup maximize and unmaximize animations

lenemter:
Review with hide whitespace

Fixes https://github.com/elementary/gala/issues/2126

Changes:
- Fix wrong snapshot animation in maximize
- Make sure to respect 'move maximized/fullscreen windows to new ws' even when animations are disabled
- Exit from methods early
- Unparent window snapshots instead of destroying them


leolost2605 (changes requested):
Looks good but when animations are disabled and i maximize/fullscreen a window it doesn't get moved to a new workspace

src/WindowManager.vala
```diff
@@ -1317,91 +1312,96 @@ namespace Gala {
...
-
-            var duration = AnimationDuration.SNAP;
-
-            kill_window_effects (actor);
+            warning ("Maximize");
 
             unowned var window = actor.get_meta_window ();
             if (window.maximized_horizontally && behavior_settings.get_boolean ("move-maximized-workspace")
                 || window.fullscreen && behavior_settings.get_boolean ("move-fullscreened-workspace")) {
+                warning ("Moving to next ws");
```
leolost2605, line 1320:
```suggestion
```

src/WindowManager.vala
```diff
@@ -1317,91 +1312,96 @@ namespace Gala {
...
 
         private void maximize (Meta.WindowActor actor, int ex, int ey, int ew, int eh) {
-            if (!AnimationsSettings.get_enable_animations ()) {
-                return;
-            }
-
-            var duration = AnimationDuration.SNAP;
-
-            kill_window_effects (actor);
+            warning ("Maximize");
```
leolost2605, line 1315:
```suggestion
```

leolost2605 (approved):
Some debugs to remove but apart from that LGTM :)

src/WindowManager.vala
lenemter: Oops

PR #2256, opened 2025-01-26
Split Dialog.vala into multiple files

lenemter:
Split Dialog.vala into multiple files according to https://docs.elementary.io/develop/writing-apps/code-style#classes-and-files


danirabbit (commented):
Don't forget to update POTFILES :)

PR #2265, opened 2025-02-12
Dock workspace switcher

lenemter:


teamcons:
wingpanel does not reappear when leaving multitasking view
it reappears a bit later upon switching workspace (regardless how, except multitasking view)

lenemter:
@teamcons Please file a new issue for that, this branch doesn't touch the positioning code at all

Issue #2268, opened 2025-02-14
Notification on a new workspace creates duplicate new workspace


lenemter:
### What Happened?

Reported here: https://github.com/elementary/dock/pull/361#issuecomment-2657783172

### Steps to Reproduce

1. Be on an empty workspace
2. Get a notification
3. See the empty workspace pop up in the dock (same thing in multitasking view)

### Expected Behavior

No new workspace

### OS Version

8.x (Circe)

### Session Type

Secure Session (Wayland)

### Software Version

Compiled from git

### Log Output

```shell

```

PR #2269, opened 2025-02-14
Workaround notifications bug

lenemter:
Fixes https://github.com/elementary/gala/issues/2268


leolost2605 (changes requested):
This breaks new workspaces creation when moving a window to a new workspace:

Screenshot from 2025-02-15 15 01 43@1x

And later crashed when opening a new maximized window that should get pushed to a new workspace. 

I've proposed a new PR that should fix the mentioned issue with (hopefully) a little less breakage: #2271

PR #2273, opened 2025-02-16
DwellClickTimer: Remove delay

lenemter:


lenemter:
@leolost2605 I removed delay because it is not needed anymore. I have no idea where it came from since mutter is responsible for dwell click, not libinput. Also gnome-shell doesn't add any delay either, so I don't think UX input is needed here

leolost2605 (approved):
@lenemter ok yeah I just wasn't sure because I don't use this feature but the change was quite noticeably so I thought there must've been a reason for it.
Anyways the changes LGTM :)

davidmhewitt:
https://github.com/elementary/gala/issues/1152

lenemter:
@davidmhewitt Thanks for linking that, I totally missed this issue

PR #2275, opened 2025-02-16
 Implement support for keeping shellclients in the multitasking view 

### LINKED ISSUE #2071: Dock overlapping right-click menu on elementary OS 8 (next)
### What Happened?

when you right-click on a file, the menu appears behind the Dock.

### Steps to Reproduce

1. open Files and position it next to the Dock
2. right-click on a file 
3. the menu will appear behind the Dock.

Captura de tela de 2024-09-23 20-45-50

### Expected Behavior

When you right-click on a file, the Dock shouldn't overlap the menu.

### OS Version

8.x next

### Session Type

Secure Session (Wayland)

### Software Version

### Log Output

_No response_

### Hardware Info

elementary OS 8 next
kernel 6.8.0-45

leolost2605:
This provides support for keeping a shellclient (the dock) visible in the multitasking view.

Everything one expects currently works. You can launch new apps, switch workspace by clicking on an app on another workspace etc..

Support was added to allow dragging windows to icon groups in the dock.

Goes with elementary/dock#375

Things todo/think about:
- When multiple instances of an app are open the window overview is shown. That's not really nice as it is right now however I wonder what we should do instead? One option that I came up with: #2350 

Fixes #2071 (at least most of the time, sometimes the user time is a bit outdated but better than main ig)


leolost2605:
This should be ready for an initial review now together with elementary/dock#375
Lots of testing and feedback would be appreciated :)
Especially testing with different monitor scales and wild monitor transforms like rotations, flips, etc. etc. would be useful since we have to do quite a bunch of coordinates conversion when dropping winows on the dock icon groups

lenemter (approved):
LGTM! There are two issues:
 1. Missing X11 support
 2. Window clones have unfocused styles in multitasking view
 
 But we can deal with them later

leolost2605:
@lenemter I added support for X here, the other thing I'd like to do in a follow up :)

src/PantheonShell.vala
```diff
@@ -292,6 +293,23 @@ namespace Gala {
...
+        unowned PanelSurface? panel_surface = resource.get_user_data<PanelSurface> ();
+        if (panel_surface.wayland_surface == null) {
+            warning ("Window tried to set visible in mutltiasking view but wayland surface is null.");
+            return;
+        }
+
+        Meta.Window? window;
+        panel_surface.wayland_surface.get ("window", out window, null);
+        if (window == null) {
+            warning ("Window tried to set hide mode but wayland surface had no associated window.");
```
lenemter, line 306:
```suggestion
            warning ("Window tried to request visible in multitasking view but wayland surface had no associated window.");
```

src/Widgets/MultitaskingView/WindowClone.vala
```diff
@@ -512,59 +513,18 @@ public class Gala.WindowClone : ActorTarget, RootTarget {
...
+        if (destination is Meta.WindowActor) {
             if (hovered) {
-                icon_group.add_window (window, false, true);
+                WindowDragProvider.get_instance ().notify_enter (window.get_id ());
             } else {
-                icon_group.remove_window (window, false);
+                WindowDragProvider.get_instance ().notify_leave ();
             }
         }
+    }
```
lenemter, line 524: Can we reverse the if statement here to avoid nesting?

```
private void destination_crossed (Clutter.Actor destination, bool hovered) {
    if (!(destination is Meta.WindowActor)) {
        return;
    }

    if (hovered) {
        WindowDragProvider.get_instance ().notify_enter (window.get_id ());
    } else {
        WindowDragProvider.get_instance ().notify_leave ();
    }
}
```

PR #2258, opened 2025-01-28
Introduce GestureController

leolost2605:
This is ready for review now. It's a pretty big PR which is unfortunately necessary because it changes how the whole gesture propagation works. For an easier review you can go with the separate commits. Use rebase to merge.

An overview of the changes:
We go from a gesturetracker that only know about the current gesture and not "where we currently are", and which sends signals to gesturepropertytransitions which are created new for every gesture, to a gesturecontroller that keeps a persistent double progress which represents the state of the UI, that is affected by the gesture action that this gesturecontroller has. E.g. 0 for multitaskingview closed, 1 for it opened. 0 for first workspace, -1 for second, -2 for third etc. It hereby snaps to full integer values and can be bounded by upper and lower limits.
We also go from sending signals to propagating gesture events via a hierarchy of GestureTargets with the root being given to the controller. 
Currently there are PropertyTargets that manipulate a property of a clutter actor, and ActorTargets that propagate the gesture events to direct descendants that are also actortargets. ActorTargets can also take other gesturetargets (e.g. PropertyTargets) and automatically propagate gesture events to them. 
Since gesture targets can receive events from multiple controllers, events are connected to a string id (this will be required for #1733).
The advantages from this PR:
- We have proper deceleration when a gesture ends (like in adwaita)
- We got a big negative diff in the widgets (in theory also in general but not right now because gesturetracker and gesturepropertytransition can only be removed once everything uses gesturecontroller (will be pretty straight forward to do that but this PR is big enough as is)) 
- We support interrupting gestures for opening the multitasking view
- The widgets have already support for #1733 which is plug and play once a backend supports it. This support can already be seen when you switch workspaces in the multitaskingview via gestures pause somewhere and then use keybinds to close the multitaskingview)
- The system is much more flexible and future proof (e.g. use the distance for finger tracking touchscreen gestures)
- it already prepares for a deduplicated version of #2225 
- The cancel behavior is much more predictable now. We round to the integer and if we have enough speed we go to the next one in the direction we were going.
 
Follow up:
- [ ] workspace switch doesn't actually use end velocity, fix it like in #2225 (active index)

Fixes #2279

PR #2281, opened 2025-02-24
WorkspaceManager: Fix crash on workspace removed

leolost2605:
Tries to do the same as #2243.
It also fixes some warnings `g_object_notify: assertion 'G_IS_OBJECT (object)' failed`.

The reason from the code comment:

> We might be here because of a signal emition from the ws machinery (e.g. workspace.window_removed).
> Often the function emitting the signal doesn't take a ref on the ws so if we remove it right
> away it will be freed. But because the function often accesses it after the singal emition this leads
> to warnings and in some cases a crash.

Might fix https://github.com/elementary/gala/issues/2242, https://github.com/elementary/gala/issues/2227, https://github.com/elementary/gala/issues/2232.
Fixes the crash described in #2258


lenemter (approved):
:+1:

PR #2284, opened 2025-02-28
ToucheggBackend: Use Priority.DEFAULT for sending events to the main thread

leolost2605:
Reason behind this (from #2258):

Ok I did some more investigation. Turns out that the reason is that in our toucheggbackend we receive events in a different thread and send them via idle to the main thread. So far so good. However we use Priority.DEFAULT_IDLE (we don't set anything) which might cause them to be stuck behind other more important events. This is the case when we do update_hover_widgets in the window clone because there they get animated via set_easing etc. which seemingly blocks lower priority events. This sometimes causes a delay which seems like a lag.


lenemter (approved):
Fixes the issue in https://github.com/elementary/gala/pull/2258. Good job!

PR #2285, opened 2025-03-01
DesktopIntegration: add 'time-appeared-on-workspace' property

lenemter:


src/DesktopIntegration.vala
```diff
@@ -43,6 +48,17 @@ public class Gala.DesktopIntegration : GLib.Object {
         // TODO: figure out if there's a better way to handle ws rearrangement
         display.window_created.connect ((window) => {
             window.workspace_changed.connect (() => windows_changed ());
+
+            time_appeared_on_workspace[window] = GLib.get_real_time ();
```
leolost2605, line 52: I think we'd rather want to use [get_monotonic_time](https://valadoc.org/glib-2.0/GLib.get_monotonic_time.html) here?

leolost2605 (changes requested):

src/DesktopIntegration.vala
```diff
@@ -43,6 +48,17 @@ public class Gala.DesktopIntegration : GLib.Object {
         // TODO: figure out if there's a better way to handle ws rearrangement
         display.window_created.connect ((window) => {
             window.workspace_changed.connect (() => windows_changed ());
```
leolost2605, line 50: One more thing, this seems to be duplicate with the `workspace_changed.connect` where we also set the time_appeared_on_workspace. Or is there a reason for that?

leolost2605 (changes requested):

src/DesktopIntegration.vala
lenemter: Fixed. Forgot to remove that

leolost2605 (approved):

leolost2605:
Is there anything still missing here or can this be merged?

lenemter:
@leolost2605 This is ready to merge. I'm just waiting for approval of https://github.com/elementary/dock/pull/388

PR #2286, opened 2025-03-01
WorkspaceManager: Fix dynamic workspaces

leolost2605:
This should hopefully work now :crossed_fingers:

Fixes elementary/dock#389

### DIFF
--- a/src/WorkspaceManager.vala
+++ b/src/WorkspaceManager.vala
@@ -179,10 +179,7 @@ public class Gala.WorkspaceManager : Object {
             && workspace != last_workspace
         ) {
             queue_remove_workspace (workspace);
-        }
-
-        // if window is the second last and empty, make it the last workspace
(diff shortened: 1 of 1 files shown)


lenemter (changes requested):
A bug I noticed:

Open 2 workspaces with one window each, start moving window from 1st ws to the 2nd. The 1st workspace is removed before you complete the gesture, so you can't cancel it

leolost2605:
I can reproduce this in main as well (even before the changes to removal/adding).
The proper way to fix that is actually switching the workspace only once the gesture finished and not when it starts which will be done in a deduplicated version of https://github.com/elementary/gala/pull/2225 based on #2258. 
I think I'd prefer to do it that way instead of introducing some weird workaround/hacks here :)

lenemter (approved):

Issue #321, opened 2018-08-13
Use org.gnome.desktop.interface enable-animations


cassidyjames:
The GNOME key's description is:

>Whether animations should be displayed. Note: This is a global key, it changes the behaviour of the window manager, the panel etc.

Since this affects GTK, it would simplify getting and setting this value for a11y if we just used and respected the GNOME setting in Gala as well.

danrabbit:
The problem with this is that disabling Gtk animations gives weird results with revealers so I don’t think we should actually expose that setting. Having Gala animations be indepdently controlled is an advantage for us

cassidyjames:
@danrabbit is that still the case with revealers? I thought that was fixed in 3.22 or something. If there weren't weird bugs with GTK, wouldn't we want one unified setting to reduce motion/animations, though?

lenemter:
@danirabbit Is this still something we want to do?

danirabbit:
@lenemter sorry for long follow up. I think we need to see if the aforementioned GTK issue has been resolved first. The problem was that when animations were disabled revealers would always preserve their space, so you'd have big blank spaces randomly in the UI for example in the panel where the battery percentage would be.

If this is fixed in Gtk then yeah that would make sense to me to follow that key and it would be nice to have animations disabled in more places for this setting

danirabbit:
Looks like that GTK bug is fixed in OS 8, so I'm +1 on switching over to the GNOME interface key. Proposed https://github.com/elementary/settings-desktop/pull/438 to go with it

PR #2288, opened 2025-03-02
Use org.gnome.desktop.interface enable-animations

lenemter:
Fixes https://github.com/elementary/gala/issues/321


leolost2605 (approved):
Changes look good to me :)

I wonder whether we should add a little postinstall to sync the keys somewhere? Maybe the settings repo if it wasn't released yet?

Issue #1998, opened 2024-07-25
Blur Behind


danirabbit:
### Problem

Shell elements are transparent or semi-transparent so that they blend into the desktop itself, but this sometimes creates issues with contrast: https://github.com/elementary/dock/issues/259

### Proposal

A way to blur behind the Dock and Panel (when it has a background)

### Prior Art (Optional)

Windows:
![b8AefxSmxXELW9BKkNo6vW-3702192291](https://github.com/user-attachments/assets/f3686815-8293-4e50-a1b0-ee5426016d0d)

macOS:
![macos-sequoia-recopie-iphone-3938195366](https://github.com/user-attachments/assets/d9aaf036-b854-45bf-b350-42674dc50616)

Chrome OS:
![MeetChromeOS_Intro](https://github.com/user-attachments/assets/5e1675eb-05c8-44b5-b0d5-72d45908e0c5)

seifeldinio:
I propose adding a light background color with 10% opacity and a backdrop filter with a 9px blur effect. This enhances visibility and maintains the aesthetic blending of shell elements with the desktop.

```css
dock {
    background: rgba(255, 255, 255, 0.1); /* 👈 Light background with 10% opacity */
    border-radius: 9px;
    box-shadow:
        inset 0 -1px 0 0 rgba(255, 255, 255, 0.1),
        inset 0 1px 0 0 rgba(255, 255, 255, 0.15),
        inset 1px 0 0 0 rgba(255, 255, 255, 0.035),
        inset -1px 0 0 0 rgba(255, 255, 255, 0.035),
        0 0 0 1px rgba(0, 0, 0, 0.4),
        0 1px 3px rgba(0, 0, 0, 0.1),
        0 3px 9px rgba(0, 0, 0, 0.15);
    margin: 9px;
    margin-top: 0;
    backdrop-filter: blur(9px); /* 👈 Blur effect */
}
```

### Mockups

To demonstrate the effectiveness of this solution, I have created four mockups showcasing the Dock over different backgrounds and text:

1. **Text Background:**

<img width="1062" alt="Above text" src="https://github.com/user-attachments/assets/af6e7230-7093-42b5-8fbd-9142d0f1ffd6">


2. **Light Background:**

<img width="1062" alt="Above light" src="https://github.com/user-attachments/assets/bd6318d7-ebcf-405b-be10-c3f3f89da863">


3. **Mid Background:**

[...]

Issue #2216, opened 2025-01-07
Use pre configure mechanism for window positioning


leolost2605:
https://mutter.gnome.org/meta/signal.Window.configure.html
https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/4076

This can be used in the windowpositioner: we only change the values of the given MetaWindowConfig and it will be applied.

Unfortunately this is OS9 material because it requires a more recent mutter version

PR #2293, opened 2025-03-05
Add Blur API

lenemter:
Screenshot from 2025-03-05 20 48 42

Fixes https://github.com/elementary/gala/issues/1998

To do:
- [x] ~~Figure out how to handle power saving (probably Granite and applications should handle that??)~~
- [x] ~~Test with multiple monitors~~ (doesn't make sense yet)
- [x] Test with display scaling
- [x] ~~Fix X11 bug when window opens~~ (probably will be fixed by https://github.com/elementary/gala/issues/2216)
- [x] Test with fractional scaling

### DIFF
--- a/protocol/pantheon-desktop-shell-v1.xml
+++ b/protocol/pantheon-desktop-shell-v1.xml
@@ -104,6 +104,24 @@
         Tell the shell that the panel would like to be visible in the multitasking view.
       </description>
     </request>
+
+    <request name="add_blur">
+      <description summary="add blur">
+        Tell the window manager to add background blur.
+      </description>
+
+      <arg name="left" type="uint"/>
+      <arg name="right" type="uint"/>
+      <arg name="top" type="uint"/>
+      <arg name="bottom" type="uint"/>
+      <arg name="clip_radius" type="uint"/>
+    </request>
+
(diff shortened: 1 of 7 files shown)


danirabbit:
Screenshot from 2025-07-28 13 05 48

I'm sure @leolost2605 has some more informed opinions about the UI, but from a design perspective I think it looks clean and is very legible :)

leolost2605 (commented):
This looks very good thank you!
I wonder though if we still want the left, right, top, bottom offsets or if we'd rather want to set the correct shadow size in gtk like it's done in the dock since elementary/dock#437 ?

lenemter:
@leolost2605 I initially tried this approach, but the problem is dock's bottom margin. If we set it as 'shadow' then there will be no margin. But if we set it as window content then the bottom margin will be blurred.

danirabbit:
@lenemter are you happy to merge this?

lenemter:
@danirabbit I'm happy with the gala part but not so much with the dock. We basically need to either hardcode bottom margin size or allow changing it via gsetting. There's probably a way to achieve this with clever CSS but I'm not ready to dig into it yet.

PR #2298, opened 2025-03-09
Cleanup ScreenshotManager and WindowManager

lenemter:
Review with 'Hide Whitespace'

Changes:
- Inline `ScreenshotManager`'s namespace
- Move handling screenshot keyboard shortcuts into ScreenshotManager
- Move `wm.launch_action` and actions it accepts into WindowManager interface


lib/WindowManager.vala
```diff
@@ -52,6 +52,14 @@ namespace Gala {
         IS_TILED
     }
 
+    [CCode (has_type_id = false)]
+    public class ActionKeys {
```
leolost2605, line 56: Maybe this should be a namespace instead?

leolost2605 (changes requested):
LGTM (apart from one minor comment) though I'm not a huge fan of having all of this in a single commit but it's fine ig.

leolost2605 (approved):
LGTM

PR #2299, opened 2025-03-09
Add 'Open in Photos' to screenshot notification

lenemter:
image

Requires https://github.com/elementary/gala/pull/2298


lenemter:
@danirabbit Fixed! I confirmed it uses translated app name as it should. But I encountered https://github.com/elementary/notifications/issues/237 which makes it look very bad with Russian where "Open in Photo Viewer" is a very long line (33 symbols)

PR #2301, opened 2025-03-09
Introduce an action filter

### LINKED ISSUE #2295: Window overview is broken
### Steps to Reproduce

1. Open window overview
2. Switch workspaces

### Expected Behavior

We probably shouldn't allow switching workspaces in window overview

### OS Version

8.x (Circe)

### Session Type

Secure Session (Wayland)

### Software Version

Compiled from git

### Log Output

```shell

```

### Hardware Info

_No response_

leolost2605:
Introduce an action filter for when the wm is modal that automatically filters corresponding keybinds but also gestures.

Fixes #2295


lib/WindowManager.vala
```diff
@@ -185,5 +196,11 @@ namespace Gala {
          * @param action_key The gsettings key of action. Available keys are stored in ActionKeys
          */
         public abstract void launch_action (string action_key);
+
+        /**
+         * Checks whether the action should currently be prohibited.
+         * @return true if the action should be prohibited, false otherwise
+         */
+        public abstract bool action_filter (GestureAction action);
```
lenemter, line 204: I think it makes more sense to name is `filter_action`? action_filter is a noun but method names should be verbs

src/Gestures/GestureController.vala
```diff
@@ -32,6 +32,8 @@ public class Gala.GestureController : Object {
      */
     private const double MAX_VELOCITY = 0.01;
 
+    public WindowManager wm { get; construct; }
+
     public GestureAction action { get; construct; }
```
lenemter, line 37: Both wm and action are construct properties. The blank line is not needed here

lib/WindowManager.vala
leolost2605: I wanted to stay in line with Metas `keybinding_filter` here :shrug:

leolost2605: Anyways, I changed it to filter_action :)

lenemter (approved):
Code looks good and fixes the linked issue

PR #2303, opened 2025-03-12
Simplify multitasking view type

lenemter:
- Don't expose multitasking view to the library as no one uses it
- Don't support overriding multitasking view via plugin because we rely on it for desktop switch now
- 'Workspace view' -> 'Multitasking view'


leolost2605 (approved):
LGTM, I assume wingpanel plugin doesn't use this anywhere?

lenemter:
@leolost2605 No, it doesn't

PR #2310, opened 2025-03-13
Shorten long keybinding lines

lenemter:
From vala-lint:
148 warnings in main -> 135 warnings


leolost2605:
I think the vapi is actually outdated here. It seems like since https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/4080 it should be more like
```	
public delegate void KeyHandlerFunc (Meta.Display display, Meta.Window? window, Clutter.Event event, Meta.KeyBinding binding);
```
and then we can also drop the additional metadata for the event type I think?
ping @tintou since he's done most of the vapi stuff :)

lenemter:
@leolost2605 Won't this KeyHandlerFunc signature only be available in mutter 48?

zeebok (approved):
Assuming the VAPI isn't out of date this seems good

leonardo-lemos:
@lenemter Can we merge this?

Issue #823, opened 2020-05-28
Check for redacted font


cassidyjames:
We hardcode "Redacted Script Regular" and "Redacted Script Light" in the app, but we don't depend on them at all. This leads to an unexpected experience on other platforms; instead, we should depend on the package that provides the redacted font.

PR #2313, opened 2025-03-15
Check for redacted font

lenemter:
Fixes https://github.com/elementary/gala/issues/823

### DIFF
--- a/src/ScreenshotManager.vala
+++ b/src/ScreenshotManager.vala
@@ -352,7 +352,7 @@ public class Gala.ScreenshotManager : Object {
 
     public async void conceal_text () throws DBusError, IOError {
         if (!is_redacted_font_available) {
-            return;
+            throw new DBusError.FAILED ("Redacted font is not installed.");
         }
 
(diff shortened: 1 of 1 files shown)


src/ScreenshotManager.vala
```diff
@@ -329,6 +351,10 @@ public class Gala.ScreenshotManager : Object {
     }
 
     public async void conceal_text () throws DBusError, IOError {
+        if (!is_redacted_font_available) {
+            return;
```
leolost2605, line 355: I think we should probably throw an error here?

leolost2605 (changes requested):
I think the bigger part here is that the UI the user uses shows that the feature is not available instead of just silently failing? So maybe we should add the same check to the screenshot app?

lenemter:
@leolost2605 Done!

https://github.com/elementary/screenshot/pull/302
https://github.com/elementary/portals/pull/130

leolost2605 (approved):
LGTM

Issue #1679, opened 2023-05-26
Reduce the number of windows_changed calls in DesktopIntegration


tintou:
We are currently triggering a `windows_changed` for every tooltip and context menu that appears on screen, we should only call it when the windows have actually changed (added/removed/renamed/focused)

lenemter:
Fixed in https://github.com/elementary/gala/pull/2314

PR #2314, opened 2025-03-15
Reduce the number of windows_changed calls

lenemter:
Addresses https://github.com/elementary/gala/issues/1679


leolost2605 (approved):
LGTM

PR #2319, opened 2025-03-16
Multitasking view fixes

### LINKED ISSUE #2297: Weird behavior when spamming Super+S and Super+1
### What Happened?

https://github.com/user-attachments/assets/af1d2d6e-55fb-4c62-8ef9-fb4094b3d878

### Steps to Reproduce

Just spam Super+1 and Super+S, and randomly window manager level keyboard shortcuts stop working (dock shortcuts work fine). And when removing the last window from a workspace, the workspace gets instantly removed.

### Expected Behavior

-

### OS Version

8.x (Circe)

### Session Type

Secure Session (Wayland)

### Software Version

Compiled from git

### Log Output

```shell

```

### Hardware Info

_No response_

leolost2605:
Two fixes for the multitasking view.

The first commit fixes #2297 where we would push modal twice leading to one of the grabs never being released.

The second commit fixes an issue where you could initiate a drag and drop on window clones while
the workspace switch animation finished which looks very strange.

### DIFF
--- a/src/Widgets/MultitaskingView/MultitaskingView.vala
+++ b/src/Widgets/MultitaskingView/MultitaskingView.vala
@@ -508,5 +508,19 @@ namespace Gala {
 
             return true;
         }
+
+        public override bool captured_event (Clutter.Event event) {
+            /* If we aren't open but receive events this means we are animating closed
+             * or we are finishing a workspace switch animation. In any case we want to
+             * prevent a drag and drop to start for the window clones which can happen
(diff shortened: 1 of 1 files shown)


lenemter (approved):
Fixes the linked issue. Thanks!

I noticed that I still can start drag operation while switching workspaces with gestures, it seems like it wasn't intended to be fixed in this branch, so I'm approving it.

PR #2320, opened 2025-03-16
Reorder workspaces

leolost2605:
First commit adds a dbus method to reorder workspaces required for elementary/dock#392

Second commit fixes a small issue where if the active workspace was reordered the multitasking view was still at the outdated active index which now is a different workspace.


lenemter (approved):
There is a visual bug when reordering workspaces for the first time but I don't really care, icon groups are super buggy overall

Issue #2199, opened 2024-12-30
Swiping between workspaces in the multitasking view can sometimes close applications underneath the cursor


wpkelso:
### What Happened?

Swiping between workspaces in multitasking view will sometimes close applications present within those workspaces if the pointer is hovered over them

### Steps to Reproduce

1. Open multitasking view with multiple workspaces active
2. Place the cursor over an application window (does not have to be over the close button
3. Do a quick horizontal 3-finger swipe, lifting off the touchpad before the animation has finished
4. Repeat 3 until an application window gets closed (seems to happen more often when quickly swiping between workspaces)

### Expected Behavior

Application windows don't close in multitasking view unless I specifically press the "x" button to close the window.

### OS Version

8.x (Circe)

### Session Type

Secure Session (Wayland)

### Software Version

Older release (I have not run all updates)

### Hardware Info

![image](https://github.com/user-attachments/assets/f02cae5b-e46c-4510-af2f-c28de6ba4785)

Synaptics touchpad (Lenovo T490)

leolost2605:
Just to add some info here: Windows are closed when middle clicked in the multitasking view (intended). Touchpads usually recognize a three finger tap as a middle click (also intended) so I'm guessing this is where this issue is coming from. This seems more like a timing thing maybe even hardware specific (here comes the famous it works for me :grimacing:) so I'm not sure we can do something about it :/
Though this might be fixed by the new touchpad backend on wayland since it will stop event propagation once a gesture is recognized.

wpkelso:
I'll check it once the new backend is released. It seems like a lot of the issues I'm encountering right now are based around touchpad swipes.

PR #2327, opened 2025-03-23
Multitasking view: swipe up to close window

lenemter:
Fixes https://github.com/elementary/gala/issues/2199

https://github.com/user-attachments/assets/8996a666-a80a-4dde-9246-df8d55198783

- [x] Figure out how not to block horizontal scroll

### DIFF
--- a/lib/Constants.vala
+++ b/lib/Constants.vala
@@ -42,6 +42,7 @@ namespace Gala {
         MULTITASKING_VIEW,
         DOCK,
         ZOOM,
+        CLOSE_WINDOW,
         N_ACTIONS
     }
 
--- a/src/Gestures/ScrollBackend.vala
+++ b/src/Gestures/ScrollBackend.vala
@@ -108,8 +108,8 @@ public class Gala.ScrollBackend : Object, GestureBackend {
         return Clutter.EVENT_PROPAGATE;
     }
 
-    private bool on_leave_event (Clutter.Event event) {
-        if (event.get_type () != LEAVE) {
+    private bool on_leave_event (Clutter.Event event) requires (event.get_type () == LEAVE) {
+        if (!started) {
             return Clutter.EVENT_PROPAGATE;
         }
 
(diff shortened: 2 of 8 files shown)


wpkelso (commented):
Thoughts on the animation being like an an old TV turning off, where the window goes flat then collapses horizontally?

danirabbit:
In iPad OS you kind of throw them off screen:


https://github.com/user-attachments/assets/e3a30ccc-6e16-43cd-bed3-c3af82a888d4

danirabbit:
Looks like Android is the same https://youtu.be/9KXsAANhh7g

but it looks like in Android you don’t need to swipe all the way up from the bottom of the screen off the top of the screen. At a certain threshold it just fades out. Which makes sense to me

lenemter:
@danirabbit does this look good to you?

https://github.com/user-attachments/assets/8996a666-a80a-4dde-9246-df8d55198783

danirabbit:
@lenemter that looks amazing 😍

wpkelso:
Does e8cc421a8b486ebd7f53f726b4b98a05705a1827 not just re-add the issue from #2199 ?

danirabbit (changes requested):
This works great with trackpad, but on touch screen it looks like a window drag operation gets started when you try to swipe the window. I guess on touch screen we should only start a drag operation from the app icon to solve this?

When you start a swipe, the close button and tooltip fade out, which makes sense, but it's really slow. I also wonder should this stick to the window like the app icon does? It looks kind of weird for the close icon to get far away from the corner of the window

lenemter:
@wpkelso I suppose we want to block middle click to clowe window on anything other than mouse and allow swiping up with gestures instead, no?

lenemter:
@danirabbit touchscreen backend is not merged yet: https://github.com/elementary/gala/pull/2096

danirabbit:
Oh my bad

lenemter:
@danirabbit Fixed!

leolost2605 (changes requested):
I can't switch workspaces with three fingers now.


https://github.com/user-attachments/assets/c40e9172-ccbe-4a16-8f6b-e49cbd8b12d3

danirabbit (approved):

leolost2605 (approved):
LGTM lets gooo

PR #2330, opened 2025-03-24
WindowSwitcher: blur background

lenemter:
New clean branch that should be easier to review

Window switcher is used as an example of how this effect can be used because Danielle showed some interest in blurring it: https://github.com/elementary/granite/pull/747#issuecomment-2704200141

Can be rebase merged

### DIFF
--- a/vapi/mutter-cogl-14.vapi
+++ b/vapi/mutter-cogl-14.vapi
@@ -514,7 +514,7 @@ namespace Cogl {
 #endif
 		public void set_uniform_1f (int uniform_location, float value);
 		public void set_uniform_1i (int uniform_location, int value);
-		public void set_uniform_float (int uniform_location, int n_components, int count, float value);
+		public void set_uniform_float (int uniform_location, int n_components, int count, [CCode (array_length = false, array_null_terminated = false)] float[] value);
 		public void set_uniform_int (int uniform_location, int n_components, int count, int value);
 		public void set_uniform_matrix (int uniform_location, int dimensions, int count, bool transpose, float value);
 		public void set_user_program (Cogl.Program program);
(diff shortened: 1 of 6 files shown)


danirabbit (approved):
Approve for UX, but still needs code review :)

vapi/mutter-cogl-14.vapi
```diff
@@ -475,7 +475,7 @@ namespace Cogl {
 #endif
 		public void set_uniform_1f (int uniform_location, float value);
 		public void set_uniform_1i (int uniform_location, int value);
-		public void set_uniform_float (int uniform_location, int n_components, int count, float value);
+		public void set_uniform_float (int uniform_location, int n_components, int count, float *value);
```
leolost2605, line 478: Is there a reason this is a pointer? According to the vapi docs if possible you should use the vala provided syntax to make the semantics clearer: https://docs.vala.dev/developer-guides/bindings/writing-a-vapi-manually/07-00-binding-a-c-function-s-parameter-and-return-types/07-07-pointers.html

lenemter: It's a pointer to the start of array of float values in C header. I don't know if there's other way to pass it

lenemter: I put it in the vapi file for now. I always assumed our vapi files were written by hand since the meson script in the vapi folder never worked for me. I guess we should ask @tintou about where should this change be

leolost2605 (approved):
Gonna be honest no clue how the blur works but structure LGTM. It has a bit of a weird frame in light mode on a light window but ig we can ignore that for now :shrug:

lenemter:
@leolost2605 I don't see anything weird here. Can you send a screenshot please?

leolost2605:
I dont know if that's intended but it looks like the background overflows the frame especially in the corners
Screenshot from 2025-05-14 12 39 22@2x

lenemter:
@leolost2605 That's a ShadowEffect issue. I'll look into fixing it later

danirabbit:
@lenemter what's blocking here? Would be great to get this in for 8.1 :)

PR #2334, opened 2025-03-25
Respect orientation in scroll backend

leolost2605:
Get rid of the weirdness in commit_progress and the required additional check in update_progress.
A scroll action will now only trigger if the scroll direction matches its orientation.
This now in theory also supports disabled animations but if multiple windows are open it will pretty much close all since they probably rearrange before you stop scrolling. Currently it's still enabled but we can disable it.
Also fixes runtime warnings about non existing objects.

Also the additional check around update_hover_widgets was removed. Idk why it was added but if we need it we should also update on switching workspaces because the close button and windowtitle appear when switching workspaces on the desktop which looks weird.

### DIFF
--- a/src/Gestures/ScrollBackend.vala
+++ b/src/Gestures/ScrollBackend.vala
@@ -85,6 +85,14 @@ public class Gala.ScrollBackend : Object, GestureBackend {
 
         if (!started) {
             if (delta_x != 0 || delta_y != 0) {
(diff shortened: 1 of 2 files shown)


lenemter (approved):
This looks great and works as expected, thanks for fixing my terrible code :)

I think we shouldn't allow closing windows this without animations because it is easy to trigger can close multiple windows at once

Issue #2308, opened 2025-03-13
Random crash when using window switcher


lenemter:
### What Happened?

Random crash when using window switcher

### Steps to Reproduce

I don't know, gala just randomly crashed

### Expected Behavior

No crash

### OS Version

8.x (Circe)

### Session Type

Secure Session (Wayland)

### Software Version

Compiled from git

### Log Output

```shell
(gdb) bt
#0  0x00007b22bdf9301c in clutter_stage_set_key_focus () at /usr/lib/x86_64-linux-gnu/mutter-14/libmutter-clutter-14.so.0
#1  0x00005b5c42403ae2 in gala_window_switcher_set_current_icon (self=0x5b5c7deb6620, value=0x5b5c7edd0680) at ../src/Widgets/WindowSwitcher/WindowSwitcher.vala:45
#2  0x00005b5c42401511 in gala_window_switcher_real_propagate (base=0x5b5c7deb6620, update_type=GALA_GESTURE_TARGET_UPDATE_TYPE_UPDATE, action=GALA_GESTURE_ACTION_SWITCH_WINDOWS, progress=0)
    at ../src/Widgets/WindowSwitcher/WindowSwitcher.vala:252
#3  0x00005b5c423b4e71 in gala_gesture_target_propagate (self=0x5b5c7deb6620, update_type=GALA_GESTURE_TARGET_UPDATE_TYPE_UPDATE, action=GALA_GESTURE_ACTION_SWITCH_WINDOWS, progress=0) at ../src/Gestures/GestureTarget.vala:22
#4  0x00005b5c423b2d10 in gala_gesture_controller_set_progress (self=0x5b5c7deb6910, value=0) at ../src/Gestures/GestureController.vala:65
#5  0x00005b5c424016e4 in gala_window_switcher_select_icon (self=0x5b5c7deb6620, icon=0x5b5c7edd0680) at ../src/Widgets/WindowSwitcher/WindowSwitcher.vala:266
#6  0x00005b5c4240202f in gala_window_switcher_collect_all_windows (self=0x5b5c7deb6620, display=0x5b5c7dde7a40, workspace=0x5b5c7ddfa8d0) at ../src/Widgets/WindowSwitcher/WindowSwitcher.vala:349
[...]

PR #2337, opened 2025-03-25
Fix window switcher crash

lenemter:
Fixes https://github.com/elementary/gala/issues/2308


lenemter:
Oops just noticed there is a null check already...

PR #2338, opened 2025-03-25
WindowSwitcher: use Clutter.Stage.set_key_focus ()

lenemter:
Fixes https://github.com/elementary/gala/issues/2308

Not sure if this will fix the issue but worth a try I guess


leolost2605 (approved):
Let's try it :)

Issue #836, opened 2020-05-31
Fade out on shutdown


FIUSHerson:
## Prerequisites
- [ ] I have searched open and closed issues for duplicates.

## Feature
**Is your feature request related to a problem? Please describe.**

Feature request is not related to a problem.

**Describe the solution you'd like**

It’s a little jarring when you go to shut down the computer, and you see the screen go black with parts of open windows there for a split second.
Maybe a fade to black before the shutdown command is issued could work to add a little bit of polish.

**Existing work**

Can’t find any alternatives. I know Windows has a fade through black to a shutdown screen then shuts off the display, but you can’t really port that over easily at the moment...

I could try making it myself, but I don’t believe I have enough experience yet to do that with enough finesse.

If this is the wrong place to post the issue, please let me know where a better place would be, and I’ll happily move it over there.

cassidyjames:
Since the new locker stuff does this, is it just a matter of calling that before shutting down?

FIUSHerson:
@cassidyjames I wanna say the greeter on elementary is running on a different x server like gnome, but I think you’re on the right track.

I also know that gnome has a slow fade to black when falling asleep by itself. I haven’t taken a look at the source, but it could be a compositor thing. Maybe a solution could be taking that and applying it at a faster rate before the shutdown command is issued.

FIUSHerson:
> Since the new locker stuff does this, is it just a matter of calling that before shutting down?

Forget my last reply. I just woke up and didn’t read it completely :).

Yeah. Just calling it before shutdown is what I’m thinking.

danrabbit:
Poked David about this during standup and he said we’d need to pull that code out of the screensaver interface so that it could be re-used for shutdown since that’s a different interface

FIUSHerson:
Might also be great if it could be implemented in the sleep button as well.

PR #2343, opened 2025-03-27
Introduce ScreenShield

lenemter:
Move the actor and all the animations from SessionLocker into a separate class that can later be used for https://github.com/elementary/gala/pull/1999 and https://github.com/elementary/gala/issues/836


src/SessionLocker.vala
```diff
@@ -62,38 +62,54 @@ public class Gala.SessionLocker : Clutter.Actor {
...
+            if (!connected_to_buses) {
+                return;
+            }
 
-    public bool is_locked { get; private set; default = false; }
-    public bool in_greeter { get; private set; default = false; }
-    public int64 activation_time { get; private set; default = 0; }
+            if (_active != value) {
+                _active = value;
+                notify_property ("active");
```
leolost2605, line 82: IIRC you don't actually have to notify manually (we don't do explicit notify https://docs.gtk.org/gobject/signal.Object.notify.html)

src/Widgets/ScreenShield.vala
```diff
@@ -0,0 +1,115 @@
...
+        user_action ();
+        return Clutter.EVENT_STOP;
+    }
+
+    public override bool key_press_event (Clutter.Event event) {
+        user_action ();
+        return Clutter.EVENT_STOP;
+    }
+
+    public override bool motion_event (Clutter.Event event) {
```
leolost2605, line 52: Can we collect all events at once in `event (Clutter.Event event)` and filter them by type instead of having three handlers doing the same thing?

src/Widgets/ScreenShield.vala
```diff
@@ -0,0 +1,115 @@
...
+        } else {
+            active = true;
+            opacity = 255;
+
+            callback ();
+        }
+    }
+
+    public void deactivate () {
+        if (!active) {
```
leolost2605, line 94: IIUC we only set active to true once the animation completes so could this lead to us trying to deactivate while animating which would be ignored? And if so is that wanted?

src/Widgets/ScreenShield.vala
```diff
@@ -0,0 +1,115 @@
...
+
+    public override void parent_set (Clutter.Actor? old_parent) {
+        clear_constraints ();
+
+        unowned var parent = get_parent ();
+        if (parent == null) {
+            return;
+        }
+
+        add_constraint (new Clutter.BindConstraint (parent, Clutter.BindCoordinate.SIZE, 0.0f));
```
leolost2605, line 39: That's cool I didn't know this existed :)

leolost2605 (changes requested):
Makes sense to me but I currently can't test it because I've got a broken setup where I can't lock anything lol

PR #2349, opened 2025-03-29
WorkspaceClone: Activate workspace here

leolost2605:
I'd like to move workspace setup to the workspacerow so the less signals we have to connect the better :)


lenemter (changes requested):
It now always closes the view

https://github.com/user-attachments/assets/28bc3e73-0c7f-4a8a-bcb9-03654dfc249c

leolost2605:
whoops should be fixed :)

lenemter (approved):

PR #2350, opened 2025-03-29
Overlay window overview

leolost2605:
The first commit now only allows opening the windowoverview with window ids (it was already deprecated and not used anywhere but the code was still in place) and therefore does a bunch of cleanup regarding dynamically adding and removing windows.

The second commit includes some visual and ux changes:
- all the windows that are shown are now raised so if you have two on a workspace but a window from another app above them and then select one (or no window) both will be on top now. This fixes some incosistencies where they animate back but then suddenly a third window appears in between them.
- We don't hide other windows instead we are now on top (above the shell group) and just dim the whole background. This fixes some more incosistencies where other windows would just disappear and pop up again. Also the wingpanel and dock appeared interactable but weren't. It will now also make sense to show this while in the multitasking view (good for #2275).

Let me know what you think :)


lenemter:
> it was already deprecated and not used anywhere but the code was still in place

It is not, in fact, deprecated and is still used for the 'expose-all-windows' shortcut (Super+A), which shows windows from all workspaces. The multitasking view can't show all windows, so I guess we still need to keep the old window overview.

leolost2605:
> It is not, in fact, deprecated and is still used for the 'expose-all-windows' shortcut (Super+A), which shows windows from all workspaces. The multitasking view can't show all windows, so I guess we still need to keep the old window overview.

Huh since #485 was closed and I thought I'd remembered that in there I somewhere read that the expose all windows can be removed I thought this was deprecated. Seems like I was wrong :sweat_smile: 
I added the expose all windows back while keeping the cleanup and visual changes. 
New windows that appear while open are now still just added behind because I think the semantics here is you have your set of windows and choose one and not managing windows (that's what the multitasking view is for) but if really wanted we can change it that they get added to the spread (though if we want to do that, then I'd like to do it in a follow up since it's kind of broken in main as well).

src/Widgets/WindowOverview.vala
```diff
@@ -1,7 +1,7 @@
 /*
  * Copyright 2012 Tom Beckmann
  * Copyright 2012 Rico Tzschichholz
- * Copyright 2023 elementary, Inc. <https://elementary.io>
+ * Copyright 2025 elementary, Inc. <https://elementary.io>
```
lenemter, line 4:
```suggestion
 * Copyright 2023-2025 elementary, Inc. <https://elementary.io>
```

leolost2605:
https://github.com/user-attachments/assets/f5d6e1e2-b775-4ed3-9917-695e81e22e28

lenemter:
I'm -1 on bringing all app windows to top. It just feels unexpected..?

PR #2362, opened 2025-04-17
Remove #2308 from metainfo

lenemter:
#2308 is not present in the stable gala release, it appeared during transition to GestureContoller


danirabbit (approved):
Oops good to know

PR #2372, opened 2025-05-08
KeyboardManager: cleanup

lenemter:
- Don't make KeyboardManager object static
- Don't check for gsettings schema
- Move 'switch-input-source' keybindings here


danirabbit (approved):
Makes sense and looks good

Issue #180, opened 2018-02-17
Different keyboard layouts for individual windows


liminspace:
Hi there!
It's very important to have different layouts for each windows if you use more than one layout.
But it is impossible in Elementary OS.
Is there some solution how to fix it?
Thanks.

ps: Elementary OS can be the first candidate to move from Ubuntu with Unity, but it must have at least base functionality like above.

peteruithoven:
Could you explain why it's important? This sounds very subjective when you don't add some basic reasoning.

liminspace:
At least because it is not comfortable when you use several programs and must change layout every time changing window. After Ubuntu, Windows, MacOS or other OS, Elementary OS is very uncomfortable for its woody layouts.
If you use more that one layout you must understand me without additional explanation.

peteruithoven:
>  If you use more that one layout you must understand me without additional explanation.

Clearly I don't, which is why I'm asking this. Why are people using multiple layouts? I don't represent elementary OS, but I'd like to see some reasoning and clarification on the importance of this feature request.

liminspace:
Imagine you have a messenger, an email client, several terminals and a code editor. You use some A language in messenger, B in email client and C in terminals and the code editor. If OS doesn't support multi-window layouts, you must every time change the layout to write text in the right language and check if current layout is correct. If OS supports this feature, you can just change window without thinking about layout and spending time for checking and switching it.

oleg104:
This option is very important for all people who are from not english-speaking country like me and Ihor. We always have at least two input languages. Most of us has three input languages/loyouts. We use different languages for Internet browsing, messaging, programming. Please add option "Different keyboard layouts for individual windows" to Elementary OS. I saw that this option was present earlier like here https://goo.gl/images/75qiGc

SunDrop:
I do agree that this option is very important!
Can you show the way to return this option?

liminspace:
I just moved to Ubuntu 18.04. After some tuning it is quite comfortable and stable.

gasinvein:
Lack of this feature is probably a showstopper for anybody whose native language isn't english.

MegOBonus:
Any updates?

peteruithoven:
Please stop asking for updates. Maybe add a bounty or try to put in some time instead?

axelprox:
Maybe someone already solve this? Or AppCenter has app to deal with it?

Dirli:
Maybe someone decided that)) If there is a wish to test, welcome
[...]

PR #2373, opened 2025-05-08
Implement different keyboard layouts for individual windows

lenemter:
Fixes https://github.com/elementary/gala/issues/180
Requires https://github.com/elementary/gala/pull/2372

### DIFF
--- a/data/gala.metainfo.xml.in
+++ b/data/gala.metainfo.xml.in
@@ -39,6 +39,7 @@
       </description>
       <issues>
         <issue url="https://github.com/elementary/gala/issues/137">Send a persistent notification after taking a screenshot</issue>
+        <issue url="https://github.com/elementary/gala/issues/180">Different keyboard layouts for individual windows</issue>
         <issue url="https://github.com/elementary/gala/issues/321">Use org.gnome.desktop.interface enable-animations</issue>
         <issue url="https://github.com/elementary/gala/issues/544">Multitasking View. Wingpanel shows for brief moment if a fullscreen wokspace is selected after coming from a wokspace that shows the wingpanel.</issue>
         <issue url="https://github.com/elementary/gala/issues/823">Check for redacted font</issue>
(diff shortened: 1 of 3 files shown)


leonardo-lemos (changes requested):
I tested with two layouts: pt_br and en_US (intl.). When I set a layout (eg. pt_br) and use it in a window, I can't change window layout to another layout (eg. en_US), Gala always reverts the layout to the first one.

lenemter:
@leonardo-lemos what do you mean 'Gala always reverts the layout to the first one'? It works fine for me

leonardo-lemos:
@lenemter There are two problems to me:

1. I can't change layout using the keyboard indicator, because if I'm using a layout A in a window and try to change to a layout B, the indicator always reverts state to layout A
2. If I'm using terminal with layout A and and code with layout A, then go back to terminal and change to layout B using a keyboard shortcut, code will be set to use layout B too.

lenemter:
@leonardo-lemos I fixed interaction with the keyboard indicator. But I cannot reproduce the second issue you have. Can you check again now, maybe it got fixed on your side?

leonardo-lemos:
@lenemter It works as expected when changing layouts using the indicator, but doesn't work properly when changing layouts using the keyboard (using the change layout shortcut). When I change layout using the shortcut, it changes layout for all the windows.

lenemter:
@leonardo-lemos Can you provide steps to reproduce? It works fine for me ¯\\\_(ツ)\_/¯

leonardo-lemos:
@lenemter Sure

https://github.com/user-attachments/assets/10e1c6a3-dfc9-466b-90ed-2113efbccb95

1. Make sure you have two keyboard layouts enabled: Portuguese (Brazil) and English (US).
2. Open a window (e.g., a text editor or browser) with the English (US) layout active.
3. Type something. In my case, I pressed the ´ key on my pt-br keyboard, which corresponds to [ in the en-us layout.
4. Switch to a different window.
5. Change the keyboard layout using the keyboard shortcut (e.g., Alt + Shift).
6. Confirm that the layout has switched to pt-br. Type ´ and verify that it appears correctly on the screen.
7. Return to the first window. You'll see that the keyboard layout has also changed to pt-br, even though it was previously set to en-us.

This seems to be an issue with Gala or the keyboard layout indicator, rather than something caused by the changes in this PR. Could you confirm?

PR #2374, opened 2025-05-08
DesktopIntegration: Don't resend windows when switching workspace

lenemter:
Requires https://github.com/elementary/dock/pull/406


leonardo-lemos (approved):
LGTM

Issue #2387, opened 2025-05-15
Update broke secondary display workspace


teamcons:
### What Happened?

this one appeared at the same time as https://github.com/elementary/gala/issues/2322 in a recent update
Making Gala effectively a one-workspace, one-monitor window manager

going in multitasking, the window on the secondary screen is just a grey empty space.
this happens only in multitasking.
The window on that secondary screen does not show anywhere there

![Image](https://github.com/user-attachments/assets/6b39d374-ee62-4980-9e8a-d3a3bfa990cf)

### Steps to Reproduce

have a secondary screen
have possibly this:  https://github.com/elementary/gala/issues/2322 (so up to date system i guess)

try using multitasking

### Expected Behavior

workspace in multitasking on secondary screen

### OS Version

8.x (Circe)

### Session Type

Secure Session (Wayland)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell
nothing new seems to show in dmesg when trying this

> sudo tail -f /var/log/syslog



2025-05-15T13:54:35.824079+02:00 Thonkpod io.elementary.d[1722]: gtk_box_remove: assertion 'gtk_widget_get_parent (child) == (GtkWidget *)box' failed
2025-05-15T13:54:35.827240+02:00 Thonkpod io.elementary.d[1722]: message repeated 5 times: [ gtk_box_remove: assertion 'gtk_widget_get_parent (child) == (GtkWidget *)box' failed]
2025-05-15T13:54:36.234315+02:00 Thonkpod gala[1631]: meta_workspace_index: assertion 'META_IS_WORKSPACE (workspace)' failed
2025-05-15T13:54:36.237391+02:00 Thonkpod io.elementary.d[1722]: gtk_box_remove: assertion 'gtk_widget_get_parent (child) == (GtkWidget *)box' failed
2025-05-15T13:54:36.237600+02:00 Thonkpod io.elementary.d[1722]: gtk_box_remove: assertion 'gtk_widget_get_parent (child) == (GtkWidget *)box' failed
2025-05-15T13:54:40.530997+02:00 Thonkpod gala[1631]: meta_workspace_index: assertion 'META_IS_WORKSPACE (workspace)' failed
2025-05-15T13:54:40.531747+02:00 Thonkpod gala[1631]: meta_workspace_index: assertion 'META_IS_WORKSPACE (workspace)' failed
[...]

Issue #2322, opened 2025-03-17
Workspace Switch


lildearn:
### What Happened?

The workspace quit switching after using it when I updated. I'm not sure what package caused the bug to surface.

### Steps to Reproduce

It's hard to reproduce. You have to have a workspace glitch, and it will only work when you double-click in the window overview, not in the dock. Workspaces don't switch window focus correctly. They quit working after awhile.

### Expected Behavior

Workspaces should switch focus to different workspace.

### OS Version

8.x (Circe)

### Session Type

Secure Session

### Software Version

Latest release (I have run all updates)

### Log Output

```shell
`ar 17 14:58:22 ubuntu gala[3968]: instance with invalid (NULL) class pointer
Mar 17 14:58:22 ubuntu gala[3968]: signal_emit_valist_unlocked: assertion 'G_TYPE_CHECK_INSTANCE (instance)' failed
Mar 17 14:58:30 ubuntu gala[3968]: Window manager warning: META_CURRENT_TIME used to choose focus window; focus window may not be correct.
Mar 17 14:58:31 ubuntu gala[3968]: Window manager warning: META_CURRENT_TIME used to choose focus window; focus window may not be correct.
Mar 17 14:58:31 ubuntu gala[3968]: Window manager warning: META_CURRENT_TIME used to choose focus window; focus window may not be correct.
Mar 17 14:58:33 ubuntu gala[3968]: Window manager warning: META_CURRENT_TIME used to choose focus window; focus window may not be correct.
Mar 17 14:58:33 ubuntu gala[3968]: Window manager warning: META_CURRENT_TIME used to choose focus window; focus window may not be correct.
Mar 17 14:58:33 ubuntu gala[3968]: Window manager warning: META_CURRENT_TIME used to choose focus window; focus window may not be correct.
Mar 17 15:02:25 ubuntu gala[3968]: Window manager warning: META_CURRENT_TIME used to choose focus window; focus window may not be correct.
Mar 17 15:02:40 ubuntu gala[3968]: Window manager warning: META_CURRENT_TIME used to choose focus window; focus window may not be correct.
Mar 17 15:02:41 ubuntu gala[3968]: Window manager warning: META_CURRENT_TIME used to choose focus window; focus window may not be correct.
Mar 17 15:02:45 ubuntu gala[3968]: Window manager warning: META_CURRENT_TIME used to choose focus window; focus window may not be correct.
[...]

PR #2388, opened 2025-05-15
WorkspaceManager: check monitor

lenemter:
Investigating https://github.com/elementary/gala/issues/2322 and https://github.com/elementary/gala/issues/2387...


danirabbit:
I can reproduce the issue with workspaces no longer switching, but oddly only when I install gala from packaging. Compiling it manually the issue doesn't persist. Not sure if that helps. But wanna make sure to test a fix from packaging to avoid that I suppose

lenemter:
Thanks! I can confirm the issue by manually building the deb package

PR #2389, opened 2025-05-15
Fix workspace switch

lenemter:
Fixes https://github.com/elementary/gala/issues/2322


leonardo-lemos (approved):

danirabbit (approved):
This appears to work for me! Nice job getting the fix out so fast

PR #2392, opened 2025-05-16
Add support for restoring previous region

lenemter:
Currently in gala 8.2.1, if you open multitasking view on X11 and close it, the dock clickable area will be a lot bigger then its visible part.

Unseting window shape region entirely doesn't work for our Dock since its window size is a lot bigger than its visible part. This branch introduces `restore-previous-region` argument for mutter hints. If it's set, we restore previous region (dock needs this), if not, unset region (wingpanel needs this).


leonardo-lemos:
It works as expected, thanks for your hard work! :heart:

PR #2393, opened 2025-05-16
MultitaskingView: Always show monitor clones

lenemter:
Fixes https://github.com/elementary/gala/issues/2387

I don't have second monitor to test but I suppose this should fix it


leonardo-lemos (approved):
I can confirm that it fixes the issue. Nice job!

Issue #2401, opened 2025-05-19
Context menus appear at the top left corner of the secondary monitor


TomiOhl:
### What Happened?

I have a setup where I have an external monitor above my laptop screen.
System-level context menus that I trigger on the laptop screen appear at the top left of the external monitor.
By system-level context menus I mean right clicking on window title bars or on an empty space on the desktop wallpaper.
E.g. the context menu when clicking on the title bar of Files:
![Image](https://github.com/user-attachments/assets/7eca585b-c72a-4419-98b0-5429b5483481)
The issue is not present if I'm right clicking on anything on the top display.

Update: if I place my external display to the right of the laptop screen, the context menus triggered on the laptop screen will appear at the right place, however, context menus triggered on the external monitor will appear on the top left corner of the laptop screen (i.e. the top-left-most section of the view?)

### Steps to Reproduce

1. Connect an external monitor
2. Right click a title bar or an empty space on the wallpaper.

### Expected Behavior

Context menu right where the pointer is

### OS Version

8.x (Circe)

### Session Type

Secure Session (Wayland)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell

```

PR #2407, opened 2025-05-20
Rewrite menus

lenemter:
Should fix https://github.com/elementary/gala/issues/2401 but I cannot test since I use single-monitor setup.

Instead of making daemon's widow full display size, make it the size of a monitor which makes a lot more sense in Wayland

### DIFF
--- a/daemon-gtk3/DBus.vala
+++ b/daemon-gtk3/DBus.vala
@@ -64,7 +64,7 @@ public class Gala.Daemon.DBus : GLib.Object {
         }
     }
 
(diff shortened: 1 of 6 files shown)


leonardo-lemos (changes requested):
I can open a menu and it works flawlessly but if I try to open it again, it opens and closes very fast. This happens with any window.

leonardo-lemos:
@lenemter I got the same result after the lastest commit, plus If I open desktop context menu on the second monitor, It opens on the display's corner or in the primary monitor but the place where I left click.

PR #2411, opened 2025-05-28
GestureController: Handle gesture interruption transparently

### LINKED ISSUE #2409: Windows sometimes get stuck in an un-interactive state after using gestures to switch workspaces
### What Happened?

rarely, but especially when a gesture gets interrupted, swiping between workspaces or into the overview on a touchpad will cause all windows to be non-interactible and have a little `X` icon like seen in the overview, **regardless** of whether the overview is open

Almost all keyboard shortcuts and interacting with the wingpanel also becomes impossible; <kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>Delete</kbd> brings up the logout prompt but it's also non-interactive, i.e it cannot be used to logout to fix the issue

### Steps to Reproduce

1. Start doing a gesture
2. Interrupt it? (start with three fingers, add a 4th one, or otherwise disrupt it)
3. Profit???

(I wish I could give more detail really)

### Expected Behavior

Gala should correctly recover from gestures being interrupted (?)

### OS Version

8.x (Circe)

### Session Type

Secure Session (Wayland)

### Software Version

### Log Output

```shell
N/A
```

### Hardware Info

_No response_

leolost2605:
Fixes #2409 

Once one finger gets lifted we switch from touch egg backend to scroll backend which would emit end and start signals which would cause a hide and show on the multitasking view which in turn would interrupt and restart the scroll gesture but not emit a end because of a wrong order. We can solve all these problems if we just don't emit start and end when the gesture gets interrupted while still going. This also is a small performance improvement because we don't go hidden shown everytime which saves us a bunch of recalculations.
Also don't allow a new gesture to start while one is already going.

(thanks @zeebok for the reproducer)

### DIFF
--- a/src/Gestures/GestureController.vala
+++ b/src/Gestures/GestureController.vala
@@ -125,17 +125,20 @@ public class Gala.GestureController : Object {
 
     private void prepare () {
         if (timeline != null) {
-            timeline.stop ();
             timeline = null;
+        } else {
+            target.propagate (START, action, progress);
         }
-
-        target.propagate (START, action, progress);
     }
 
     private bool gesture_detected (GestureBackend backend, Gesture gesture, uint32 timestamp) {
+        if (recognizing || !enabled) {
+            return false;
+        }
+
         var recognized_action = GestureSettings.get_action (gesture, out _action_info);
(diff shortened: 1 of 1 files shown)


lenemter (approved):
LGTM

PR #2413, opened 2025-05-30
Daemon: Use size request instead of default size

lenemter:
Fixes https://github.com/elementary/gala/issues/2401

I tested the fix on wayland with qemu+spice but I cannot test it on X11 :(


leonardo-lemos (approved):
LGTM

Issue #2412, opened 2025-05-28
Multitasking View is oversized with fractional scaling


danirabbit:
### What Happened?

With fractional scaling set to 150%, the multitasking view appears to be oversized:

![Image](https://github.com/user-attachments/assets/1a6c62c1-8a00-402d-ac4d-072f28bf4fc8)

### Steps to Reproduce

1. Enable experimental fractional scaling and set to 150%
2. Enter multitasking view

### Expected Behavior

Things should be scaled appropriately and appear similar to how they do in an integer scaling setup

### OS Version

8.x (Circe)

### Session Type

Secure Session (Wayland)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell

```

### Hardware Info

Starlite MkV with the new 2160x1440 panel

PR #2416, opened 2025-05-30
WindowIcon: Support fractional scaling

lenemter:
Addresses https://github.com/elementary/gala/issues/2412

I decided to not touch icon groups because the code there is already broken and I don't wanna break it further since we plan to replace them with the dock some day


danirabbit:
Hmm can't confirm that worked

Screenshot from 2025-05-30 10 26 58@2x

PR #2419, opened 2025-05-31
ShellWindow: Clip window actor when applying translations

leolost2605:
Fixes https://github.com/elementary/dock/issues/425
Closes #2418 

There are things in the shell group that can be on secondary monitors (most notably freely movable transient windows like the connect to hidden network window). Also sometime soon the monitor labels should go there so they will need to be on secondary monitors. So instead of clipping the whole group only clip the respective actors.

### DIFF
--- a/src/ShellClients/ShellWindow.vala
+++ b/src/ShellClients/ShellWindow.vala
@@ -147,18 +147,19 @@ public class Gala.ShellWindow : PositionedWindow, GestureTarget {
     }
 
     private void update_clip () {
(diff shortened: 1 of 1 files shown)


leolost2605:
Ok the issue was that since we position according to frame rect and not actual window actor position a part of the dock is always on the second monitor. So now instead of just offsetting the translation we clip to the actual monitor dimensions.

software-artificer:
Thanks a lot folks, appreciate the quick turnaround!

PR #2423, opened 2025-05-31
WindowSwitcher: correctly draw border on HiDPI

lenemter:
1. Correctly scale stroke width
2. Offseting border by 0.5 works only on 1.0 scaling, divide our stroke width by 2 instead
3. Shadow's border radius should be 10 because it's one pixel outside of window switcher
4. Don't scale shadow's border radius, the effect does this automatically


src/Widgets/WindowSwitcher/WindowSwitcher.vala
```diff
@@ -211,9 +209,9 @@ public class Gala.WindowSwitcher : CanvasActor, GestureTarget, RootTarget {
 
         // Offset by 0.5 so cairo draws a stroke on real pixels
```
lenemter, line 210:
```suggestion
```

PR #2425, opened 2025-06-02
Release the Barrier once the threshhold is reached

### LINKED ISSUE #2424: Mouse pointer often stuck between displays when external display is the primary display and is arranged above built-in display
### What Happened?

The mouse pointer gets stuck at the edge of displays when trying to move the pointer across displays.

This happens when the external display is:

- Set as the primary display
- Arranged to be above the built-in display

### Steps to Reproduce

1. Set external display as primary display
2. Arrange displays do that external display is above the built-in display
 Attempt to move the mouse pointer across the displays. The pointer will often get stuck at the edge of the displays.

Screenshot from 2024-10-31 18 26 11

### Expected Behavior

The mouse pointer should freely move across the displays.

It works perfectly fine when the built-in display is set as the primary display.

### OS Version

8.x (Early Access)

### Session Type

Secure Session (Wayland)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

leolost2605:
We want to release the barrier once it was triggered for a certain amount of time to allow the cursor to travel to a new monitor. We can play with the amount of time a bit but 200 seemed ok to me :shrug: 

Fixes #2424 
Fixes elementary/dock#68


lenemter (approved):
Code looks good. But I can't test it since I don't have a second monitor

leonardo-lemos (changes requested):
It works, but I think it could be improved. The cursor transition between monitors doesn't feel smooth — it gets stuck between them for too long. Maybe reducing the delay time would help.

Issue #340, opened 2018-09-21
"Move to left workspace" is a no-op on left-most workspace


jlnr:
Having a no-op in the menu is confusing. I guess this would be fixed by #187, but for now it might be an easier quick fix to disable the entry on the workspace?

PR #2431, opened 2025-06-06
Daemon: disable "Move to Workspace" when it does nothing

lenemter:
Fixes https://github.com/elementary/gala/issues/340


leonardo-lemos (approved):
LGTM

PR #2436, opened 2025-06-09
Remove 'Open in Photo Viewer' action from screenshot notification

lenemter:
https://github.com/elementary/notifications/issues/237 is not resolved and therefore this looks ugly on some locales


leonardo-lemos (approved):

danirabbit:
Tbh I would have preferred to remove the show in files button since you can show in files from within the photo viewer and opening in photos gives you more immediate options for cropping etc that folks had previously asked for in screenshot

Marukesu:
I'm a little late in both PRs, but we could set "open in photos" as the default action instead of using a button, Then we don't need to translate any string, nor have a button that could break the notification layout.

Issue #2438, opened 2025-06-10
Gala randomly segfaults while closing Libreoffice windows


djaeris:
### What Happened?

Under some unknown circumstances, gala will segfault when closing windows belonging to Libreoffice. If the document is not saved, the app will issue an save dialog, and will crash after the dialog is dismissed (e.g. by hitting the save button) and window should be closed. 

### Steps to Reproduce

1. Close a Libreoffice window
2. Get kicked out of session and into the greeter

### Expected Behavior

The window being closed should close and nothing else.

### OS Version

8.x (Circe)

### Session Type

Secure Session (Wayland)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell
jún 10 12:10:24 aerisbase gala[1794]: meta_workspace_index: assertion 'META_IS_WORKSPACE (workspace)' failed
jún 10 12:10:24 aerisbase gala[1794]: meta_workspace_index: assertion 'META_IS_WORKSPACE (workspace)' failed
jún 10 12:10:25 aerisbase gala[1794]: **
jún 10 12:10:25 aerisbase gala[1794]: ERROR:src/gala.p/WindowStateSaver.c:774:vala_g_array_remove_index: assertion failed: (length > index)
jún 10 12:10:25 aerisbase gala[1794]: Bail out! ERROR:src/gala.p/WindowStateSaver.c:774:vala_g_array_remove_index: assertion failed: (length > index)
jún 10 12:10:25 aerisbase systemd[1]: Created slice system-systemd\x2dcoredump.slice - Slice /system/systemd-coredump.
jún 10 12:10:25 aerisbase systemd[1]: Started systemd-coredump@0-13449-0.service - Process Core Dump (PID 13449/UID 0).
jún 10 12:10:25 aerisbase systemd-coredump[13450]: Process 1794 (gala) of user 1000 dumped core.
                                                    
                                                    Module libgcc_s.so.1 from deb gcc-14-14.2.0-4ubuntu2~24.04.amd64
                                                    Module libstdc++.so.6 from deb gcc-14-14.2.0-4ubuntu2~24.04.amd64
                                                    Module libzstd.so.1 from deb libzstd-1.5.5+dfsg2-2build1.1.amd64
                                                    Module libudev.so.1 from deb systemd-255.4-1ubuntu8.8.amd64
                                                    Module libsystemd.so.0 from deb systemd-255.4-1ubuntu8.8.amd64
                                                    Stack trace of thread 1794:
[...]

PR #2443, opened 2025-06-12
WindowStateSaver: fix crash

lenemter:
Fixes https://github.com/elementary/gala/issues/2438


vjr (approved):
Code change makes sense.

PR #2459, opened 2025-07-07
ShellWindow: Disable unredirection when visible

leolost2605:
Shell windows are in the shell group and are therefore above fullscreen windows. They can also be visible when a window is fullscreen via the new reveal feature. Also the animation plays when a window is already full screen. So we want to make sure to disable unredirection when we are visible so that we are actually visible.

Fixes an issue where when you are in a fullscreen firefox playing a video or in videos playing a video you wouldn't be able to reveal the wingpanel.


lenemter (approved):
LGTM

PR #2463, opened 2025-07-10
Draw rounded rectangles using shader

lenemter:
Significantly optimizes our drawing code by using shader to draw rounded rectangles. Multitasking view gets a huge performance boost by not manually checking/drawing ActiveShape since it's now done on GPU.

The shaders code looks weird but I just couldn't make it produce consistent results. I played with offsets and this configuration gives the most symmetrical and sharp results

### DIFF
--- a/lib/RoundedCornersEffect.vala
+++ b/lib/RoundedCornersEffect.vala
@@ -60,6 +60,8 @@ public class Gala.RoundedCornersEffect : Clutter.ShaderEffect {
             actor.height
         };
 
+        warning ("%f", actor.height);
+
         var bounds_value = GLib.Value (typeof (Clutter.ShaderFloat));
         Clutter.Value.set_shader_float (bounds_value, bounds);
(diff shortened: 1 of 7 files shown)


lenemter:
I don't think this is necessary as in many cases (e.g. Tooltip) the border radius didn't align with our stylesheet

lib/RoundedCornersEffect.vala
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    public override void set_actor (Clutter.Actor? actor) {
+        base.set_actor (actor);
+
+        if (actor == null) {
+            return;
+        }
+
+        actor.notify["width"].connect (update_actor_size);
```
leolost2605, line 57: maybe it would be good to disconnect this if a new actor is set even though we never use it rn it might lead to problems later 🤷

leolost2605 (approved):
Can't say much about the shader code and can't test right now but apart from that LGTM :) 🚀⚡

Issue #2368, opened 2025-05-01
Build crashes


vyvir:
### What Happened?

Trying to compile the latest Gala commit on OpenSUSE Leap 15.6, but build fails

### Steps to Reproduce

1. Compiled latest Gala with OBS (https://build.opensuse.org/package/show/home:asdhio:branches:X11:Pantheon/gala)
2. Build failed

### Expected Behavior

1. Compiled latest Gala with OBS (https://build.opensuse.org/package/show/home:asdhio:branches:X11:Pantheon/gala)
2. Build succeeded

### OS Version

Other Linux

### Session Type

Classic Session (X11, This is the default)

### Software Version

Compiled from git

### Log Output

```shell
[   44s] ../src/Gestures/ActorTarget.vala:33.9-33.19: error: The name `child_added' does not exist in the context of `Gala.ActorTarget'
[   44s]    33 |         child_added.connect (on_child_added);
[   44s]       |         ^~~~~~~~~~~                          
[   44s] Compilation failed: 1 error(s), 0 warning(s)
[...]

PR #2479, opened 2025-07-17
Add OpenSuse Leap CI

lenemter:
Not sure if we want to support this, but I don't see reasons why not. Makes fixing https://github.com/elementary/gala/issues/2368 easier since we still 'support' Mutter 45


danirabbit (approved):
Can't hurt to have more CI. Makes it easier to catch breakages for our downstreams!

PR #2482, opened 2025-07-22
Fix OpenSUSE Leap build

lenemter:
Fixes https://github.com/elementary/gala/issues/2368


danirabbit (approved):
CI passes. Nice work :rocket:

Issue #1680, opened 2023-06-08
Filters applied to non-window screenshots


lenemter:
### What Happened?

When screenshoting a window with enabled filters, screenshots doesn't reflect the filter.

### Steps to Reproduce

1. Turn on any filter (e.g. grayscale)
2. Screenshot a window

### Expected Behavior

Screenshot with a filter.

### OS Version

7.x (Horus)

### Software Version

Latest release (I have run all updates)

danirabbit:
I don’t think this is a bug. I think it’s a feature. I’m not sure there’s really value in having these filters applied to a screenshot since it’s your personal accommodation

lenemter:
> I don’t think this is a bug. I think it’s a feature. I’m not sure there’s really value in having these filters applied to a screenshot since it’s your personal accommodation

But filters are applied to screen screenshots currently. Should we disable this behaviour then?

tintou:
The filter shouldn't be applied to the screenshot content, otherwise:
 * Other people will see that you have a filter enabled if you share the image, and that's none of their business
 * We end-up with the filter being applied twice if we then view the screenshot with the same filter

PR #2486, opened 2025-07-24
Pause filters before taking screenshots

lenemter:
Fixes https://github.com/elementary/gala/issues/1680


leolost2605:
any reason we don't just remove the effect and add it back?

lenemter:
@leolost2605 FilterManager has some complicated (and broken) logic for transitions. It's easier to add a flag in the shader than to remove effect mid-transition, remember it's state and then restore it

leolost2605 (approved):

Issue #2057, opened 2024-09-18
v8: Trying to change desktop bg to a color don´t work


megatux:
### What Happened?

Tried to set a color as bg background, not an image.
I was set as black but don´t apply any color.

### Steps to Reproduce

Go to bg settings from right click on desktop
Select a color

### Expected Behavior

Apply the color to the desktop

### OS Version

8.x (Early Access)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell
daily build from 9/17
```

### Hardware Info

Two displays. External fullHD LG and laptop fullHD.

### Screenshot

![Captura de pantalla de 2024-09-18 01 01 16](https://github.com/user-attachments/assets/83e485a9-aae7-4c14-9537-b19e205413aa)

BTW, it`s also odd that the color picker dialog opens in the other display and not the same display than the origin window :

![Captura de pantalla de 2024-09-18 01 11 08](https://github.com/user-attachments/assets/ede0d3b4-090a-43fd-adfe-f6d1d93a203f)

PR #1819, opened 2024-01-04
Remove content.set_rounded_clip_bounds

leolost2605:
Like I said in [this comment](https://github.com/elementary/gala/issues/1815#issuecomment-1877226917) I've got no idea what this is doing exactly but removing it fixes #1815 with no regressions found 🤷

PR #2487, opened 2025-07-25
Fix changing background color

lenemter:
Fixes https://github.com/elementary/gala/issues/2057


lenemter:
@leolost2605 Using `set_rounded_clip_bounds` caused you some problems earlier (#1819) Can you check if you encounter them in this branch, please?

leolost2605:
It works on my laptop but I'll test later today with multiple monitors :)

leolost2605 (approved):
Works and LGTM :)

PR #2488, opened 2025-07-25
Cleanup Wayland sockets on logout

lenemter:
Fixes https://github.com/elementary/gala/issues/2489
The solution is taken from https://github.com/flatpak/flatpak/issues/3880#issuecomment-1265349004


ryonakano (approved):
Confirmed Flatpak apps still runs even after switching from Classic to Secure and Secure to Classic.

Issue #2079, opened 2024-10-05
Minor: Gala app icon shows up in Dock briefly


teamcons:
### What Happened?

When right-clicking titlebars to show the window manager menu (Always On Top, Show on all desktops, etc...), the dock enlarges to make space for an app icon before downsizing immediately. Usually before the icon appears

In some cases the gala app icon appears

### Steps to Reproduce

With dock being present:
Right click on titlebar / app bar? of an application
Look at what the dock does

### Expected Behavior

The dock should not react to gala context menu appearing

### OS Version

8.x (Early Access)

### Session Type

Secure Session (Wayland)

### Software Version

Latest release (I have run all updates)

PR #2490, opened 2025-07-25
WindowTracker: ignore windows from our Wayland clients

lenemter:
Fixes https://github.com/elementary/gala/issues/2079

Makes DaemonManager use ManagedClient instead of reimplementing it

### DIFF
--- a/src/DaemonManager.vala
+++ b/src/DaemonManager.vala
@@ -18,7 +18,7 @@ public class Gala.DaemonManager : GLib.Object {
 
     public Meta.Display display { get; construct; }
 
-    private Meta.WaylandClient daemon_client;
+    private ManagedClient client;
     private Daemon? daemon_proxy = null;
 
     public DaemonManager (Meta.Display display) {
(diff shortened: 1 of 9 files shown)


leolost2605 (commented):
I kinda don't love having yet another place where we can check whether something isn't a normal window.
Also in theory wayland clients may launch normal windows that would falsely be ignored here. And the other way round more windows can probably be ignored that aren't explicitly launched by a wayland client.
Can we maybe add this to ShellClientsManager.is_positioned_window (and maybe rename that to is_shell_window since it does more than just positioned windows for quite some time idk) similarly to how notifications are handled there?
Let me know what u think :)

lenemter:
@leolost2605 We currently use ManagedClient only to launch shell windows so I guess it's fine. By the way, gnome recently merged API to set window types directly, so we may want to remove ManagedClient and WaylandClient in OS 9 (https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/4491)

leolost2605 (approved):
LGTM

PR #2496, opened 2025-07-27
MultitaskingView/Gesture: Add overshoot animation

### LINKED ISSUE #1026: 3/4 finger over-swiping down on the desktop should zoom the wallpaper a bit
## Problem

If the user swipes up they get a 1:1 animation entering the multitasking view, doing the opposite on the desktop does nothing. 

## Proposal

The desktop  should "overscroll" the animation and zoom the desktop wallpaper a bit. 
There should be a little animation to convey that the action not doable just like the workspace end overscroll. 

## Prior Art

iOS sort of has this

leolost2605:
Fixes https://github.com/elementary/gala/issues/1026

It currently can't be triggered in one go due to the touchegg backend stopping at 100% but that will be fixed with the new touchpad backend I'll hopefully soon start finishing :)

To test right now you can in the open multitasking view try opening it and for the other direction while closed try closing it.


lenemter (changes requested):
There are a couple of issues:
1. The logs are filled with `value "4294967295" of type 'guint' is invalid or out of range for property 'opacity' of type 'guint'`
2. If you swipe up too much, window clone can get partially off screen. Also, when swiping down on the desktop, workspace wallpapers may overlay.

Screenshot from 2025-07-31 19 21 23

Screenshot from 2025-07-31 19 27 05

Issue #649, opened 2019-12-08
PiP action desktop file


donadigo:
It seems like we're getting a lot of "I never knew PiP existed in the system", even though we have the shortcut displayed in the shortcuts overlay, I think we could include a desktop file for initiating PiP, much like we do with multitasking view.

cassidyjames:
I'm not super on board with adding a .desktop file for Gala features. Multitasking View is there because we want it on the dock by default, but we really would prefer it not go in Applications Menu because it's not an app. But it's there now because otherwise there's not a good way to get it back onto the dock if it's accidentally removed.

I think long term a better route would be to integrate PiP into apps themselves; manually-invoked PiP is handy but it really feels like a workaround. Being more integrated into i.e. the Videos app could help expose it as a feature, which then would help people know what it is when they see that feature in the Shortcuts Overlay.

sprite-1:
> But it's there now because otherwise there's not a good way to get it back onto the dock if it's accidentally removed.

How about adding it to the Docklet portion of the Plank's menu?

![image](https://user-images.githubusercontent.com/33685433/70453101-27737d80-1ae3-11ea-8348-d4c5d43780ef.png)

cassidyjames:
@sprite-1 yeah I think that's a better place, though it's still pretty hidden. But we talked about implementing it either as a hidden .desktop or an applet, then having a switch in _System Settings_ → _Desktop_ → _Dock_ to enable it maybe.

sprite-1:
@cassidyjames I hope it's still movable in the icon order when/if you do end up making it a System Setting entry

danrabbit:
After some discussion in Slack it sounds like we're maybe in agreement that instead of a launcher this would make more since as a menu item in the header context menu. @cassidyjames can we go ahead and confirm and update the issue title to that effect?

PR #2508, opened 2025-07-29
Rewrite window menu system

lenemter:
Introduces WindowMenuManager that controls window menu items and what they do. This gives us a lot more control over window menus in general since now wm has proper callbacks support for window menu items. I plan to use it for https://github.com/elementary/gala/issues/649. Plus this introduces ability to write custom plugins with this functionality (e.g. for https://github.com/elementary/gala/pull/1493 because likely it won't be merged) 

Allows us to deprecate Gala.ActionType in the future and cleanup DBus interface with the window manager implementation of it, but that should be done in OS 9.


leolost2605:
@lenemter sorry that I never got around to do a review here, I always kinda skipped over it because I didn't have much time and it is a big diff. But I've got a week or two with plenty of time now so if you're still interested in this feel free to update it with the new changes from main and I'll be happy to do a review :)
Also I don't know if you've considered it and found it unfitting for this use case but I just want to quickly point to [DBusMenuModel](https://valadoc.org/gio-2.0/GLib.DBusMenuModel.html) which we might be able to use here?

lenemter:
@leolost2605 I didn't know about DBusMenuModel but I don't think it fits this use case. Window menu items can change depending on the window and I don't know if DBusMenuModel supports hiding/showing items changing descriptions on the fly, etc. Also it seems to be tied to actions which we don't use in gala in general

Issue #1069, opened 2021-03-02
There isn't a way to drag windows and items between workspaces


hanaral:
## What Happened

There is currently no way to drag an item or a window from one workspace to another. Trying to drag a window to the top or the sides of a desktop just tiles it rather than moving it. It is only possible for windows to be moved from the overview - but there is nothing for actual DnD despite elementary HIG seeing window management being done via workspaces rather than showing/hiding them

## Expected Behavior

1. Dragging a window titlebar or item to against the top of the workspace opens the multitasking overview (with the item still being dragged)
2. Hovering over a workspace in the same spot with the item enters that workspace

## Platform Information

elementary OS 6 Daily

lenemter:
Fixed in https://github.com/elementary/gala/pull/2275

PR #2512, opened 2025-07-31
Allow dropping windows directly to dock

lenemter:
Fixes https://github.com/elementary/gala/issues/1069
Fixes https://github.com/elementary/gala/issues/569

@danirabbit What do you think about this?

### DIFF
--- a/src/DBus.vala
+++ b/src/DBus.vala
@@ -18,7 +18,7 @@ public class Gala.DBus {
             SESSION, "io.elementary.gala", NONE,
             (connection) => {
                 try {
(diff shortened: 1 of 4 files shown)


danirabbit:
I was a huge fan of the other prototypes of #569. I'm not seeing that I can drag windows to the dock outside of multitasking view with this branch however. Am I missing something?

danirabbit:
Ah okay, I can see now! One thing that was nice about the previous prototype for #569 is that it did the thing we do in multitasking view of iconifying the window. Otherwise it's hard to see from just the icon appearing in the workspace switcher.

Also, this seems to crash for me every time I try to drag a window to the workspace switcher while in multitasking view

PR #2518, opened 2025-08-02
Introduce GestureAction.CUSTOM

lenemter:
Replaces our DOCK and CLOSE_WINDOW gesture actions with CUSTOM. This allows us to use this action in plugins (e.g. for hiding PiP). This change shouldn't affect anything since users of DOCK and CLOSE_WINDOW are self-contained and don't have complicated target hierarchy.


leolost2605 (approved):
IMO we could also just add the custom action and leave the other two there but for me it's fine either way 🤷

PR #2536, opened 2025-09-17
Support keyboard navigation over multiple monitors in multitasking view

### LINKED ISSUE #2648: Pressing Esc doesn't close multitasking view
### What Happened?

Pressing Esc doesn't close multitasking view after switching workspace by pressing on the workspace

### Steps to Reproduce

1. Open 1 window on the 1st workspace
2. Open multitasking view
3. Press on the 2nd workspace background
4. Press Esc

### Expected Behavior

Multitasking view closes

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

Secure Session (Wayland)

### Software Version

Compiled from git

### Log Output

```shell

```

### Hardware Info

_No response_

### LINKED ISSUE #2651: Cannot select window by keyboard in multitasking view after workspace switching
### What Happened?

Hey,

I discovered this feature in multitasking view, where you can select a window with the keyboard using `left`, `right`, `up`, and `down`. In other simple window managers (such as Openbox), there is directional focus, but here, it is even smarter to do that in the multitasking view, as windows are laid out in a clean geometrical space (directional focus when you have overlapping windows is hard in terms of algorithm).

Anyway, it works on the simple use case, but it gets stuck, and I cannot select anything anymore if I switch the workspace.

### Steps to Reproduce

1. Open multitasking view from a workspace where you have multiple windows.
2. Select a window by doing `right`/`left`/`up`/`down`, and you will see a border.
3. Still on the multitasking view, swap to another workspace.
4. Come back to the workspace where you had the window selected.
5. You are stuck, `right`/`left`/`up`/`down` don't work anymore.

### Expected Behavior

You should be able to continue selecting the window.

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

Secure Session (Wayland)

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

leolost2605:
Instead of manually handling keyboard focus where needed, do it like GTK does it with some adaptations.

We have the FocusController which is added to the "root" focusable of a focus tree (e.g. mtv or window overview).

Then we have the focusable class which is subclassed by the actors in the actor tree where focus is needed. The focusable handles deciding which implementor gets the focus based on the direction.

We now also hide the focus indicator after 5 seconds.

Fixes #2651 
Fixes #2648 

We can easily add navigation via Tab now too but I would leave that to a follow up


danirabbit:
@lenemter could you give this a review? :)

lib/Gestures/ActorTarget.vala
```diff
@@ -10,7 +10,7 @@
  * It will propagate gesture events to all direct descendants that are also {@link ActorTarget}s.
  * If a new child (or target via {@link add_target}) is added, its progress will be synced.
  */
-public class Gala.ActorTarget : Clutter.Actor, GestureTarget {
```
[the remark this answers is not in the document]
leolost2605: > ActorTarget (which is supposed to handle gestures) unexpectedly handles focus state now

Well it doesn't handle focus state but it derives from a class that does.

I get what you're saying but we kind of really need a base class otherwise everything will just be a giant hack (at least what I can think of, like I said I'm still very much open to other ideas). And I'm not sure anyone does it differently? Both GNOME Shell and GTK have focus management in the widget base class and do it similar to this PR.
Once again what do you think about introducing a widget? And then renaming `ActorTarget` to `WidgetTarget`? This would solve overlooking actortarget extending focusable.
This way widgets also explicitly use focusable because they are using a baseclass that's called **Widget**(Target) and widgets bring focus management.

leolost2605:
@lenemter I've added two commits on top that rename focusable to widget and then actortarget to widget target. Let me know what you think!

lenemter:
@leolost2605 I think it would be better to leave ActorTarget as is, rename Focusable to Widget, and make Widget subclass ActorTarget. This way Widget uses gesture infrastructure, not the other way around. What do you think?

leolost2605:
@lenemter sure we can do that, I updated it.
Seems like something broke active shape correctly displaying (on fractional scaling?) but it's also on main and I'll open an issue for that #2756 
Screenshot from 2026-02-07 13 45 52@1x

lenemter (changes requested):
For some reason when I hold down left/right arrows in with window focused and then release it, the left/right continues to be repeated

leolost2605:
> For some reason when I hold down left/right arrows in with window focused and then release it, the left/right continues to be repeated

@lenemter Hmm I can't reproduce this but maybe I misunderstood. What I did is with two windows open in the multitaskingview hold arrow key right -> the right most window gets selected -> the bell plays on repeat because the key is held but we can't navigate further to the right -> as soon as I release the key the bell stops

> We should check if actor is Widget and use property to store the data, since we don't use qdata anywhere else

Now that we have the widget i've even been thinking whether we should drop focuscontroller and introduce a `Root` subclass of widget which would handle the focus controller stuff and would then have the focus_visible property. That's how GTK does it with `Gtk.Window`. What do you think?

leolost2605:
@lenemter I updated it with a `Root` now

lib/Widgets/Widget.vala
```diff
@@ -0,0 +1,107 @@
...
+    public bool can_focus { get; set; default = false; }
+    public bool has_visible_focus { get; private set; default = false; }
+
+    construct {
+        key_focus_in.connect (focus_changed);
+        key_focus_out.connect (focus_changed);
+    }
+
+    internal void focus_changed () {
+        has_visible_focus = has_key_focus () && get_root ().focus_visible;
```
lenemter: We need to check that `get_root ()` didn't return null

lib/Widgets/Widget.vala
```diff
@@ -0,0 +1,107 @@
...
+ */
+
+public class Gala.Widget : ActorTarget {
+    public bool can_focus { get; set; default = false; }
+    public bool has_visible_focus { get; private set; default = false; }
+
+    construct {
+        key_focus_in.connect (focus_changed);
+        key_focus_out.connect (focus_changed);
+    }
```
lenemter, line 15: Why not use something like this?

```suggestion
    public override void key_focus_in () {
        focus_changed ();
    }

    public override void key_focus_out () {
        focus_changed ();
    }
```

lib/Widgets/Widget.vala
```diff
@@ -0,0 +1,107 @@
...
+            if (actor is Root) {
+                return (Root) actor;
+            }
+        }
+
+        return null;
+    }
+
+    public bool focus (FocusDirection direction) {
+        var focus_actor = get_stage ().get_key_focus ();
```
lenemter, line 32:
```suggestion
        unowned var focus_actor = get_stage ().get_key_focus ();
```

lib/Widgets/Widget.vala
```diff
@@ -0,0 +1,107 @@
...
+            }
+
+            return move_focus (direction);
+        } else {
+            if (move_focus (direction)) {
+                return true;
+            }
+
+            return grab_focus ();
+        }
```
lenemter, line 70:
```suggestion
        }

        if (move_focus (direction)) {
            return true;
        }

        return grab_focus ();
```

lib/Widgets/FocusUtils.vala
```diff
@@ -0,0 +1,83 @@
...
+                direction == RIGHT && rect.x >= focus_rect.x + focus_rect.width
+            );
+        }));
+
+        children.retain_all (to_retain);
+    }
+
+    private inline Mtk.Rectangle get_allocation_rect (Clutter.Actor actor) {
+        return {(int) actor.x, (int) actor.y, (int) actor.width, (int) actor.height};
+    }
```
lenemter: Why don't you use `Clutter.Actor.allocation`? It uses Clutter.ActorBox which has x1, x2, y1, y2. That should simplify the code in the `filter_children_for_direction`

lenemter (changes requested):

lib/Widgets/Widget.vala
leolost2605: Any reason to prefer that? It's a lot more bloat and I think connecting to the signals is quite understandable

lib/Widgets/Widget.vala
leolost2605: This might be a bit far fetched but I think theoretically if the focus actor for some reason handles loosing focus by destroying itself and therefore finalizing we might have a dangling pointer here AFAICT. This wouldn't cause a problem right now because we don't use it after we call `move_focus` but there's no guarantee that we won't use it in the future.

lib/Widgets/Widget.vala
leolost2605: IMO in this case since both branches are "equal" and not just a precondition check keeping the else makes it a bit more understandable.

lib/Widgets/FocusUtils.vala
leolost2605: Thank you I didn't even think of that :+1:

lenemter (approved):

PR #2537, opened 2025-09-17
WindowClone: Use a clone as drag handle

leolost2605:


leolost2605:
Should be fixed now

lenemter (changes requested):
I get new warning when the window returns to its place after drag

```
Sep 21 10:46:30 HYM-W56 gala[52262]: invalid (NULL) pointer instance
Sep 21 10:46:30 HYM-W56 gala[52262]: g_signal_handler_disconnect: assertion 'G_TYPE_CHECK_INSTANCE (instance)' failed
```

leolost2605:
Ah yeah that's fixed now too :grimacing:

PR #2538, opened 2025-09-17
HideTracker: add pressure reveal

lenemter:
Fixes https://github.com/elementary/dock/issues/414

### DIFF
--- a/src/HotCorners/Barrier.vala
+++ b/src/HotCorners/Barrier.vala
@@ -11,9 +11,9 @@ public class Gala.Barrier : Object {
 
     public bool triggered { get; set; default = false; }
 
-    public int trigger_pressure_threshold { get; construct; }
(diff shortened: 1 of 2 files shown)


danirabbit:
Oh well hey that's quite simple then haha. Plank's value was 60 here, which personally feels way too hefty. But I think 2.5 is pretty light, was there somewhere specific that value comes from? What do you think about like 16? That should give plenty of room to overshoot if you have tool buttons at the bottom of the window, but I'm still able to reliably reveal the dock when I'm meaning to

lenemter:
@danirabbit 60 from plank and 2.5 mean different things. I think 60 in plank means that after colliding with the screen edge the cursor would need to travel additional 60 pixels *over some time*, while here 2.5 means that the cursor needs to travel 2.5 pixels *in one event*

PR #2539, opened 2025-09-23
Use GLib.Once for singletons

lenemter:
... and remove redundant constructor of BackgroundCache


lenemter:
@leolost2605 Didn't know about this syntax. I changed other instances of GLib.Once to use it too.

tintou:
Note that ideally, we should have no singleton, everything should have a parent object

PR #2546, opened 2025-09-25
gtk3-daemon: don't depend on libgala

lenemter:


tintou:
I'd personally create a separate .vala file that contains this and just include it in the daemon so we're sure that it's always synced

tintou:
Something like https://github.com/elementary/gala/pull/2549 what do you think?

PR #2549, opened 2025-09-26
gtk3/4-daemon: don't depend on libgala

tintou:
Move the enums to a shared file so that we can only keep the common part apart.


tintou:
Done!

Issue #2452, opened 2025-06-23
Create a test plan


tintou:
It is really easy to introduce regressions, it would be great to have a clear test-plan to make sure that all the expectations are covered

leolost2605:
Not a test plan but I'll lay something out to get the conversation going. Feedback, ideas, suggestions and everything else is very much appreciated since I have pretty much zero experience with testing in vala/glib.

Start with some easy objectives (I've got concrete ideas here and will propose a PR soon):
- Introduce a testcase base class adapted from the [libgee testcase class ](https://gitlab.gnome.org/GNOME/libgee/-/blob/d2e135e98718e37cae04599689fea8fb145b0b74/tests/testcase.vala)
  - In the future we could put it into a framework for use elsewhere (e.g. appcenter) but for now I think it would be better to use it for a bit to see limitations, etc.
- Use it to test GestureController and in general the Gesture infrastructure
  - GestureController is easily unit testable since it doesn't really require anything from mutter and we can easily add a mock backend
  - GestureController has a lot of edge cases I came across which makes it a good starting point for writing a bunch of tests

Next up: Identify what could be unit tested maybe by priority
- Can we subclass Meta.Window for a MockWindow?
- We can probably test a few things in utils

System tests:
- Like GNOME Shell does using a [helper](https://github.com/GNOME/gnome-shell/blob/baeeb2ba89997c93d9fab04daa2e25b084b72ebe/js/ui/scripting.js) that can spawn and wait for windows. They currently only use it for performance testing but explicitly state it could be used for correctness testing as well.
- A concrete area of note for example:
  - ShellClients and interaction with windows
    - Are they positioned correctly
    - Do they hide correctly
    - etc.

Ayoub-glitsh:
Hello ,
I’ve read through the issue and the discussion so far, and I think this is an important task to help prevent regressions.  
I’d like to work on this issue if it’s still open and contribute to defining or improving the test plan.
Please let me know if that’s okay and how you’d prefer me to proceed.
Thank you!

leolost2605:
Hi @Ayoub-glitsh thanks for your interest!
Usually we work on what's most important to us so if there is something specific you'd like to do or that you're interested in feel free to ask questions or open a PR and we can work on it :)
Testcases or even just identifying parts of the code base that could be unit tested are always appreciated!
For getting started I think in `Utils.vala` are some simple methods that could be tested if you want to get to know the testing infrastructure a bit.
[...]

PR #2550, opened 2025-09-27
Add a simple GestureController test

leolost2605:
Starting work on #2452
Required minor changes on gesture controller (allowing to enable a custom backend)

We have a testcase class that should be subclassed by individual tests.

Each test then should provide a main function that will run its testcase.

Lmk what you think, any feedback, etc. is very much appreciated :)

The tests are currently only run for the "build" CI should they be run for the others too?

I will add more tests for edge cases for the gesture controller in follow ups but I just wanted to see whether this is going in the right direction :)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,9 +36,12 @@ jobs:
       env:
         DESTDIR: out
       run: |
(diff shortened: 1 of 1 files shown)


leolost2605:
Superseded by #2565

PR #2522, opened 2025-08-06
WindowClone: display modal dialogs on top of parent window

lenemter:
Closes https://github.com/elementary/gala/issues/198

# Before
Screenshot from 2025-08-06 21 14 57

# After
Screenshot from 2025-08-06 21 13 56

PR #2561, opened 2025-10-09
Introduce a WindowListModel

leolost2605:
Closes #2353 

Introduce a window list model that allows us to collect the filtering and sorting code for different views of the current window list in a single place

Since this window list model now also guarantees that all windows that appear in it have a valid actor i.e. get_compositor_private won't return null this will allow for a bunch of clean up in a follow up :)

This should also help quite a bit with #2522 to allow more deduplication there


leolost2605:
This should be ready for review now and thorough testing would be appreciated :)

PR #2565, opened 2025-10-10
Introduce a first simple test

leolost2605:
Requires #2564 

Starting work on https://github.com/elementary/gala/issues/2452

We have a testcase class that should be subclassed by individual tests.

Each test then should provide a main function that will run its testcase.

Lmk what you think, any feedback, etc. is very much appreciated :)

The tests are currently only run for the "build" CI should they be run for the others too?


leolost2605:
@danirabbit since the build job now also runs the tests I renamed it to build-and-test. Would you prefer changing it back to build or can you update repository rules to require `build-and-test` instead of `build`?

leolost2605:
@danirabbit slight reping :)

danirabbit:
Oops thanks for pinging me again. I can update the rules 🚀

Ayoub-glitsh:
Hello 👋  
I noticed that PR #2565 introduced the initial test infrastructure and a first simple test, which is a great step forward for this issue.
If this issue is still open, I’d be happy to help continue the work for example by contributing to a more complete test plan, adding additional test cases, or improving test-related documentation.
Please let me know how you’d like me to proceed.
Thanks!

PR #2573, opened 2025-10-11
Introduce PropertyAnimator

lenemter:
Converts some window manager effect to use PropertyAnimator, which can handle disabled animations correctly by itself.

### DIFF
--- a/lib/Gestures/ActorTarget.vala
+++ b/lib/Gestures/ActorTarget.vala
@@ -84,6 +84,7 @@ public class Gala.ActorTarget : Clutter.Actor, GestureTarget {
         switch (update_type) {
             case START:
                 ongoing_animations++;
+                notify_property ("animating");
                 start_progress (action);
                 break;
             case UPDATE:
@@ -94,6 +95,7 @@ public class Gala.ActorTarget : Clutter.Actor, GestureTarget {
                 break;
             case END:
                 ongoing_animations--;
+                notify_property ("animating");
                 end_progress (action);
                 break;
         }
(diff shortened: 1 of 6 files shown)


leolost2605 (commented):
I'm wondering whether it would be better to just have two utils methods something like this
```vala
Utils.prepare_animation (actor, time, mode);
// set actor properties 
actor.translation_x = 5;
Utils.run_animation (actor, () => { 
    // do some stuff after the animation finished
});
```
The utils methods would then just save and restore easing state depending on whether animations are enabled and the `run_animation` would then immediately call the callback or connect to the finished signal of an animation and then call the callback.
This would be less code and you can't accidentally write a property name wrong or have a type wrong.
This was just an idea and I'm open to other arguments though so lmk what you think 🤷

lenemter:
@leolost2605 I don't like this approach. The main downside of what you suggested is that there's no way to get the animated properties, so you have to use Clutter.Actor.transitions_completed and it can lead to some issues -- if you animate the actor two times, the on_stopped handlers might get fired at incorrect timing.

I don't really like passing properties around as strings, but I don't think there's a way around that unfortunately =( For now I added a warning when property name is incorrect. I'll see if I can find more robust way to pass properties (GLib.ParamSpec maybe?)

leolost2605:
You could do some black magic with the notify signal but yeah I agree that's a bit unfortunate. Still I'm not a fan of the API it just feels a bit clunky having a constructor with so many side effects and just spawning an object like this isn't really a thing you see often. It's also very limiting, e.g. if we want to add more ways how to specify what to animate. Maybe something similar to the builder pattern would work? Aka
```vala
var builder = new AnimationBuilder (actor, duration, easing_mode);
builder.add_property (name, to_val);
builder.add_property (name, to_val);
builder.add_property_with_from (name, from_val, to_val);
// Would also allow us to add more convenience APIs as we need them or more advanced api e.g.
builder.add_transition (transition);
builder.run (() => print ("finished"));
```
Just an idea not sure about all of the names etc.

PR #2576, opened 2025-10-13
Enable backend

leolost2605:
Allow to enable custom backends. This is especially useful for automated testing to mock backends

For this we also need to not treat the scroll backend special so we introduce a custom hint to gestures which will make them not require a matching configured action. This will also come in handy with the pan backend.


leolost2605:
Superseded by #2757

PR #2582, opened 2025-10-18
lenemter/session start smooth

lenemter:


src/ShellClients/PanelWindow.vala
```diff
@@ -119,9 +121,15 @@ public class Gala.PanelWindow : ShellWindow, RootTarget {
...
+        starting = false;
+        workspace_hide_tracker.recalculate_all_workspaces ();
+    }
+
     protected override double get_hidden_progress () {
         if (starting) {
-            return user_gesture_controller.progress;
+            // Don't show when starting
+            return 1.0;
         }
```
leolost2605:
```suggestion
```
I would remove this all together then because IMO it's better to allow the user to immediately reveal it instead of trying that and then having it pop in when actually starting since we jump from 1.0 to 0.0

leolost2605 (changes requested):
Makes a lot more sense thank you!

src/ShellClients/PanelWindow.vala
lenemter: Makes sense, thanks for the suggestion!

leolost2605 (approved):

Issue #2587, opened 2025-10-20
Pointer bumps to the screen edge and travels sideways when moving between two displays


TomiOhl:
### What Happened?

The issue I'm experiencing did not happen before July.
When using two displays, if I try to move the pointer with the mouse/touchpad to the other display, the pointer bumps into the screen edge and continues sideways. Have to swipe really long for it to "break through" and get into the other display.
Using a touchpad, most of the time it means that I can only get to the second display with two swipes. If I'm dragging a window, that means the window get's maximized on first screen instead of being moved to the second.

### Steps to Reproduce

1. Have two displays
2. Try to move the pointer between the two displays

### Expected Behavior

There should be no bump if I'm not dragging a window. Even when I'm dragging a window, the bump should be less, like before

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

Secure Session (Wayland)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell

```

zeebok:
Do you still see this issue? I just tested on my daily build machine, and I don't see this happen. I wonder if the layout is related maybe?

PR #2588, opened 2025-10-21
Release barrier events when panel is fully visible

lenemter:
Addresses https://github.com/elementary/gala/issues/2587

I can't test this branch because I don't have a second monitor :(


danirabbit:
@elementary/project-managers can we find someone to test this fix?

zeebok:
>  can we find someone to test this fix?

I don't have this issue with my dual screen setup but I can at least test that this does not cause a breakage if that is good enough?

PR #2595, opened 2025-10-27
Bring back old icon groups

lenemter:
Bring back old icon groups and hide them behind a build flag. 

I made minimal adjustments to WorkspaceClone/IconGroup code to use WindowListModel. Everything else was just copied from the old codebase.

I made sure to not change anything that's not behind a build flag so we can cleanly remove them again when we'll release new workspace switcher.

I want to do this in order to finally release a new version

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -148,6 +148,10 @@ if get_option('systemd')
     vala_flags += ['--define', 'WITH_SYSTEMD']
 endif
 
+if get_option('old-icon-groups')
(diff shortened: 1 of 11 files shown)


danirabbit (approved):
Works as expected. Also doesn't crash with dock daily which is nice because then we don't have to worry about making sure to release at the same time. Nice work!

danirabbit:
I pushed a daily packaging branch that sets this false and made sure to set it true for regular packaging and update recipes etc. So should be good to go there

PR #2600, opened 2025-10-28
MultitaskingView: Add overshoot animation

### LINKED ISSUE #1026: 3/4 finger over-swiping down on the desktop should zoom the wallpaper a bit

lenemter:
Fixes https://github.com/elementary/gala/issues/1026
Closes https://github.com/elementary/gala/pull/2496

### DIFF
--- a/lib/Gestures/PropertyTarget.vala
+++ b/lib/Gestures/PropertyTarget.vala
@@ -10,11 +10,24 @@ public class Gala.PropertyTarget : Object, GestureTarget {
     // Don't take a reference since we are most of the time owned by the target
     public weak Object? target { get; private set; }
     public string property { get; construct; }
-
+    public bool allow_overshoot { get; construct; }
     public Clutter.Interval interval { get; construct; }
 
-    public PropertyTarget (GestureAction action, Object target, string property, Type value_type, Value from_value, Value to_value) {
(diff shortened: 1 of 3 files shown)


leolost2605 (commented):
Maybe we should move the allow overshoot to the controller and then clamp there to the actual clamp values we set?

lenemter:
@leolost2605 I don't think that would work. We need to allow some animations to overshoot while not the others. E.g. background can overshoot, while window clones cannot

leolost2605:
Idk IMO it would be better if everything that a gesture would move if done in the correct direction, would also overshoot:

https://github.com/user-attachments/assets/1fb96715-6d65-4794-a552-0026c244b858

Especially in this case I think it's important that the window clones move when trying to open the multitaskingview in the wrong direction. If only the background moves it's very easy to miss if one has a somewhat big window open let alone a maximized one.

danirabbit:
Gonna bump this to the next release since it has conflicts and there's some requested changes

PR #2601, opened 2025-10-28
CloseButton: make monitor scale settable

lenemter:
Cherry-picked from https://github.com/elementary/gala/pull/2356


plugins/pip/PopupWindow.vala
```diff
@@ -97,7 +97,7 @@ public class Gala.Plugins.PIP.PopupWindow : Clutter.Actor {
 
         set_position (x_position, y_position);
 
-        close_button = new Gala.CloseButton (scale) {
+        close_button = new Gala.CloseButton () {
```
leolost2605: Don't we have to set the scale here somewhere? Or is it always guaranteed to be one?

leolost2605 (approved):
LGTM!

Issue #335, opened 2018-09-17
Close button isn't always on top


danrabbit:
![screenshot from 2018-09-17 15 00 38 2x](https://user-images.githubusercontent.com/7277719/45652773-8ab26380-ba8a-11e8-8e81-f0d3481fa3ca.png)

The close button for windows in multitasking view should always be on top. Sometimes it shows below other windows

PR #2602, opened 2025-10-28
WindowCloneContainer: use close button size for gaps size

lenemter:
Fixes https://github.com/elementary/gala/issues/335


leolost2605 (approved):
Makes sense and great that we now have a constant for the button size 👍

PR #2606, opened 2025-10-29
WindowSwitcher: correctly scale icons and rounded corners

lenemter:


leolost2605 (changes requested):

lenemter:
@leolost2605 Oooh you're right. WindowSwitcher code is a mess and needs some refactoring but I hope that would be enough for now.

PR #2609, opened 2025-10-29
DBus: cleanup

lenemter:
1. Create `print_warning` method to simplify handling name loss
2. Use shorter enums
3. Register objects after we own the name, not when we connect to dbus in general

### DIFF
--- a/lib/Gestures/ActorTarget.vala
+++ b/lib/Gestures/ActorTarget.vala
@@ -84,6 +84,7 @@ public class Gala.ActorTarget : Clutter.Actor, GestureTarget {
         switch (update_type) {
             case START:
                 ongoing_animations++;
+                notify_property ("animating");
                 start_progress (action);
                 break;
             case UPDATE:
@@ -94,6 +95,7 @@ public class Gala.ActorTarget : Clutter.Actor, GestureTarget {
                 break;
             case END:
                 ongoing_animations--;
+                notify_property ("animating");
                 end_progress (action);
                 break;
         }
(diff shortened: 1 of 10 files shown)


src/DBus.vala
```diff
@@ -14,67 +14,75 @@ public class Gala.DBus {
...
+                }
             },
-            () => {},
-            () => warning ("Could not acquire name\n") );
+            print_warning
+        );
 
-        Bus.own_name (BusType.SESSION, "org.gnome.Shell", BusNameOwnerFlags.NONE,
-            (connection) => {
+        Bus.own_name (SESSION, "org.gnome.Shell", NONE, null,
```
leolost2605: and here

src/DBus.vala
```diff
@@ -14,67 +14,75 @@ public class Gala.DBus {
...
-            () => {},
-            () => critical ("Could not acquire name") );
+            print_warning
+        );
 
-        Bus.own_name (BusType.SESSION, "org.gnome.Shell.Screenshot", BusNameOwnerFlags.REPLACE,
-            () => {},
-            () => {},
-            () => critical ("Could not acquire name") );
+        Bus.own_name (SESSION, "org.gnome.Shell.Screenshot", REPLACE,
```
leolost2605: and here

src/DBus.vala
```diff
@@ -14,67 +14,75 @@ public class Gala.DBus {
...
+                    warning (e.message);
+                }
             },
-            () => {},
-            () => critical ("Could not acquire ScreenSaver bus") );
+            print_warning
+        );
+    }
+
+    private static void print_warning (GLib.DBusConnection connection, string name) {
```
leolost2605: I would rename this to something like `on_name_lost`

PR #2610, opened 2025-10-30
WindowClone: Fix some warnings on allocate

leolost2605:
- outer_rect.width can be 0 in which case the clone_scale_factor is nan which leads to us allocating a size of nan. Therefore set the scale to 1 if the width is 0.
- clone isn't a child of us so don't allocate it
- bound the window title max width by its min width. The calculated max width could even be less than 0 which lead to allocating a negative size.


lenemter (approved):
LGTM

Issue #2552, opened 2025-09-28
Windows placed in undisplayed area


ProgramCrafter:
### What Happened?

In multi-screen setup shown below, any opened windows are placed in non-displayed area of screen space.

### Steps to Reproduce

1. Have two monitors in the following configuration.

<img width="833" height="627" alt="Image" src="https://github.com/user-attachments/assets/b6dcefa5-e7c8-4c9f-a698-6ba9e96cfd24" />

(Check that Wingpanel and dock are displayed on the one to the right.)

2. Try launching any app. See its indicator, but no window at all.
3. Possible ways of follow-up interaction:
    - Super+Up: the window is brought in and maximized; it is also possible to minimize it back to nowhere
    - Alt+F7: the window is brought in to the mouse pointer and can be dragged freely onwards
    - PrtScr: the window is rendered in the top left area, the one without associated displays

### Expected Behavior

Windows do not enter undisplayed area on launch.

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

Secure Session (Wayland)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell

```

macmpi:
Even when external screen is disconnected, windows keeps opening in that "external" (and missing) display.
Mouse can be lost in that space too.
Suspend/wake does not re-sync available monitors. Only rebooting gets available monitors synced.

This is very inconvenient in common office usecase where a laptop may be often connected/disconnected from external monitor (desk/meetings transitions): rebooting at each transition is not acceptable.

Thanks for consideration.

ProgramCrafter:
> Even when external screen is disconnected, windows keeps opening in that "external" (and missing) display.

Did not happen for me in all the few dozens of times, though I am turning the display off prior to disconnecting the cable.

macmpi:
> though I am turning the display off prior to disconnecting the cable.

I'm not indeed...quite common practice though (albeit dirty).

Issue #2526, opened 2025-08-11
Windows positoined outside the viewport


WebShapedBiz:
### What Happened?

Hi.
On Wayland, after the latest update, v8.2.5, I have an issue where on the app's first run, the window gets positioned below the viewpoint and pushed to the right just so that the window control buttons are not visible.
Also, the window is positioned cca 3 wingpanel heights below the wingpanel and cannot be dragged to the top. I can maximise and restore the window, which puts the window on top, but then the bottom part of the window is pushed below the viewport.

### Steps to Reproduce

1. After the login, start, for example, Chrome (deb installation) and "enjoy"

### Expected Behavior

1. App's window shouldn't be positioned outside the viewport
2. App's window should be draggable to the top of the screen, not stuck cca 3 wingpanel heights below

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

Secure Session (Wayland)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell

```

PR #2611, opened 2025-10-30
WindowStateSaver: validate saved window position before placing

lenemter:
Fixes https://github.com/elementary/gala/issues/2552
Fixes https://github.com/elementary/gala/issues/2526

### DIFF
--- a/data/gala.metainfo.xml.in
+++ b/data/gala.metainfo.xml.in
@@ -37,6 +37,8 @@
         </ul>
       </description>
       <issues>
+        <issue url="https://github.com/elementary/gala/issues/2526">Windows positoined outside the viewport</issue>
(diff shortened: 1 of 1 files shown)


src/WindowStateSaver.vala
```diff
@@ -125,7 +127,10 @@ public class Gala.WindowStateSaver : GLib.Object {
                 }
             }
 
-            window.move_resize_frame (false, last_x, last_y, last_width, last_height);
```
leolost2605: Yeah I was wondering, shouldn't move_resize_frame with user op `false` constrain windows to be in the display area? A comment might be good to explain what's different here vs. the default constraints

lenemter: Done!

PR #2621, opened 2025-11-06
Gestures: Keep smooth progress even with disabled animations

### LINKED ISSUE #2620: Reduce Motion breaks multitouch gestures: no live feedback, “teleporting” transitions
### What Happened?

With Reduce Motion enabled, multitouch gestures (e.g., workspace swipe) in Gala does not provide live, continuous feedback. Instead, the UI “teleports” to the final state after the gesture completes. This makes gestures feel broken and leads to mis-recognition (e.g., horizontal vs vertical). 

Per the guidance in Granite’s Stylesheet Targets, gestures should behave like drag handles and remain responsive even when animations are reduced.

Reference:

https://github.com/elementary/granite/wiki/Stylesheet-Targets#reduced-motion

> “Widgets should still react to multitouch as normal since they behave more like drag handles, but don't animate when interacted in other ways like keyboard shortcuts.”

To precisely detail what happens when Reduce Motion is `on`:

- While swiping on the touchpad, there’s no intermediate visual feedback; the interface doesn’t track finger position in real time.
- At the end of the gesture, the view jumps directly to the new state (no animation at all).
- Because there’s no visual tracking, gestures are easier to mis-trigger (e.g., the system sometimes interprets a left/right swipe instead of an up/down one).

### Steps to Reproduce

1. Enable Reduce Motion in System Settings.
2. On the desktop, perform a multitouch swipe (e.g., switch workspaces or open the overview via gesture).
3. Observe that there is no live movement during the swipe; only a sudden jump occurs at the end.

### Expected Behavior

Interface should continuously track finger position during the gesture (live feedback), while keyboard-triggered transitions can be reduced/removed.

This has been discussed with @danirabbit on: https://github.com/elementary/settings-desktop/issues/458

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

Secure Session (Wayland)

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

leolost2605:
Fixes #2620 

Keep smooth progress during the gesture but still skip the finishing animation. This is also how adwaita does it for e.g. navigation view, carousel, etc.

### DIFF
--- a/lib/Gestures/GestureController.vala
+++ b/lib/Gestures/GestureController.vala
@@ -64,7 +64,6 @@ public class Gala.GestureController : Object {
 
     /**
      * When disabled gesture progress will stay where the gesture ended and not snap to full integers values.
-     * This will also cause the controller to emit smooth progress information even if animations are disabled.
      */
     public bool snap { get; construct set; default = true; }
 
@@ -171,12 +170,6 @@ public class Gala.GestureController : Object {
                 !GestureSettings.is_natural_scroll_enabled (gesture.performed_on_device_type)
             ) {
                 direction_multiplier *= -1;
-            }
-
-            if (snap && !Meta.Prefs.get_gnome_animations ()) {
-                recognizing = false;
-                prepare ();
-                finish (0, progress + direction_multiplier);
             }
 
             recognizing_backend = backend;


lenemter (approved):
LGTM

lenemter:
I'm gonna go ahead and merge this since Danielle confirmed this is the desired behaviour

lobre:
> Keep smooth progress during the gesture but still skip the finishing animation.

I came here to report the fact that animations were not "complete" because as soon as the gesture is released, it "snaps" and skips the ending transition. For example, I swipe three fingers horizontally to transition to another desktop, and even if I don't complete my gesture until the end, as soon as I release, I am directly there on the next workspace, without the final smoothing of the transition.

And I see your comment, which now makes more sense.

I just wanted to know if that was normal, but I guess I have my answer. 

And just to say that the behavior was surprising at first, but I think I like the way it works in the end, as it is a good balance of reduced motions.

PR #2628, opened 2025-11-12
Improve drag and drop action event consuming

leolost2605:
From the commit message:
> We don't want to consume events let alone grab the actor
before we actually are sure that we will handle this drag.
This will improve situations where multiple gestures
can be started from the same position most notably
in the multitasking view drag and drop of window
clones and (once the new touchbackend is merged) swipe
left and right for switching workspaces.
This will also allow to use a separate click action instead
of having to implement clicking in the dnd action as well.

This will not cause any change in behavior right now because touch egg is grabbing events directly from the hardware but it will improve the behavior once our own touchscreen backend is merged and it will make things like one finger swipe up to close the window and still supporting drag and drop possible.

### DIFF
--- a/lib/DragDropAction.vala
+++ b/lib/DragDropAction.vala
@@ -201,7 +201,6 @@ namespace Gala {
                         return Clutter.EVENT_PROPAGATE;
                     }
 
-                    grab_actor (actor, event.get_device ());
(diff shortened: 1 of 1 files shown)


lenemter (approved):
LGTM

PR #2629, opened 2025-11-12
Leolost/touchscreen backend

leolost2605:


leolost2605:
I will start work on this again after OS 9 because mutter rewrote their gesture actions, dropping PanAction and introducing PanGesture. While they work in a very similar way the API is often just slightly different which would lead to many `#if MUTTER_49` so waiting for OS 9 is probably the better option here.

PR #2635, opened 2025-11-14
TouchpadBackend: Fix missing reset once gesture ends

### LINKED ISSUE #2633: Three finger swipes in different directions sometimes collide
### What Happened?

I have a setup like this:
- three finger swipe up for multitasking view
- three finger swipe sideways for app switcher
Swiping sideways via three fingers is sometimes registered as a swipe up, and instead of the app switcher I can see the start of the animation for multitasking view. If this happens, it will keep happening, so far I could get out of this state via two ways: using the keyboard shortcut for app switcher (even once) or swiping sideways very short swipes eventually actually triggers the app switcher.

### Steps to Reproduce

Have a setup like this:
- three finger swipe up for multitasking view
- three finger swipe sideways for app switcher
Use the device, switch windows, workspaces etc and use the gestures. I'm not aware of any specific scenario that could reliably trigger it

### Expected Behavior

Swiping sideways shouldn't be registered as swiping up

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

Secure Session (Wayland)

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

leolost2605:
The touchpad backend wouldn't reset when the gesture ends if it didn't handle it.

Fixes #2633

### DIFF
--- a/lib/Gestures/TouchpadBackend.vala
+++ b/lib/Gestures/TouchpadBackend.vala
@@ -56,11 +56,12 @@ private class Gala.TouchpadBackend : Object, GestureBackend {
             return Clutter.EVENT_PROPAGATE;
         }
 
-        if (state == IGNORED) {
-            if (event.get_gesture_phase () == END || event.get_gesture_phase () == CANCEL) {
-                reset ();
-            }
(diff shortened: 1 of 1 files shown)


lenemter (approved):
LGTM and makes the cycle windows gesture a lot more reliable

PR #2637, opened 2025-11-14
WindowSwitcher: Reduce gesture sensitivity

### LINKED ISSUE #2632: Windows switcher gesture overshoot
### What Happened?

Set up a three finger swipe for windows switcher.
In initial OS 8 I recall it worked from edge to edge, i.e. I could swipe from left to the right edge or backwards, but swiping too much to one side didn't result in selecting an app from the other side.
So if I had three apps, let's say Files, Terminal, Code, then swiping right I could select Terminal or Code, and when I reached past Code, the selection didn't jump to Files again, but "bumped into a wall". This was a lot easier to use, especially with only 2 apps opened, especially since the sensitivity of the gesture feels way too high and it's almost impossible to easily switch to the previous app.

### Steps to Reproduce

Set up the window switcher to a touchpad gesture and use it

### Expected Behavior

Swiping to the edge of the switcher shouldn't move the selection to the other side (probably using keyboard shortcuts it should still work like before).
And I'm not sure if there is something we could do with sensitivity/speed, it's possible it would be fine with the above change

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

Secure Session (Wayland)

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

leolost2605:
Fixes #2632 

Keep the wrap around but reduce sensitivity

### DIFF
--- a/src/Widgets/WindowSwitcher/WindowSwitcher.vala
+++ b/src/Widgets/WindowSwitcher/WindowSwitcher.vala
@@ -11,7 +11,7 @@ public class Gala.WindowSwitcher : CanvasActor, GestureTarget, RootTarget {
     public const int WRAPPER_PADDING = 12;
 
     private const int MIN_OFFSET = 64;
-    private const double GESTURE_STEP = 0.1;
+    private const double GESTURE_STEP = 0.2;
 
     public Clutter.Actor? actor { get { return this; } }
     public WindowManager wm { get; construct; }


lenemter:
0.25 feels a bit too much. 0.2 feels a bit easier to use

leolost2605:
Yeah `0.2` is fine with me :+1:

PR #2641, opened 2025-11-15
WindowClone: Fix has-pointer handling with reduced motion

lenemter:
For some reason window clone's has-pointer is bugged with disabled animations :(


leolost2605:
I'm usually more in favor of fixing causes rather than symptoms but in this case I'm also thinking if we should just have the close button visible all the time anyways? The style guidelines discourage using hover for actual functionality. Maybe @danirabbit has an opinion here?

danirabbit:
@leolost2605 I think previously we left this as a hover effect since we have swipe-to-close on touch. But I think it would be fine to always show close buttons :)

leolost2605:
@danirabbit yeah that was my first thought too but afaict we don't have swipe to close on touch?

danirabbit:
@leolost2605 oh you're right because that's waiting on the touchscreen backend right? Either way, I'm +1 on showing close buttons all the time

PR #2643, opened 2025-11-15
StyleManager: get data from settings portal

lenemter:
Not sure why this wasn't done in the first place but this significantly simplifies StyleManager


lenemter:
But we use Bus.watch_name so it should be fine ig

leolost2605:
Yeah I can't think of a problem caused by this rn but maybe @tintou can tell us more

PR #2649, opened 2025-11-17
SessionLocker: fix warning

lenemter:
I don't think it's possible to fix the warning directly, so I decided to add a debug


leolost2605 (approved):
Ig you could make the struct public since it is in theory public by being an external API but this LGTM too

PR #2655, opened 2025-11-20
Fix RoundedCornersEffect

lenemter:
Third time's a charm

Rename .vert shaders to .frag because well, these are fragment shaders, not vertex shaders.

Fixes the strange offsets in shader code by correctly calculating offsets using clutter_actor_box_enlarge_for_effects which is mutter private api. This works with and without fractional scaling which was the flaw of my previous attempts

### DIFF
--- a/src/Background/Background.vala
+++ b/src/Background/Background.vala
@@ -103,7 +103,7 @@ public class Gala.Background : Object {
 #endif
         if (color == null) {
 #if HAS_MUTTER47
-            color = Cogl.Color.from_string ("black");
+            color = Cogl.Color.from_string ("#000000");
 #else
             color = Clutter.Color.from_string ("black");
 #endif
@@ -122,7 +122,7 @@ public class Gala.Background : Object {
 #endif
             if (second_color == null) {
 #if HAS_MUTTER47
-                second_color = Cogl.Color.from_string ("black");
+                second_color = Cogl.Color.from_string ("#000000");
 #else
                 second_color = Clutter.Color.from_string ("black");
 #endif
--- a/src/Background/BackgroundContainer.vala
+++ b/src/Background/BackgroundContainer.vala
@@ -29,7 +29,7 @@ public class Gala.BackgroundContainer : Meta.BackgroundGroup {
         });
 
 #if HAS_MUTTER47
-        background_color = Cogl.Color.from_string ("Black");
+        background_color = Cogl.Color.from_string ("#000000");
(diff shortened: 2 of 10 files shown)


lib/RoundedCornersEffect.vala
```diff
@@ -59,18 +66,74 @@ public class Gala.RoundedCornersEffect : Clutter.ShaderEffect {
...
             actor.notify["height"].connect (update_actor_size);
 
+            update_clip_radius ();
             update_actor_size ();
         }
     }
 
+    private void update_clip_radius () requires (actor != null) {
+        var resource_scale = actor.get_resource_scale ();
+        set_uniform_value ("clip_radius", Utils.scale_to_int (clip_radius, monitor_scale / resource_scale));
```
leolost2605: Why do we divide by the resource scale here?

lenemter: With https://github.com/elementary/gala/pull/2664, and I checked that removing monitor_scale from Utils.scale_to_int (clip_radius, monitor_scale ...) produces incorrect corner radius too

leolost2605: Huh that's weird because I got this (which looks pretty right to me) with scale monitor framebuffer, 2x scale on this branch but with all of the get_resource_scale usage removed both in clip radius and actor size.
Screenshot from 2025-11-25 20 39 59
whereas with the exact same settings on this branch but without any modifications i.e. with dividing by resource scale i get this 
Screenshot from 2025-11-25 20 43 55

lenemter:
@leolost2605 Gtk4 tooltips seem to have the same issue, so I guess 4px is intended..?

image

leolost2605:
> @leolost2605 Gtk4 tooltips seem to have the same issue, so I guess 4px is intended..?

Ah all right that's good to know but IMO it looks better there because on both sides it is fully opaque whereas in the screenshot from this branch only one side is really opaque and the other quite transparent.

Also another thing I noticed is that there are different corners on the top and bottom (this is at 1.25x scaling):

image

I don't know, if you prefer we can merge this for now and then continue fine tuning later since it is very self contained. 
But can we maybe split out the file renaming to a separate commit/PR?

tintou (approved):

lenemter:
I think I understand where the issue is, but I'm too busy to look into it rn
I'm gonna merge this now since it's a lot better than in main and we can always iterate on it later

PR #2660, opened 2025-11-21
Enable 'scale-monitor-framebuffer' by default

leolost2605:
Ubuntu has been doing this for years now and I've been running it for over two years on elementary as well and it is a big improvement over text size scaling.
Afaict it doesn't have any performance implications or weird behavior at all nowadays and is pretty much production ready.


danirabbit (approved):
I've been testing this on my StarLite as well and there's a couple cases where scaling is odd but I think they're mostly fixed in main and it's much more usable. So I'm +1 here

PR #2662, opened 2025-11-21
Drop manually scaling UI elements

### LINKED ISSUE #2412: Multitasking View is oversized with fractional scaling

leolost2605:
When using 'scale monitor framebuffer' actor sizes etc. are already in logical sizes so we don't have to scale anything.

Fixes #2412 

This will cause the ui elements to be too small on
- X 
- wayland without scale-monitor-framebuffer

Regarding X that won't be a problem in the future but I'm not sure we should already introduce regressions for X before OS 9?
Regarding without scale monitor framebuffer I would argue we should treat not having scale monitor framebuffer as unsupported. Otherwise we would still have to keep all the code for a only rarely tested code path.

Depending on other opinions on already introducing regressions for X this is blocked until OS 9.

### DIFF
--- a/src/Widgets/MultitaskingView/MonitorClone.vala
+++ b/src/Widgets/MultitaskingView/MonitorClone.vala
@@ -16,7 +16,6 @@ public class Gala.MonitorClone : ActorTarget {
 
     public WindowManager wm { get; construct; }
     public int monitor { get; construct; }
-    public float monitor_scale { get; construct set; }
 
     private WindowCloneContainer window_container;
     private BackgroundManager background;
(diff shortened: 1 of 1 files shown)


danirabbit:
I think we should probably weight until OS 9 because we probably will have to drop the X session completely anyways. But at least for now let's not break it intentionally 😅

PR #2663, opened 2025-11-21
Fix blurry icons when running with fractional scaling

leolost2605:
We finally respect the resource scale to load the icons at a higher resolution when we can apply it.


leolost2605:
Superseded by #2665

PR #2673, opened 2025-11-29
Use gala icon for window icon

leolost2605:
Includes #2665 

Fixes blurry window icons when running fractional scaling


danirabbit (approved):
Looks great. Love to see some stuff out of utils. Nice job!

Issue #2676, opened 2025-12-03
Window selection from dock overview doesn't focus the chosen window


lobre:
### **Edit**

The issue is not strictly related to Reduce Motion being enabled, as I can reproduce it when disabled. But it is way easier to reproduce with Reduce Motion enabled. Like the problem happens every 2 clicks when enabled, while only "sometimes" when disabled.

### What Happened?

With "Reduce Motion" enabled in system settings, clicking an app in the dock that has multiple open windows shows the expected window overview, but selecting a window often doesn't focus it. The overview closes and I'm returned to whatever window I was previously on. Sometimes repeating the action a few times eventually works, but it's inconsistent.

I'm not sure if this belongs to the dock or to Gala, but I'm reporting it here first.

### Steps to Reproduce

1. Enable "Reduce Motion" in system settings.
2. Open Firefox in workspace 1.
3. Open two Terminal windows in workspace 2.
4. From either workspace, click the Terminal icon in the dock to show the window overview.
5. Click one of the Terminal windows. You will see that most times, it won't switch workspace and focus that window.

### Expected Behavior

The selected window should be focused immediately, regardless of the workspace, just like when animations are enabled.

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

Secure Session (Wayland)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell

```

lobre:
If I can add also, even when the focus ends up working, I still see a blink between overview and focused window instead of directly switching.

lobre:
Thanks for transferring this issue to gala. Effectively, I see I can reproduce simply by trying to click a window when opening the multitasking view. So nothing to do with the dock.

leolost2605:
I think this is the same issue that caused the hover widgets to sometimes not appear (see #2641) and for me it also happens without reduce motion. I did some investigation and seems like #2680 fixes it.

o-alquimista:
> Thanks for transferring this issue to gala. Effectively, I see I can reproduce simply by trying to click a window when opening the multitasking view. So nothing to do with the dock.

Is the "reduce motion" setting really needed to reproduce this?

I'm seeing a very similar problem in the multitasking view (reported as #2681), but I don't have that setting enabled.

lobre:
> Is the "reduce motion" setting really needed to reproduce this?

[...]

PR #2680, opened 2025-12-05
Gala.Text: Don't extend Clutter.Text

leolost2605:
This seems to mess with... something. At least the picking (or at least something with the event delivery) was wrong.
That was noticeable in the multitasking view where the window clones have tooltips. Sometimes these tooltips (that use gala text) caused the window clones to not receive events most notably enter and leave as well as button press events.

I think this is the same issues that caused the hover widgets to sometimes not appear (see #2641) at least this PR now also fixes the enter/leave events.

Fixes https://github.com/elementary/gala/issues/2676

### DIFF
--- a/lib/Text.vala
+++ b/lib/Text.vala
@@ -6,14 +6,34 @@
 /*
  * Clutter.Text that automatically changes font-name to the system one
  */
-public class Gala.Text : Clutter.Text {
+public class Gala.Text : Clutter.Actor {
     private static GLib.Settings gnome_interface_settings;
+
+#if HAS_MUTTER47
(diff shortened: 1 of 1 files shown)


lenemter (changes requested):
This breaks text in WindowSwitcher and WindowClone
Screenshot from 2025-12-27 07 37 12

leolost2605:
Ah oops I forgot to set the layout manager. Should be fixed now!

lenemter (approved):
LGTM

lenemter:
This breaks OpenSUSE build. But i guess it's fine since new OpenSUSE 16 now which provides mutter 49 (I probably need to update the CI...)

PR #2704, opened 2025-12-27
Fix Wingpanel dialogs dock and workspace behavior

### LINKED ISSUE #645: Wingpanel dialogs are shown as an application on the dock Wayland
### What Happened?

When a wingpanel dialog is shown for example the shutdown dialog, if the selected session is the secure session (Wayland) it will show on the dock as "Wingpanel".

https://github.com/user-attachments/assets/e92b5b52-1758-4444-b398-bf820cc31ed9

### Steps to Reproduce

1. Be on Elementary OS 8 with all updates installed.
2. Be on the Secure Session (Wayland)
3. Open a Wingpanel dialog. Example: Trying to shutdown the computer via the quick settings will open a dialog.

### Expected Behavior

The dialog should not be shown as an application on the dock just like how it does on the Classic Session (X11)

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

Secure Session (Wayland)

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

lenemter:
Fixes https://github.com/elementary/wingpanel/issues/645

### DIFF
--- a/lib/Utils.vala
+++ b/lib/Utils.vala
@@ -338,6 +338,17 @@ namespace Gala {
             }
         }
 
+        public static inline bool get_window_and_ancestors_normal (Meta.Window window) {
+            var normal = true;
+            window.foreach_ancestor ((ancestor_window) => {
+                normal = normal || get_window_is_normal (ancestor_window);
+
+                return normal;
+            });
+
+            return normal;
+        }
+
         public static int calculate_button_size (float monitor_scale) {
             return Utils.scale_to_int (BUTTON_SIZE, monitor_scale);
         }
--- a/src/DesktopIntegration.vala
+++ b/src/DesktopIntegration.vala
@@ -74,19 +74,7 @@ public class Gala.DesktopIntegration : GLib.Object {
     }
 
     private bool is_eligible_window (Meta.Window window) {
-        if (window.is_override_redirect ()) {
-            return false;
-        }
-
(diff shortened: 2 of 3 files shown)


leolost2605:
Wouldn't `get_window_is_normal (window) && !ShellClientsManager.get_default ().is_positioned_window (window)` also work? That would also prevent things like the polkit dialog to show up.

lenemter:
@leolost2605 That would prevent onboarding showing up in the dock (which should not on top but that's another issue). I don't have a strong opinion here and both solutions are fine for me

leolost2605:
@lenemter that was fixed in #2709

lenemter:
@leolost2605 TBH the reason why I implemented this way is not to rely on ShellClientsManager. I want to use window properties as much as possible instead of relying on pantheon shell protocol

leolost2605:
Any particular reason for that? Pantheon shell protocol and therefore the ShellClientsManager is now with wayland the only authority about what windows are shell windows so IMO it makes sense to only use that instead of a hint set by it which is kind of an implementation detail. Also in the future mutter wants to move away from window types so I think it makes sense we already go with our thing.
Also like I said this would also prevent things like the polkit dialog or end session dialog to show up which makes sense to me.

zeebok (approved):
Code seems decent, but there are conflicts

Issue #2675, opened 2025-12-02
Blur behind modal shell protocol


danirabbit:
### Problem

We have a very cool modal effect now for the end session dialog, for example, but it would be even cooler to blur behind

### Proposal

Blur everything behind the modal dim

PR #2713, opened 2026-01-05
Blur behind extended behavior protocol

lenemter:
Fixes https://github.com/elementary/gala/issues/2675


leolost2605:
Oh I understood #2675 to be about not blurring the window but the background.
@danirabbit can you clarify?

leolost2605 (approved):
I think this makes sense either way, LGTM

danirabbit:
Ah yeah I meant #2715 

I'm not certain I understand what this branch does actually?

PR #2715, opened 2026-01-05
ModalGroup: Blur background

### LINKED ISSUE #2675: Blur behind modal shell protocol

leolost2605:
This was my understanding of #2675 

The screen recording seems to be broken with background blur, in reality nothing is offset/zoomed in :shrug: 

https://github.com/user-attachments/assets/ec7f20ce-2dc3-43d7-83d4-2af833b42c84

Fixes #2675

### DIFF
--- a/src/Widgets/ModalGroup.vala
+++ b/src/Widgets/ModalGroup.vala
@@ -10,11 +10,16 @@
  * the desktop behind them and only allow interaction with them.
  * Not to be confused with WindowManager.push_modal which only
  * works for our own Clutter.Actors.
+ * Note that windows shouldn't be added to this actor directly but
+ * instead to {@link window_group}.
  */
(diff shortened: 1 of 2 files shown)


leolost2605:
@lenemter can I use one actor with the background blur effect that spans the whole area or do I need one per monitor?

leolost2605:
I'll have access to an external monitor later and I'd like to test with it regarding the background blur effect so I'll wait with merging

leolost2605:
It seems to work fine with multiple monitors both with scale monitor framebuffer enabled and disabled, so I'm kinda inclined to just merge as is and iterate later if there turns out to be a problem I missed :)

PR #2731, opened 2026-01-21
Simplify CI and convert tabs to spaces

lenemter:


danirabbit:
I think this was done intentionally so that generated vapis won't be linted, but @tintou would know for sure

Issue #2724, opened 2026-01-14
When using wl-copy/wl-paste (wl-clipboard) the dock shows up although Terminal is maximized


zinga666:
### What Happened?

Note: I don't know if this is the same issue as https://github.com/elementary/gala/issues/2630 and neither if it is an issue of gala or the dock but I guess it is some window focus/state problem (hence I put it here).

I use neovim in a maximized terminal window for editing code and the dock shows up and hides all the time although it is configured to "hide when the focused window overlaps the dock".

Usually this is not an issue and since it happens only when deleting/yanking/pasting text in neovim I figured it might have to do with `wl-copy`/`wl-paste` which is needed in neovim on Wayland to access the system clipboard.

And indeed I could reproduce the issue just using `wl-copy`/`wl-paste` in a maximized terminal window.

### Steps to Reproduce

1. `sudo apt install wl-clipboard` (provides `wl-copy`/`wl-paste`).
2. Configure the dock to "hide when the focused window overlaps the dock".
3. Open terminal window and maximize it.
4. In the terminal execute `wl-copy hello dock`.
5. In the terminal execute `wl-paste` several times.

At some point the dock shows up although it shouldn't and does not hide again. The terminal window is still focused though and executing another `wl-paste` hides it.

Hovering the dock with the mouse does not hide the dock.

<img width="1920" height="1200" alt="Image" src="https://github.com/user-attachments/assets/9a3e4c66-08b7-4877-aab8-b1afc6bdf5eb" />

### Expected Behavior

The dock should stay hidden.

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

Secure Session (Wayland, This is the default)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell

```

PR #2735, opened 2026-01-21
Fix wl copy paste

lenemter:
Fixes https://github.com/elementary/gala/issues/2724

Not the best solution, but I couldn't find anything better

### DIFF
--- a/data/gala.metainfo.xml.in
+++ b/data/gala.metainfo.xml.in
@@ -36,6 +36,7 @@
         </ul>
       </description>
       <issues>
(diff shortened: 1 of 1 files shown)


leolost2605 (approved):
Yeah I don't love it either but LGTM

leolost2605:
Hmm one thing though: Might there be cases where a window might not know it's size (or for some reason doesn't provide it) at the beginning and spawn with 1x1 and then resize to the proper size? Because get_window_is_normal isn't really expected to return something different for the same window so this would lead to a bunch of problems...

Issue #2746, opened 2026-01-28
6th and 13th key press of Alt+Tab have a bug


zinga666:
### What Happened?

When using Alt+Tab to switch between windows there are two issues:
- the 6th Tab key press is skipped, i.e. doesn't move to the next entry
- the 13th Tab key press skips one entry

### Steps to Reproduce

1. Open a few windows, e.g. Files, Terminal, System Settings, Code, etc.
2. Press Alt (and keep it pressed).
3. Press Tab 7 times (don't release Alt) -> the 6th key press doesn't move to the next entry
4. Press Tab another 6 times -> the 13th key press skips one entry in the list

### Expected Behavior

Every press of the Tab key should move one entry forward in the list.

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

Secure Session (Wayland, This is the default)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell

```

PR #2750, opened 2026-02-06
Fix rounding error in WindowSwitcher

lenemter:
Fixes https://github.com/elementary/gala/issues/2746

Casting to int rounds the number down, and for some reason 6.0 was rounded to 5... Floating number magic


leolost2605 (changes requested):
Makes sense only one comment

PR #2755, opened 2026-02-07
Cleanup gesture code

lenemter:
Fixes some warnings


leolost2605 (approved):
We're gonna need quite a bit of that again in the future but let's just reintroduce it then

PR #2760, opened 2026-02-10
Tests: Introduce a MutterTestCase

leolost2605:
We need this for any tests that need a clutter.actor because to construct one it needs a context setup.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,9 +39,15 @@ jobs:
         meson build -Ddocumentation=true -Dtests=true
         ninja -C build
         ninja -C build install
-    - name: Run Tests
+    - name: Prepare Test Environment
       run: |
-        meson test -v -C build --no-suite Gala/mutter
+        mkdir -p -m 700 ${{ github.workspace }}/runtime-dir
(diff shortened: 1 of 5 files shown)


leolost2605:
Unfortunately we need quite a setup in CI to run tests requiring a mutter context which I don't have the knowledge to do right now. However we can still run the tests locally so ig for now that might be better than nothing? I introduced a new "mutter" suite (any better name suggestion?) for the tests that need the mutter context to be set up (any tests that want a clutter actor) and exclude it from CI.

The main blockers I found for CI:
- we need a user runtime dir
- we need a certain dbus interface to be available but I forgot what it was
- probably more (the gnome shell CI setup file is really long)

If someone has ideas how to get that up and running I'd be happy to hear them :)

leolost2605:
Note: To run the tests locally you have to be a user (no sudo) and there can't be a mutter instance already running on the same dbus. So you can either run it from the tty or via `dbus-run-session meson test`

leolost2605:
I managed to get the CI working so this should be ready now 🎉

zeebok (approved):
Seems to work locally! One thing that might be good to do in a future PR if applicable is update the README to explain how to run the tests locally

PR #2770, opened 2026-02-15
Add CRT effect shader and animation while shutting down

SubhadeepJasu:
Add a nice CRT TV-Off animation when shutdown is called.

https://github.com/user-attachments/assets/aede56d3-e63f-4b33-97a7-ce3440de7282

### DIFF
--- a/src/Main.vala
+++ b/src/Main.vala
@@ -93,6 +93,9 @@ namespace Gala {
 
         WindowStateSaver.on_shutdown ();
 
+        var shutdown_curtain = new Gala.ShutdownCurtain (ctx);
+        shutdown_curtain.animate ();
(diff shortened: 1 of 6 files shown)


src/Main.vala
```diff
@@ -93,6 +93,9 @@ namespace Gala {
 
         WindowStateSaver.on_shutdown ();
 
+        var shutdown_curtain = new Gala.ShutdownCurtain (ctx);
```
leolost2605: AFAIK at this point Mutter has already completely shutdown. So we don't have a stage nor a frame clock nor anything anymore

SubhadeepJasu:
Okay I will check that

PR #2773, opened 2026-02-16
Super+Tab: create a new workspace when there's just one

### LINKED ISSUE #1083: I can't switch to a new workspace freely
When I have just one workspace with windows, I'd expect to be able to switch to another one with START+TAB. I am able to switch to it with START+RIGHT_ARROW, so the inability to use the first shortcut is a bit weird.

lenemter:
Fixes https://github.com/elementary/gala/issues/1083


leolost2605:
IMO having a special case here could be confusing. I've got no strong opinion whether cycling should include the last workspace or not, as long as it's consistent but I think it should be consistent 🤷
But I'm not too good at ux so just leaving this as my opinion :)

PR #2777, opened 2026-02-20
CI: Update dev target to libmutter-17-dev

danirabbit:


danirabbit:
@tintou I tried elsewhere to add `libsystemd-dev` to the CI with no success. Is there another package that we need or something with the container itself?

garaevdi:
Shouldn't this update to `libmitter-18-dev` (gnome 50) or am I missing smth? Also, if I'm not mistaken, CI needs `systemd-dev` package.

danirabbit:
@garaevdi It looks like both 17 and 18 are available in resolute. But 18 won't build until we merge your branch :)

PR #2784, opened 2026-02-22
Move WorkspaceClone management to WorkspaceRow

leolost2605:
First handle size and monitor scale internally in the workspaceclone since we handle some monitor geometry here already anyways.

Then move the creation and deletion of workspace clones from the multitaskingview to the workspacerow. This allows for a bunch of cleanup.


zeebok (approved):
Quite the cleanup!  Looks good to me and does not seem to break anything in MT view for me

PR #2787, opened 2026-03-01
 MultitaskingView: Fix occassions where the active workspace could be wrong

### LINKED ISSUE #2701: Moving away the only window of a workspace makes it switch to an empty workspace
### What Happened?

https://github.com/user-attachments/assets/456ef8cb-bc2d-4afb-bf12-92dbe8212c65

### Steps to Reproduce

1. Open one app in workspace 1.
2. Open one app in workspace 2.
3. Switch to workspace 1 and enter multitasking view.
4. Drag and drop that app onto workspace 2.

### Expected Behavior

When workspace 1 is closed (for being empty), it should switch to workspace 2.

Currently, it switches to a new empty workspace, which I feel is unexpected.

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

leolost2605:
Fixes #2701 

I've introduced a jump method for the gesturecontroller here which also makes sure we send a commit and reset any running timeline whereas setting the progress wouldn't do that. I eventually want to make progress private set and move to this new method because having wrong commits is not really great but I'd leave that to a future PR.

I've also fixed goto comparing progress == to which could be the case - even if we won't stay at that `to` - when a goto was called before but the timeline hasn't started yet. This would cause an early return even though the second goto should overwrite the first.

### DIFF
--- a/lib/Gestures/GestureController.vala
+++ b/lib/Gestures/GestureController.vala
@@ -319,6 +319,18 @@ public class Gala.GestureController : Object {
         finish ((to > progress ? 1 : -1) * 1, to);
     }
 
+    public void jump (double to) {
+        if (running && !recognizing) {
+            /* We are animating to a snap point so stop the animation */
+            finished ();
+        }
+
+        var clamped_to = to.clamp ((int) overshoot_lower_clamp, (int) overshoot_upper_clamp);
+
+        target?.propagate (COMMIT, action, clamped_to);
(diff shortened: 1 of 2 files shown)


zeebok (approved):
Does the trick for me when I tested!

PR #2789, opened 2026-03-02
Implement maximize gesture

### LINKED ISSUE #1697: Implement "Toggle Maximized" gesture in Gala
### What Happened?

Right now "Toggle Maximized" gesture is handled by Touchégg, which doesn't support Wayland.

### Steps to Reproduce

1. Launch Wayland session
2. Try to use "Toggle Maximized" gesture

### Expected Behavior

"Toggle Maximized" gesture in Gala

### OS Version

7.x (Horus)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

leolost2605:
Fixes #1697 

On my system touchegg had a gesture configured for four finger swipe down which would mess with this gesture but idk if that's just on my system. If the gesture is configured for you too just kill touchegg and you can test this.

To test make sure you enable it via dconf editor and not via the settings because that would also enable the touchegg gesture which also messes with this.

I'm not sure how the old gesture worked but I decided to go for swipe up -> maximize, swipe down -> unmaximize. 
It will animate a tile from the window position to maximized and from maximized to the old window position. I'd be happy about some UX input here whether it's ok like it's currently or what could be improved :)

https://github.com/user-attachments/assets/c21c3f85-60b9-451a-a35f-b769b8ded172

### DIFF
--- a/src/Widgets/WindowMaximizer.vala
+++ b/src/Widgets/WindowMaximizer.vala
@@ -1,0 +1,94 @@
+/*
+ * Copyright 2026 elementary, Inc. (https://elementary.io)
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ *
+ * Authored by: Leonhard Kargl <leo.kargl@proton.me>
+ */
+
+public class Gala.WindowMaximizer : ActorTarget, RootTarget {
+    public WindowManager wm { get; construct; }
+
+    public Clutter.Actor? actor { get { return tile; } }
+
+    private Clutter.Actor tile;
+    private GestureController controller;
+
+    private Meta.Window? current_window;
+
+    public WindowMaximizer (WindowManager wm) {
+        Object (wm: wm);
+    }
+
+    construct {
+        tile = new Clutter.Actor () {
+            opacity = 200
+        };
+        Drawing.StyleManager.get_instance ().bind_property ("theme-accent-color", tile, "background-color", SYNC_CREATE);
+
+        add_child (tile);
+
+        controller = new GestureController (TOGGLE_MAXIMIZED);
+        controller.add_trigger (new GlobalTrigger (TOGGLE_MAXIMIZED, wm));
+        add_gesture_controller (controller);
+
+    }
+
+    public override void start_progress (GestureAction action) requires (action == TOGGLE_MAXIMIZED) {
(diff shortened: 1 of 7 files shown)


src/Widgets/WindowMaximizer.vala
```diff
@@ -0,0 +1,94 @@
...
+        if (!current_window.maximized_horizontally) {
+            return current_window.get_frame_rect ();
+        }
+
+        var unmaximized_geometry = WindowListener.get_default ().get_unmaximized_state_geometry (current_window);
+        if (unmaximized_geometry != null) {
+            return unmaximized_geometry.inner;
+        }
+
+        return { 0, 0, 0, 0 };
```
garaevdi: Could this be something like `return { workarea.width / 2, workarea.height / 2, 0, 0}` ?

At least for me, shrinking to the center resembles unmaximizing more than shrinking to the corner.

garaevdi:
Tested it once more, but still it's kinda broken. Could `get_unmaximized_state_geometry ()` return garbage data instead of null? This could also be related to my multimonitor setup or to the fact that I'm using mutter 49.

https://github.com/user-attachments/assets/6eafd4e8-034a-42bc-a3d9-8f8f5fd65c30

src/Widgets/WindowMaximizer.vala
```diff
@@ -0,0 +1,94 @@
...
+        var unmaximized_geometry = WindowListener.get_default ().get_unmaximized_state_geometry (current_window);
+        if (unmaximized_geometry != null) {
+            return unmaximized_geometry.inner;
+        }
+
+        return { workarea.x + (workarea.width / 2), workarea.y + (workarea.height / 2), 0, 0 };
+    }
+
+    public override void commit_progress (GestureAction action, double commit) requires (action == TOGGLE_MAXIMIZED) {
+        if (current_window == null) {
```
garaevdi, line 72: This check should be something like: `if (current_window == null || current_window.get_window_type () != Meta.WindowType.NORMAL || !current_window.can_maximize ())` so this wouldn't happen

https://github.com/user-attachments/assets/f5b47cb9-4913-4510-b95f-a77b6d2279a7

leolost2605: Should be fixed

PR #2803, opened 2026-03-26
OSK infrastructure

### LINKED ISSUE #2: Modernize the keyboard layout
## Problem
The keyboard layout used in elementary 5.1 (and 6.0 beta afaik) feels very out of place in the current design space of in Elementary OS. I see that it works right now as a last resort tool that can be used when a keyboard is not attached to a computer, however as more and more laptops are also 2in1s I think it would make sense to support a keyboard that can be used on the screen. I don't do this often on my 2in1 but I do find it uncomfortable when I need to use the OSK on elementary. The styling on it feels dated, the default size is very small and hard to type on with two hands.

## Proposal
I would purpose that the current OSK is replaced with the one that is used in GNOME (see this video: https://www.youtube.com/watch?v=6yw2bb-Kupk). 

Having the keyboard span the whole width of the screen allows you to more easily type on it.

I understand if touchscreen only use is not a priority/goal of elementary os, after-all it is not gnome, however I believe that if this change does not take too much resources to implement it could greatly improve the out-of-the-box touch screen experience on elementary os.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://app.bountysource.com/issues/113095796-modernize-the-keyboard-layout?utm_campaign=plugin&utm_content=tracker%2F75324126&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F75324126&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #3: Keyboard appears squished on lockscreen portrait mode hiDPI screen
Hi, I recently tried the elementaryOS 6 public beta 2, observed when:

- On the lockscreen
- And in (rotated) portrait mode
- And on a hiDPI display

The on-screen keyboard will appear squished such that is horizontal width is about 2/3rds what's expected (vertical height i believe was the same). It's still sortof usable but not as much as a regular size. I don't recall observing this when trying to use the keyboard in portrait mode when _not_ on the lockscreen.

Don't have a photo at this time (had to uninstall the beta for now) but can try to get one later if needed.

(Observed on an HP Elite X2 G4 tablet with a 3000x2000 display)

### LINKED ISSUE #5: onscreen keyboard to small
When you start the on screen keyboard it is to small. 
This will be used for people who will need big letters or for a touch screen.
meaning when opening it. is should be useful and not this small.

Launchpad Details: [#LP1497549](https://bugs.launchpad.net/bugs/1497549) kay van der Zander - 2015-09-19 12:27:52 +0000

### LINKED ISSUE #1460: Better support of 2in1 ("convertible") laptops
### Problem

I use a XPS 13 2in1 which I really love.
Also, I used a lenovo yoga 13 in the past, which I also used. Back then, I was using windows 8 and I really liked how a on-screen keyboard automatically appeared when I focused input boxes in the login screen or any other application when I touched onto the input and thus gave it the focus while the screen was turned around.

### Proposal

I would like to have elementary OS display an on screen keyboard as well when the device is operating in "converted" mode

### Prior Art (Optional)

_No response_

leolost2605:
Fixes #1460 
Fixes elementary/keyboard#2
Fixes elementary/keyboard#3
Fixes elementary/keyboard#5

Note: To test this you need to start an app with GTK_IM_MODULE=wayland

TODO:
- [x] icons for special buttons
- [x] Popup keys -> follow up
- [x] Language selection -> follow up
- [x] Emoji selection -> follow up
- [x] IBus suggestions -> follow up


leolost2605:
This should be enough infrastructure for an initial osk:

https://github.com/user-attachments/assets/677b6e27-e73f-4c88-82bd-06815a6699b9

@danirabbit @desiderantes any opinions/suggestions on design etc.? (I pinged you because I saw you talked about OSK in discord). If you have some ideas regarding structure/design/etc. that would be great! 

If not I'll probably do it like GNOME does it so that we can also reuse their file format for keyboard layouts :shrug:

danirabbit:
@leolost2605 being able to re-use layouts from an existing OSK sounds like a good idea to me! We can always revisit if for some reason it's too restrictive, but otherwise I like doing less work 😜

desiderantes:
There's some wayland protocols for virtual keyboards that might be of interest ( I experimented implementing them, because I was playing with basing the logic on the Android keyboard). Mainly the one for [virtual keyboard](https://wayland.app/protocols/virtual-keyboard-unstable-v1) and maybe [input v2](https://wayland.app/protocols/xx-input-method-v2)

leolost2605:
> There's some wayland protocols for virtual keyboards that might be of interest ( I experimented implementing them, because I was playing with basing the logic on the Android keyboard). Mainly the one for [virtual keyboard](https://wayland.app/protocols/virtual-keyboard-unstable-v1) and maybe [input v2](https://wayland.app/protocols/xx-input-method-v2)

Thanks for the input! I don't think input v2 applies here. Also we have usually done any interaction with the compositor that doesn't involve actual surfaces via dbus so I would keep doing it via dbus in this case too.

leolost2605:
I've been using this for a few months now and it's working great. I haven't marked this ready the whole time because I wanted to get the TODOs done but I think it's probably better to save them for a follow up since the main thing is already working great. So this should be ready for review now. The main TODO that maybe should still be done in this PR are icons for
- delete
- enter
- shift
- hiding the keyboard

To test this you need to be in touch mode (tablets by default, 2 in 1s  when in tablet mode) or enable it manually via `/org/gnome/desktop/a11y/applications/screen-keyboard-enabled`. Also the same things as in #2793 apply (GTK_IM_MODULE must be unset or set to wayland)

image

danirabbit:
If someone doesn't get to it before me I'll propose icons Monday!

wpkelso:
@danirabbit Many keyboards I've seen use symbols that have unicode codepoints in the [misc. technical block](https://symbl.cc/en/unicode/blocks/miscellaneous-technical/) or the [arrows block] for the action and modifier keys. That might could be another solution? Some of them are... questionable in GitHub's choice of font lol

For example:
⌘ ⌥ ⌫ ⌦ ⌨ ⎇ ⎆ ⏎ ⏿

danirabbit:
it looks like we actually get icon names from the layout json files:

- [ ] osk-caps-lock-symbolic
- [ ] osk-delete-symbolic
- [ ] osk-emoji-picker-symbolic
- [ ] osk-enter-symbolic
- [ ] osk-hide-symbolic
- [ ] osk-layout-symbolic
- [ ] osk-shift-symbolic
- [ ] osk-zwnj-symbolic

danirabbit (approved):
Approving on behalf of UX! This is a super awesome start. Already pretty nice to type on. Can't wait to continue to iterate here 🩷

danirabbit:
@lenemter any opposition to getting this in?

daemon/OSK/KeyboardModel/KeyboardModelBuilder.vala
```diff
@@ -0,0 +1,151 @@
...
+        //  var popup_key = new Key (
+        //      1.0f,
+        //      1.0f,
+        //      ACTION_PREFIX + "popup." + popup_key_string,
+        //      null,
+        //      popup_key_string,
+        //      null
+        //  );
+        //  current_key_popup_keys.append (popup_key);
+    }
```
lenemter:
```suggestion
```

lenemter (approved):

PR #2805, opened 2026-03-31
WindowClone: display modal dialogs on top of parent window

### LINKED ISSUE #198: Multitasking should be disabled when an application sends an important message.

lenemter:
Closes https://github.com/elementary/gala/issues/198

# Before
Screenshot from 2025-08-06 21 14 57

# After
Screenshot from 2025-08-06 21 13 56

### DIFF
--- a/src/Widgets/MultitaskingView/WindowClone.vala
+++ b/src/Widgets/MultitaskingView/WindowClone.vala
@@ -67,10 +67,12 @@ public class Gala.WindowClone : Widget, RootTarget {
     private ulong check_confirm_dialog_cb = 0;
 
     private Clutter.Actor clone_container;
+    private Clutter.Actor child_clone_container;
     private Gala.CloseButton close_button;
     private ActiveShape active_shape;
     private Clutter.Actor window_icon;
     private Tooltip window_title;
+    private GLib.ListModel window_list_model;
 
     private GestureController gesture_controller;
 
@@ -131,9 +133,14 @@ public class Gala.WindowClone : Widget, RootTarget {
         };
         bind_property ("monitor-scale", active_shape, "monitor-scale");
 
-        clone_container = new Clutter.Actor () {
-            pivot_point = { 0.5f, 0.5f }
-        };
+        window_list_model = new WindowListModel (wm.get_display (), STACKING, true, false, -1, null, new ChildFilter (window));
+        window_list_model.items_changed.connect_after (update_targets);
+
+        child_clone_container = new Clutter.Actor ();
+        child_clone_container.bind_model (window_list_model, create_child_func);
+
+        clone_container = new Clutter.Actor ();
+        clone_container.add_child (child_clone_container);
 
         window_title = new Tooltip (monitor_scale);
         bind_property ("monitor-scale", window_title, "monitor-scale");
@@ -176,6 +183,12 @@ public class Gala.WindowClone : Widget, RootTarget {
         active_shape.restore_easing_state ();
     }
 
+    private Clutter.Actor create_child_func (Object obj) requires (obj is Meta.Window) {
+        unowned var child_window = (Meta.Window) obj;
+
+        return new Clutter.Clone ((Meta.WindowActor) child_window.get_compositor_private ());
+    }
+
     private void reallocate () {
         window_icon = new WindowIcon (window, WINDOW_ICON_SIZE, (int)Math.round (monitor_scale)) {
             visible = mode != SINGLE_APP_OVERVIEW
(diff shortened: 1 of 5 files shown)


leolost2605 (changes requested):
This looks really cool, thank you! Just a few comments :)

src/Widgets/MultitaskingView/WindowClone.vala
```diff
@@ -179,6 +183,32 @@ public class Gala.WindowClone : ActorTarget, RootTarget {
         finish_drag ();
     }
 
+    private bool filter_child_window (Object obj) requires (obj is Meta.Window) {
+        unowned var window_to_filter = (Meta.Window) obj;
+        return window.is_ancestor_of_transient (window_to_filter);
+    }
+
+    private void on_items_changed (uint pos, uint removed, uint added) {
```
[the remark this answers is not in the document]
leolost2605: If we iterate over the children with an index we could just get the windows from the window list model couldn't we? Since that uses an array list the access time is also still constant. Also I would use 
for (var child = child_container.get_first_child (); child != null; child = child.get next sibling ())
Instead of get_children because get_children already iterates over the children so we would do it twice.

src/Widgets/MultitaskingView/WindowClone.vala
```diff
@@ -639,4 +689,21 @@ public class Gala.WindowClone : ActorTarget, RootTarget {
...
+
+    private class ChildFilter : Gtk.Filter {
+        public Meta.Window parent_window { private get; construct; }
+
+        public ChildFilter (Meta.Window parent_window) {
+            Object (parent_window: parent_window);
+        }
+
+        public override Gtk.FilterMatch get_strictness () {
+            return ALL;
```
leolost2605: We don't match all items only some which is the default so you don't have to override this.

src/Widgets/MultitaskingView/WindowClone.vala
```diff
@@ -279,6 +293,37 @@ public class Gala.WindowClone : ActorTarget, RootTarget {
         add_target (new PropertyTarget (MULTITASKING_VIEW, window_title, "opacity", typeof (uint), 0u, 255u));
 
         add_target (new PropertyTarget (MULTITASKING_VIEW, close_button, "opacity", typeof (uint), 0u, 255u));
+
+        var window_buffer_rect = window.get_buffer_rect ();
+        var window_shadow_spread_x = window_rect.x - window_buffer_rect.x;
+        var window_shadow_spread_y = window_rect.y - window_buffer_rect.y;
+
+        for (var i = 0; i < child_clone_container.get_n_children (); i++) {
```
leolost2605: Like i said in the previous comment I think we should iterate directly over the children via the 
for (get first child; child != null; child = child.get next sibling)
and manually count up the index because get child at index has O(n) complexity so we would end up with quadratic instead of linear which isnt that big of difference since we probably wont ever have lots of transients but it's also not a lot cleaner so i think we should do it linear.

src/Widgets/MultitaskingView/WindowClone.vala
```diff
@@ -141,9 +143,15 @@ public class Gala.WindowClone : ActorTarget, RootTarget {
...
 
-        clone_container = new Clutter.Actor () {
-            pivot_point = { 0.5f, 0.5f }
-        };
+        window_list_model = new WindowListModel (wm.get_display (), STACKING, true, false, -1, null, new ChildFilter (window));
+
+        child_clone_container = new Clutter.Actor ();
+        child_clone_container.bind_model (window_list_model, create_child_func);
+        child_clone_container.child_added.connect (update_targets);
+        child_clone_container.child_removed.connect (update_targets);
```
leolost2605: Since multiple changes to the children can be condensed in one items changed emission and we resetup all targets anyways in update_targets i think we should connect to items changed instead of possibly updating the targets multiple times for one items changed. So we should either use connect_after or connect to the signal after we called bind model.

leolost2605 (changes requested):
Just a few more comments :)

leolost2605 (changes requested):
This looks really cool! I think you need to change the drag logic to hide the whole clone container and not just the clone though. Otherwise if you drag it looks like this:

Image

leolost2605 (approved):
Looks great and looks good to me :)
I think we should merge this after 8.5 was released though just to make sure we get some testing in daily since it is a pretty big and involved change and just to be sure we catch some glitches (e.g. moving a modal dialog to a new workspace via the gesture is a bit buggy but it's not great in main as well so I wouldn't treat it as blocking)

PR #2806, opened 2026-04-02
SwipeTrigger: don't hold strong reference on the actor

lenemter:
Fixes memory leak found while testing https://github.com/elementary/gala/pull/2805


lib/Gestures/SwipeTrigger.vala
```diff
@@ -11,11 +11,24 @@
  * It enables touchpad and (once supported) touchscreen backends for the given actor.
  */
 public class Gala.SwipeTrigger : Object, GestureTrigger {
-    public Clutter.Actor actor { get; construct; }
+    public weak Clutter.Actor? actor;
```
leolost2605:
```suggestion
    private weak Clutter.Actor? actor;
```

leolost2605 (changes requested):

PR #2807, opened 2026-04-02
WindowSwitcher: correctly handle multiple keybind modifiers

lenemter:
Fixes https://github.com/elementary/gala/issues/2794

### DIFF
--- a/src/Widgets/WindowSwitcher/WindowSwitcher.vala
+++ b/src/Widgets/WindowSwitcher/WindowSwitcher.vala
@@ -13,7 +13,7 @@ public class Gala.WindowSwitcher : AbstractSwitcher, GestureTarget, RootTarget {
     public bool opened { get; private set; default = false; }
 
     private GestureController gesture_controller;
-    private int modifier_mask;
+    private Clutter.ModifierType modifier_mask;
     private Gala.ModalProxy? modal_proxy;
     private int previous_icon_index = 0;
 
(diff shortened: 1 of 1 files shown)


leolost2605 (changes requested):
I think it would help to know what the original reason was for only using the primary modifier in a mask. I had a quick a look and gnome shell is still doing it? The only reason I could think of are the backwards bindings e.g. to make sure we close the switcher if alt or super is released even though shift is still held?
Also for some reason this breaks Super + ` (which is the default shortcut for cycling windows of an app)?

ifazk:
The primaryModifier code was introduced in Gnome in the following commit.

- https://github.com/GNOME/gnome-shell/commit/566d566f267f35bd3f259ff1fae9268b328bf924
- https://bugzilla.gnome.org/show_bug.cgi?id=645200

The above discusses computing the primaryModifier instead of hard coding it to Alt for Ctrl+Alt+Tab and Alt+Tab.

The next place to look at is the following commit implementing Ctrl+Alt+Tab. The initial commit already starts of with Alt being hard coded, copied from the Alt+Tab implementation.

- https://github.com/GNOME/gnome-shell/commit/d3de4e3fbda046c844f32d61c494960c5b1bc52d
- https://bugzilla.gnome.org/show_bug.cgi?id=618885

I'm pasting the comment about the hard coding being copied from AltTab, but please read the context around it. The comment is about a server round trip, not about the key combination.

```diff
@@ +148,3 @@
+    _keyReleaseEvent : function(actor, event) {
+        let [x, y, mods] = global.get_pointer();
+        let state = mods & Clutter.ModifierType.MOD1_MASK;

Wow, that's an ugly hack round-tripping on every keyrelease event. But I guess it's copied straight-over from altTab.js so likely is as-good-as-possible. Though it doesn't look like the usage of global.get_pointer() was even considered for [bug 629368](https://bugzilla.gnome.org/show_bug.cgi?id=629368) ... yeah, the Mutter code to "get it right" is complex, but it would be better to export it out than just doing something cheesy, slow, and racy. Not a blocker for this patch, but I'd appreciate it if we had a bug filed. (OK, maybe this is better-enemy-of-good-enough, but if we already *have* the better code.)
```

For completeness, code mentioning `MOD1_MASK` in Alt+Tab was introduce in the following two commits. But hard coding Alt as the modifier to watch for makes sense if you are only dealing with Alt+Tab, so I don't how much to read into these.

- https://github.com/GNOME/gnome-shell/commit/373fa3c3250f4309ebee9e384200c14211edbf6c
- https://github.com/GNOME/gnome-shell/commit/907dd4381b7252646be3cac026fd794ca8b58768

I think it's one of these two that we're dealing with, but hard to know which. You might have other ideas as well.

1. There is no technical reason for using the primary modifier. Things started of as hard coded to have a single modifier and that behaviour was kept when the code was refactored.
2. It is easier to deal with key-release for a single modifier key.

PR #2808, opened 2026-04-03
WorkspaceHideTracker: send `window state changed` signal with delay

### LINKED ISSUE #2724: When using wl-copy/wl-paste (wl-clipboard) the dock shows up although Terminal is maximized

lenemter:
Fixes https://github.com/elementary/gala/issues/2724

### DIFF
--- a/lib/Gestures/WorkspaceHideTracker.vala
+++ b/lib/Gestures/WorkspaceHideTracker.vala
@@ -1,6 +1,6 @@
 /*
  * SPDX-License-Identifier: GPL-3.0-or-later
- * SPDX-FileCopyrightText: 2025 elementary, Inc. (https://elementary.io)
+ * SPDX-FileCopyrightText: 2025-2026 elementary, Inc. (https://elementary.io)
  */
 
 public class Gala.WorkspaceHideTracker : Object {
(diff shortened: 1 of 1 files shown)


leolost2605 (changes requested):
I'm not sure simply adding a timeout here is a good idea. It might fix it for this specific case but that seems more like treating a symptom. Also while 100ms isnt really long it's still noticeable and I don't think we want that.

Instead I think there's still a bug in the dock reveal logic somewhere because I've seen other reports about the dock not unrevealing and had it happen myself a few times. In this case what should happen IMO is 
wl paste -> wl paste window appears -> dock starts revealing -> wl paste window disappears -> dock starts unrevealing. And since a dock unreveal should cancel a reveal and the wl paste window appears and disappears pretty much instantly afaict the user shouldn't see the dock.
Lmk what you think :)

PR #2809, opened 2026-04-04
lib: sort files into folders

lenemter:
Everything seems to be in either /lib or /lib/Gestures folders and it's hard to navigate

image


leolost2605 (approved):
Makes sense to me. We could even add another folder AppSystem for app appcache and appsystem. Also don't forget POTFILES :)

PR #2811, opened 2026-04-05
Tooltip: use background color from stylesheet and add text shadow

lenemter:
The new background color for tooltips was taken from the stylesheet

For the text shadow, I implemented BoxShadowEffect since box blur is very cheap, and text shadows don't need to be pixel-perfect anyway

Currently, the effect can be added via Gala.BoxBlurManager, but I plan to remove it once I figure out how to properly implement https://github.com/elementary/gala/pull/2776 :)


zeebok:
I am having a hard time seeing what this changes exactly. I thought maybe the tooltips on thins like the dock and wingpanel but it doesn't seem to change anything with those.

The code itself seems solid though!

danirabbit (approved):
Gala draws its own tooltips in Multitasking View. Nice job, looks good :)

PR #2812, opened 2026-04-07
Fix POTFILES

leolost2605:
Remnants from #2809 

Fixes CI on main


lenemter (approved):
Oops. Thanks for fixing this

PR #2815, opened 2026-04-11
Show git commit hash in `gala --version`

lenemter:
Should make it easier for early access users to get the accurate version information

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,6 +36,7 @@ jobs:
       env:
         DESTDIR: out
       run: |
+        git config --global --add safe.directory "$(pwd)"
         meson build -Ddocumentation=true -Dtests=true
         ninja -C build
         ninja -C build install
@@ -78,6 +79,7 @@ jobs:
         env:
           DESTDIR: out
         run: |
+          git config --global --add safe.directory "$(pwd)"
           meson build
           ninja -C build install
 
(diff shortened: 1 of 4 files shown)


meson.build
```diff
@@ -25,6 +25,16 @@ conf.set('GETTEXT_PACKAGE', meson.project_name())
 conf.set('LOCALEDIR', locale_dir)
 conf.set('PLUGINDIR', plugins_dir)
 conf.set('VERSION', meson.project_version())
+
+git = find_program('git', required: false)
+git_hash = 'unknown'
+if git.found()
+	# fixes 'fatal: detected dubious ownership in repository at '/__w/gala/gala' in CI
+	run_command(['git', 'config', '--global', '--add', 'safe.directory', meson.current_source_dir ()], capture : true, check : true)
```
ryonakano: I suppose this should be handled in the workflow files instead of here.

.github/workflows/main.yml
```diff
@@ -36,6 +36,7 @@ jobs:
       env:
         DESTDIR: out
       run: |
+        git config --global --add safe.directory "$(pwd)"
```
ryonakano, line 39:
```suggestion
        git config --global --add safe.directory "$GITHUB_WORKSPACE"
```

It seems like we use this environment variable in other places. Also this is (slightly) more efficient than a command substitution.

.github/workflows/main.yml
```diff
@@ -78,6 +79,7 @@ jobs:
         env:
           DESTDIR: out
         run: |
+          git config --global --add safe.directory "$(pwd)"
```
ryonakano, line 88:
```suggestion
          git config --global --add safe.directory "$GITHUB_WORKSPACE"
```

Same here.

PR #2820, opened 2026-04-20
README: Add section about how to run the tests

leolost2605:
Includes some instructions about how to run the tests. Lmk if it should be more detailed or needs a bit of rewording to make it more clear.


zeebok (approved):
This looks good to me

PR #2826, opened 2026-04-23
WindowManager/ShellClients: Explicitly override window group

leolost2605:
To select the correct group for shell windows the window manager currently has to know what the possibilities are and then query the shell clients manager to see if the windows are one of that.
It also only works if we know the correct info about the windows before their window actor is turning visible.

Instead of this hacky solution we introduce a `override_window_group` in the wm. This is then called by the shellclientsmanager at that moment when it knows what the window will be.

This has a few advantages:
- We have a clear interface for how to handle changing the group of a window that also makes sure the wm knows about it and doesn't try to mess with it.
- It scales a lot better by allowing the actual places that know that the window should be somewhere else to change its group instead of the wm having to query everywhere. This will be useful for example with the osk which will also need its own group.
- It reduces the public api the shellclientsmanager has to expose
- It should finally fix the wingpanel sometimes being invisible in the multitasking view which for me still happens even after #2646
- The second commit can remove some now unecessary code

### DIFF
--- a/src/ShellClients/ShellClientsManager.vala
+++ b/src/ShellClients/ShellClientsManager.vala
@@ -8,7 +8,7 @@
 public class Gala.ShellClientsManager : Object, GestureTarget {
     private static ShellClientsManager instance;
 
-    public static void init (WindowManager wm) {
+    public static void init (WindowManagerGala wm) {
         if (instance != null) {
             return;
         }
@@ -20,7 +20,7 @@ public class Gala.ShellClientsManager : Object, GestureTarget {
         return instance;
     }
 
-    public WindowManager wm { get; construct; }
+    public WindowManagerGala wm { get; construct; }
 
     private NotificationsClient notifications_client;
     private ManagedClient[] protocol_clients = {};
@@ -31,7 +31,7 @@ public class Gala.ShellClientsManager : Object, GestureTarget {
     private GLib.HashTable<Meta.Window, ExtendedBehaviorWindow> positioned_windows = new GLib.HashTable<Meta.Window, ExtendedBehaviorWindow> (null, null);
     private GLib.HashTable<Meta.Window, MonitorLabelWindow> monitor_label_windows = new GLib.HashTable<Meta.Window, MonitorLabelWindow> (null, null);
 
-    private ShellClientsManager (WindowManager wm) {
+    private ShellClientsManager (WindowManagerGala wm) {
         Object (wm: wm);
     }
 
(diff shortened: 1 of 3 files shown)


src/WindowManager.vala
```diff
@@ -1040,23 +1035,50 @@ namespace Gala {
...
+            InternalUtils.wait_for_window_actor_visible (window, (actor) => {
+                InternalUtils.clutter_actor_reparent (actor, new_group);
+
+                // FIXME: workaround for https://github.com/elementary/dock/issues/537
+                actor.set_scale (1.0, 1.0);
+                actor.opacity = 255;
+            });
+        }
+
         private void check_shell_window (Meta.WindowActor actor) {
```
lenemter: Since this function no longer checks shell windows, maybe it should be renamed?

src/ShellClients/ShellClientsManager.vala
```diff
@@ -32,7 +32,7 @@ public class Gala.ShellClientsManager : Object, GestureTarget {
     private GLib.HashTable<Meta.Window, MonitorLabelWindow> monitor_label_windows = new GLib.HashTable<Meta.Window, MonitorLabelWindow> (null, null);
 
     private ShellClientsManager (WindowManager wm) {
-        Object (wm: wm);
+        Object (wm: (WindowManagerGala) wm);
```
lenemter:
```
    private ShellClientsManager (WindowManagerGala wm) {
        Object (wm: wm);
```

leolost2605:
@lenemter I've been thinking a bit more about it and maybe we should provide an enum of layers instead of allowing to provide any clutter.actor? What do you think about that? This would allow us to keep a bit more of an overview what layers exist and where goes what.

lenemter:
@leolost2605 I think it's a good idea to put different windows into different layers and an enum seems to be a good and intuitive solution, so that's a :+1: from me

leolost2605:
@lenemter all right I updated it and addressed the review comments so should be ready again :)

src/WindowManager.vala
```diff
@@ -1040,23 +1040,59 @@ namespace Gala {
...
 
-            if (ShellClientsManager.get_instance ().is_shell_window (window)) {
-                InternalUtils.clutter_actor_reparent (actor, shell_group);
+            /* Check if we're a transient of a window with an overridden group and if so place there */
+            Clutter.Actor? overridden_group = null;
+            window.foreach_ancestor ((ancestor) => {
+                if (overridden_window_group.has_key (ancestor)) {
+                    overridden_group = overridden_window_group[ancestor];
+                    return false;
+                }
```
lenemter, line 1082: Wouldn't it be easier to store `WindowGroup` in `overridden_window_group` and then call `override_window_group` for the transient window instead of reparenting manually?

leolost2605: Make sense, I updated it :)

lenemter:
> It should finally fix the wingpanel sometimes being invisible in the multitasking view which for me still happens even after https://github.com/elementary/gala/pull/2646

I still can reliable reproduce this issue, but I guess it's unrelated to this codepath

lenemter (approved):

PR #2828, opened 2026-04-24
MultitaskingView: Fix escape after workspace switch

leolost2605:
After a workspace switch you currently have to press one of the arrow keys once so that the gala.root captures the input focus again and only then pressing esc will work.
This PR makes sure it will always work after switching workspaces.


zeebok:
If I move to a workspace with a browser open the escape key still doesn't work, it seems to give focus to the browser. Even the arrow keys seem to scroll the browser page! This still occurs even after installing this fix.

lenemter (approved):
LGTM!

PR #2831, opened 2026-04-25
Inrtoduce BrightnessManager dbus interface

garaevdi:
With gnome 49 screen brightness management was moved into mutter. This is a simple and naive implementation of backlight interface which could be controlled via DBus.

Note: to build this you would need mutter 50.2


leolost2605:
Did you check whether it might already be exposed via org.gnome.mutter.displayconfig ?

garaevdi:
> Did you check whether it might already be exposed via org.gnome.mutter.displayconfig ?

Yes, but I think it's better to abstract it with a gala interface. Otherwise both switchboard-plug-power and wingpanel-power would need to implement the same things (handling monitors changes, converting raw brightness values to percentages).

<sub>Also you won't have to deal with complicated dbus types</sub>

src/BrightnessManager.vala
```diff
@@ -0,0 +1,42 @@
...
+            backlight.get_brightness_info (out min, out max);
+            backlight.set_brightness ((int) (value * (max - min) / 100) + min);
+        }
+    }
+
+    private unowned Meta.MonitorManager monitor_manager;
+    private unowned Meta.Backlight? backlight;
+
+    public BrightnessManager (WindowManagerGala wm) {
+        monitor_manager = wm.get_display ().get_context ().get_backend ().get_monitor_manager ();
```
leolost2605: We should use gobject style construction here. I.e. the constructor only contains a call to Object (...) and the rest is handled in the `construct` block.

src/BrightnessManager.vala
```diff
@@ -0,0 +1,42 @@
...
+            int min, max;
+            backlight.get_brightness_info (out min, out max);
+            backlight.set_brightness ((int) (value * (max - min) / 100) + min);
+        }
+    }
+
+    private unowned Meta.MonitorManager monitor_manager;
+    private unowned Meta.Backlight? backlight;
+
+    public BrightnessManager (WindowManagerGala wm) {
```
leolost2605: Any reason this is WindowManagerGala and not just WindowManager? Or maybe even just Display or MonitorManager?

src/Misc/BrightnessManager.vala
```diff
@@ -0,0 +1,42 @@
+[DBus (name = "io.elementary.gala.BrightnessManager")]
```
leolost2605, line 9: Missing a license header

src/meson.build
```diff
@@ -76,6 +76,10 @@ gala_bin_sources = files(
     'Widgets/WindowSwitcher/WindowSwitcherIcon.vala',
 )
 
+if mutter50_dep.found()
+    gala_bin_sources+=files('BrightnessManager.vala')
```
leolost2605: Instead of doing this we usually just wrap the whole contents of the file in an `#if`

vapi/Meta-18.metadata
```diff
@@ -12,6 +12,7 @@ BackgroundContent cheader_filename="meta/meta-background-content.h"
 BackgroundGroup cheader_filename="meta/meta-background-group.h"
 BackgroundImage cheader_filename="meta/meta-background-image.h"
 BackgroundImageCache cheader_filename="meta/meta-background-image.h"
+Backlight cheader_filename="meta/meta-backlight.h"
```
leolost2605, line 15: I'm not too knowledgeable in vapi but since you adjusted the vapi itself do we still need the metadata?

leolost2605 (changes requested):
> Yes, but I think it's better to abstract it with a gala interface. Otherwise both switchboard-plug-power and wingpanel-power would need to implement the same things (handling monitors changes, converting raw brightness values to percentages).

Hmm yeah that makes sense. Also I just saw there are vague plans to drop the backlight from the dbus api anyways :sweat_smile: 
Though I think we should use having to implement this to already prepare for individual multi monitor backlight control which AFAIR is the reason why gnome moved the brightness control into mutter.

Also ig we have to implement save and restore too since it was dropped from mutter (though that only made 51 so ig we can follow up):
https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/4814
https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/4006

src/BrightnessManager.vala
garaevdi: Fixed

src/BrightnessManager.vala
garaevdi: Just blindly followed other dbus interfaces implementations, now fixed

src/Misc/BrightnessManager.vala
garaevdi: Added

src/meson.build
garaevdi: Fixed

vapi/Meta-18.metadata
garaevdi: Yes, it used by vapigen to generate correct bindings, so the future porters wouldn't change correct headers with default `meta/meta.h` header

garaevdi:
@leolost2605 Rebased onto main and added keyboard shortcuts. 
Not sure why this builds on development-target tho... IIRC mutter was missing includes in its' public headers. [A fix](https://gitlab.gnome.org/GNOME/mutter/-/commit/1a73942e4042300415645841a371181b4ccbf942) is already there, but it is only available since 50.2, while resolute ships 50.1

leolost2605:
>  Though I think we should use having to implement this to already prepare for individual multi monitor backlight control which AFAIR is the reason why gnome moved the brightness control into mutter.

Any thoughts on this? I'd like to do it now because it requires a different public API.

garaevdi:
> > Though I think we should use having to implement this to already prepare for individual multi monitor backlight control which AFAIR is the reason why gnome moved the brightness control into mutter.
> 
> Any thoughts on this? I'd like to do it now because it requires a different public API.

Not really, IMO some kind of manager object is still needed to provide a way to control "global" brightness scale or perhaps lock other monitors' scales to a global one. The last commit does something like this, but it's super WIP.

(And also it wouldn't compile due to another missing include in mutters' headers 🫠️)

leolost2605:
> "global" brightness scale

Yeah I was wondering if we even need that? I mean what would be the "global" scale if you have two monitors which both support brightness adjustment?

I had a quick look through your new commit and I think I wouldn't do individual dbus objects for individual monitors. The interface isn't expected to grow much bigger and having a single interface and addressing via an index is much simpler for the user of the interface IMO.

Maybe something like this:
```
// Core part
signal void monitors_changed ();
signal void monitor_brightness_changed (uint index);

void set_monitor_brightness (uint index, int brightness);
int get_monitor_brightness (uint index);
string get_monitor_name (uint index);
int get_n_monitors ();

// Some utilities
// For keybinds?
set_primary_monitor_brightness (int brightness);

// Idk do we want that?
set_all_monitor_brightness ();
```
- The `index` doesn't actual mean anything and it is only guaranteed to address the same monitor until the next monitors_changed

PR #2832, opened 2026-04-25
Support running as greeter

leolost2605:
With some minimal changes we can support running gala as the compositor for the greeter and given the many things gala has to do these days I think this is probably the way to go (e.g. also on screen keyboard, input methods, etc.)

To keep a nice overview of things that are different between the different session types I decided to introduce a `SessionSettings` namespace that actually has meaningful options like `should_blur_background`, `should_fade_in`, etc instead of having the appropriate places just query the session type and then decide what to do. IMO it's good to keep this overview instead of having to go hunt through all files and also it's easy to adjust how we get the session type etc., but I'd like to hear other opinions. Or if there are other ideas for how to handle these differences between the session types I'd be happy to hear them.

This already works with #elementary/greeter#878 but it still needs a bit of polish.

### DIFF
--- a/data/io.elementary.desktop.wm.shell
+++ b/data/io.elementary.desktop.wm.shell
@@ -7,3 +7,19 @@ session-type=desktop
 launch-on-x=true
 args=io.elementary.dock
 session-type=desktop
+
+[Greeter wingpanel]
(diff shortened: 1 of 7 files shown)


leolost2605:
Superseded by #2854

PR #2833, opened 2026-04-26
Tests: avoid recompiling libgala for each test

lenemter:
Reuse libgala library instead of compiling a test + libgala into every test executable.

Because of this change I introduced a TESTS preprocessor variable to expose needed internal parts of the library to the tests.

This should significantly speed up compilation for large numbers of tests.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,9 +36,14 @@ jobs:
       env:
         DESTDIR: out
       run: |
-        meson build -Ddocumentation=true -Dtests=true
-        ninja -C build
-        ninja -C build install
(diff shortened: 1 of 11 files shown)


leolost2605 (changes requested):
Makes sense to me but I think we should use a separate new library so that enabling tests doesn't change anything about the actual one. Also the #if TESTS arent really nice, is there a way to do it without introducing a lot of these conditions? Maybe have a macro that evaluates to private or public? Does vala support that?

leolost2605:
Just a quick google didn't investigate further but maybe --internal-vapi and friends might work?

PR #2835, opened 2026-04-29
Tests: compile into a single executable

lenemter:
Closes https://github.com/elementary/gala/pull/2833

Achieves the same performance boost as https://github.com/elementary/gala/pull/2833, but doesn't introduce unnecessary #if conditions


tests/meson.build
```diff
@@ -1,6 +1,32 @@
...
+test_executable = executable(
+    'io.elementary.gala.tests',
+    common_test_sources,
+    lib_test_sources,
+    gala_lib_sources,
+    dependencies: gala_base_dep,
+    install: false,
+)
+
+lib_test_sources = files()
```
leolost2605: I don't think this is needed?

tests/meson.build
```diff
@@ -1,6 +1,32 @@
...
+foreach lib_test : lib_tests
+    test(lib_test, test_executable, args: ['--test-name=@0@'.format(lib_test)] ,is_parallel: false)
```
leolost2605: I think we should keep specifying the suite here

tests/meson.build
```diff
@@ -1,6 +1,32 @@
...
+lib_tests = [
+    'GestureControllerTest',
+    'PropertyTargetTest',
+    'SetupTest',
+    'SwipeTriggerTest',
+]
+
+lib_test_sources = files()
+foreach lib_test : lib_tests
+    lib_test_sources += files('lib/@0@.vala'.format(lib_test))
```
leolost2605: Do we need the files wrap here?
Also we might want to use `'lib' / '@0@.vala'.format(lib_test)` just for convention and readability.

tests/TestCase.vala
```diff
@@ -42,11 +38,4 @@ public abstract class Gala.TestCase : Object {
 
     public virtual void tear_down () {
     }
-
-    public void assert_finalize_object<G> (ref G data) {
```
leolost2605, line 46: This seems unrelated, maybe split it to another commit?

tests/TestCase.vala
```diff
@@ -12,18 +12,14 @@
...
     private GLib.TestSuite suite;
 
     construct {
-        suite = new GLib.TestSuite (name);
+        suite = new GLib.TestSuite (this.get_class ().get_type ().name ());
     }
 
-    public int run (string[] args) {
-        Test.init (ref args);
+    public override void constructed () {
```
leolost2605, line 21: Why do we need to put this in constructed? Can't we just put it into construct?

leolost2605 (changes requested):
Makes sense to me! A few comments :)
Also maybe we should use the glib.test built in way to specify which tests to run? https://valadoc.org/glib-2.0/GLib.Test.init.html says we can just use `-p` to specify what to run. So we would just construct all testcases, put them in an array and then call test.init and test.run. This would also automatically give us more complete and finegrained control, e.g. to not only specify tests but also skip some, etc. We could also keep the name property if we maybe want to name them differently because we don't rely on the type name then.

tests/TestCase.vala
lenemter: `TestSuite.add_suite ()` takes ownership of provided suite, so the private suite variable becomes null. So in order to add tests to the suit we need to do `TestSuite.get_root ().add_suite ((owned) suite)` after all the tests have been added

lenemter:
@leolost2605 I don't think it's worth adding name field, at least until we really need it. Currently it's just an unneeded repetition in the code

lenemter:
@leolost2605 for some reason this broke tests for mutter 50. I pushed a commit with a fix to this branch but let me know if you want to merge it separately and I'll create a new PR

tests/meson.build
```diff
@@ -1,6 +1,30 @@
...
+    'io.elementary.gala.tests',
+    common_test_sources,
+    lib_test_sources,
+    gala_lib_sources,
+    dependencies: gala_base_dep,
+    install: false,
+)
+
+foreach lib_test : lib_tests
+    test(lib_test, test_executable, args: ['-p', '/Gala@0@'.format(lib_test)], suite: ['Gala'], is_parallel: false)
```
leolost2605: For the suite I meant especially the gala/lib part:
```suggestion
    test(lib_test, test_executable, args: ['-p', '/Gala@0@'.format(lib_test)], suite: ['Gala', 'Gala/lib'], is_parallel: false)
```

tests/Main.vala
```diff
@@ -0,0 +1,24 @@
...
+            typeof (GestureControllerTest),
+            typeof (PropertyTargetTest),
+            typeof (SwipeTriggerTest),
+        };
+
+        Object[] tests = {};
+
+        for (var i = 0; i < test_types.length; i++) {
+            tests += GLib.Object.new_with_properties (test_types[i], {}, {});
+        }
```
leolost2605: Any reason to go via the types and then for loop and not just

```suggestion
        TestCase[] tests = {
            new SetupTest (),
            new GestureControllerTest (),
            new PropertyTargetTest (),
            new SwipeTriggerTest (),
        };
```

tests/MutterTestCase.vala
```diff
@@ -25,12 +25,16 @@ public abstract class Gala.MutterTestCase : Gala.TestCase {
...
 
-    protected Meta.Context context { get; private set; }
-    protected Clutter.Stage stage { get { return (Clutter.Stage) context.get_backend ().get_stage (); } }
+    protected static Meta.Context? context { get; private set; }
+    protected static Clutter.Stage? stage { get { return (Clutter.Stage) context?.get_backend ().get_stage (); } }
 
     private MainLoop? main_loop;
 
     construct {
+        if (context != null) {
```
leolost2605, line 34: This should probably just be a static construct then.

leolost2605 (changes requested):
Looks good, just a few more comments :)

> @leolost2605 for some reason this broke tests for mutter 50. I pushed a commit with a fix to this branch but let me know if you want to merge it separately and I'll create a new PR

Yeah seems like we were doing remove_child (null) there so makes sense. If you clean up the commit history and add the fix as the first commit I think it's fine to merge with this

tests/MutterTestCase.vala
lenemter: It doesn't work for some reason =(

I suppose we don't want to use it anyway since not all tests use MutterTestCase

lenemter:
@leolost2605 Done!

leolost2605 (changes requested):
I've been thinking some more and I'm not sure anymore whether this is a good idea in its current form.
My main concern is that we loose any isolation of the test cases with this by implicitly allowing the possibility to run all at once. This wouldn't be as bad if we could just reconstruct the whole meta.context but since that doesn't work it poses a few problems:
- Resources that are accessed from multiple test cases might look different depending on if they are run together or alone and in which order. E.g. if one test case forgets to clean up the stage we might get problems there.
- We loose the possibility of customization of the environment for a specific test case. For example I think we probably want to provide the possibility for a test case to specify the type of the meta.plugin. For some a very simple mock is enough, other need a specific method mocked and maybe we want the actual windowmanagergala in some tests. Another thing is the monitor setup. In the light of recent multi monitor crashes we probably want a test case that tests multi monitor setups etc.

I came pretty far with `--internal-vapi` and a shared library but didn't quite finish it and will have a look again later this week. I think if we want to do single executable we will have to make sure that only a single test case can be run and don't want to have any static variables. A main file that decides the type, constructs it, and just calls run on the testcase would probably be better then.

Anyways just some of my thoughts on this, I'd be happy to hear other arguments and opinions :)

lenemter:
@leolost2605 everry test is still isolated though, no? Every test launches a new io.elementary.gala.tests and we set test(..., is_parallel: false)

And customization of the test still can be done since the tests don't run in the same process and don't share mutter context afaict

leolost2605:
> @leolost2605 everry test is still isolated though, no? Every test launches a new io.elementary.gala.tests and we set test(..., is_parallel: false)
> 
> And customization of the test still can be done since the tests don't run in the same process and don't share mutter context afaict

Like I said the approach itself makes sense but currently the isolation just happens to be there because we run the tests separately. I think we should really enforce it so no static meta.context, no possibility to run two tests at the same time already in the executable, etc.
This will also make the actual implementation of any customizations easier.

Like this (from the comment above):

> I think if we want to do single executable we will have to make sure that only a single test case can be run at one time and don't want to have any static variables. A main file that decides the type, constructs it, and just calls run on the testcase would probably work then.

PR #2837, opened 2026-04-30
Release 8.5.0

danirabbit:


danirabbit:
@leolost2605 @lenemter seems like a good batch of updates. Any more to sneak in? :) 

Not sure if we want to merge the ibus stuff after releasing this so we have some more time to test in Daily?

leolost2605 (approved):
> Not sure if we want to merge the ibus stuff after releasing this so we have some more time to test in Daily?

Yeah I think merging after the release is a good idea :)

lenemter (approved):

Issue #2844, opened 2026-05-12
Gala crash, session closed and new login required


ligursor:
### What Happened?

coredumpctl debug gala
           PID: 32829 (gala)
           UID: 1000 (roberto)
           GID: 1000 (roberto)
        Signal: 5 (TRAP)
     Timestamp: Tue 2026-05-12 08:33:33 CEST (3min 59s ago)
  Command Line: /usr/bin/gala
    Executable: /usr/bin/gala
 Control Group: /user.slice/user-1000.slice/user@1000.service/session.slice/_io.elementary.gala@wayland.service
          Unit: user@1000.service
     User Unit: io.elementary.gala@wayland.service
         Slice: user-1000.slice
     Owner UID: 1000 (roberto)
       Boot ID: e2750474a21643a3bb7b908317eb8e0d
    Machine ID: 98bc2188b3df1eac96b4d27f692ea61d
      Hostname: HP-EliteBook-640-G10
       Storage: /var/lib/systemd/coredump/core.gala.1000.e2750474a21643a3bb7b908317eb8e0d.32829.1778567613000000.zst (present)
  Size on Disk: 14.7M
       Message: Process 32829 (gala) of user 1000 dumped core.
                
                Module libgcc_s.so.1 from deb gcc-14-14.2.0-4ubuntu2~24.04.1.amd64
                Module libstdc++.so.6 from deb gcc-14-14.2.0-4ubuntu2~24.04.1.amd64
                Module libzstd.so.1 from deb libzstd-1.5.5+dfsg2-2build1.1.amd64
                Module libudev.so.1 from deb systemd-255.4-1ubuntu8.15.amd64
                Module libsystemd.so.0 from deb systemd-255.4-1ubuntu8.15.amd64
                Stack trace of thread 32829:
                #0  0x0000703bc6cd80af g_log_structured_array (libglib-2.0.so.0 + 0x620af)
                #1  0x0000703bc6cd839c g_log_default_handler (libglib-2.0.so.0 + 0x6239c)
                #2  0x0000703bc6cd8640 g_logv (libglib-2.0.so.0 + 0x62640)
                #3  0x0000703bc6cd8923 g_log (libglib-2.0.so.0 + 0x62923)
                #4  0x0000703bc6cd97a5 g_malloc0 (libglib-2.0.so.0 + 0x637a5)
                #5  0x0000703bc6eb8b76 n/a (libgala.so.0 + 0x2cb76)
[...]

Issue #2842, opened 2026-05-11
Crash in clutter_actor_child_model__items_changed after opening export dialog in GIMP


utf-4096:
### What Happened?

I got a session crash after trying to export a JPEG in GIMP.

### Steps to Reproduce

(I was not able to reproduce the issue myself, but this is what I was doing when I got the crash)

GIMP installed from AppCenter, non-system-wide.

1. Open JPEG file in GIMP through the *Open with...* context menu in Files
2. Crop something
3. Export using Ctrl+Shift+E (should open an export dialog)
4. See session crash

### Expected Behavior

No crash.

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

Secure Session (Wayland, This is the default)

### Software Version

Early Access or Compiled from git

### Log Output

```shell
#0  0x0000000000000000 in ??? ()
#1  0x00007d9c51757126 in clutter_actor_child_model__items_changed (model=0x59c7a3aec060, position=<optimized out>, removed=<optimized out>, added=<optimized out>, user_data=0x59c7dfff1c80) at ../clutter/clutter/clutter-actor.c:18108
#2  0x00007d9c52c322fa in g_closure_invoke (closure=0x59c7adddf590, return_value=0x0, n_param_values=4, param_values=0x7fff98169590, invocation_hint=0x7fff981694e0) at ../../../gobject/gclosure.c:834
#3  0x00007d9c52c6190c in signal_emit_unlocked_R.isra.0 (node=node@entry=0x7fff981696b0, detail=detail@entry=0, instance=instance@entry=0x59c7a3aec060, emission_return=emission_return@entry=0x0, instance_and_params=instance_and_params@entry=0x7fff98169590) at ../../../gobject/gsignal.c:3888
#4  0x00007d9c52c52591 in signal_emit_valist_unlocked (instance=instance@entry=0x59c7a3aec060, signal_id=signal_id@entry=255, detail=detail@entry=0, var_args=var_args@entry=0x7fff98169820) at ../../../gobject/gsignal.c:3520
#5  0x00007d9c52c527c1 in g_signal_emit_valist (instance=0x59c7a3aec060, signal_id=255, detail=0, var_args=var_args@entry=0x7fff98169820) at ../../../gobject/gsignal.c:3263
#6  0x00007d9c52c52883 in g_signal_emit (instance=<optimized out>, signal_id=<optimized out>, detail=<optimized out>) at ../../../gobject/gsignal.c:3583
[...]

Issue #2839, opened 2026-05-05
Semi-crash crash when alt-tabbing from secondary monitor to the first


zeebok:
### What Happened?

This is going to be a little weird. If I specifically click focus an app on my right/secondary monitor and then press alt-tab, with the next window to switch to being on the left/primary monitor, gala crashes and then returns the primary monitor to normal. All windows are still running as normal even on the second monitor, I just can't see them as they are blocked by the crash overlay. A quick glance in the log shows a huge allocation attempt?

<img width="4000" height="3000" alt="Image" src="https://github.com/user-attachments/assets/1efaf99e-6430-4baf-b129-bf1473d4397d" />

I have not tested if moving the layout or which is primary effects things, but I can if this description and the log don't point to anything. I have seen this issue over a few updates from the system settings.

To be clear, it is only when I click on the app on the right and alt-tab. If I alt-tab to switch to the app on the right, and alt-tab again it works fine.

### Steps to Reproduce

1. Get a dual monitor setup (mine is primary on the left, secondary on the right)
2. Open 2 apps, place a single app on each monitor
3. Click on the app on the right/secondary monitor
4. Press alt-tab

### Expected Behavior

No crash when alt-tabbing no matter what order of apps or monitor layout.

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

Secure Session (Wayland, This is the default)

### Software Version

Early Access or Compiled from git

### Log Output

```shell
May 05 00:22:43 xenon gala[50430]: gala_drawing_canvas_set_size: assertion 'new_width >= -1 && new_height >= -1' >
May 05 00:22:43 xenon kernel: show_signal: 24 callbacks suppressed
May 05 00:22:43 xenon kernel: traps: gala[50430] trap int3 ip:76c1946fd0af sp:7ffe54672b90 error:0 in libglib-2.0>
May 05 00:22:43 xenon gala[50430]: clutter_paint_volume_set_width: assertion 'width >= 0.0f' failed
May 05 00:22:43 xenon gala[50430]: clutter_paint_volume_set_width: assertion 'width >= 0.0f' failed
May 05 00:22:43 xenon gala[50430]: clutter_paint_volume_set_width: assertion 'width >= 0.0f' failed
May 05 00:22:43 xenon gala[50430]: clutter_paint_volume_set_width: assertion 'width >= 0.0f' failed
[...]

PR #2845, opened 2026-05-12
Release 8.5.1

leolost2605:
For #2840 
This fixes a 100% of the time crash on multi monitor setups when alt tabbing, see #2844, #2839, etc.

Note that this also includes #2805 which probably introduces #2842 which brings two questions:
- Should this be 8.6.0 since it is a pretty noticeable feature?
- Should we revert, release and then add it again?

### DIFF
--- a/data/gala.metainfo.xml.in
+++ b/data/gala.metainfo.xml.in
@@ -27,6 +27,19 @@
   <update_contact>contact_at_elementary.io</update_contact>
 
   <releases>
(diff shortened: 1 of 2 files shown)


stsdc:
Would do the second option. This report #2842 is pretty fresh, I would dogfood #2805 for a week or two.

leolost2605:
Opened #2846 for the revert

lenemter (approved):

stsdc (approved):

PR #2849, opened 2026-05-14
Tests: Use a single executable

leolost2605:
Supersedes and closes #2835 

This achieves the same performance boost but makes sure the tests can only be run isolated.
Most notably having the run method an instance method of a test case makes sure of that.

I know that this might be pedantic but the meta context stuff is quite fragile and it took a while to get it working so I think having it with as many guardrails as possible is probably for the better.

### DIFF
--- a/tests/lib/Main.vala
+++ b/tests/lib/Main.vala
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2026 elementary, Inc. (https://elementary.io)
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+namespace Gala {
+    public int main (string[] args) {
+        var test_name = args[1];
+
+        var test_case = get_test_case (test_name);
+
+        if (test_case == null) {
+            warning ("TestCase %s not found", test_name);
+            return 1;
+        }
+
+        return test_case.run (args);
+    }
+
+    private TestCase? get_test_case (string name) {
+        Type[] test_types = {
+            typeof (GestureControllerTest),
+            typeof (PropertyTargetTest),
+            typeof (SetupTest),
+            typeof (SwipeTriggerTest),
+        };
+
+        foreach (var type in test_types) {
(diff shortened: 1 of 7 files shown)


leolost2605:
@lenemter what do you think about this?

tests/lib/Main.vala
```diff
@@ -0,0 +1,29 @@
...
+
+    private TestCase? get_test_case (string name) {
+        switch (name) {
+            case "GestureControllerTest": return new GestureControllerTest ();
+            case "PropertyTargetTest": return new PropertyTargetTest ();
+            case "SetupTest": return new SetupTest ();
+            case "SwipeTriggerTest": return new SwipeTriggerTest ();
+            default: return null;
+        }
+    }
```
lenemter, line 35: I don't like duplicating class names as strings. Can we replace it with something like this?

```
Type[] test_types = {
    typeof (GestureControllerTest),
    typeof (PropertyTargetTest),
    typeof (SetupTest),
    typeof (SwipeTriggerTest),
};

TestCase? test_case = null;

for (var i = 0; i < test_types.length; i++) {
    var test_type = test_types[i];
    if (test_type.name () == args[1]) {
        test_case = (TestCase) GLib.Object.new_with_properties (test_type, {}, {});
        break;
    }
}

if (test_case == null) {
    warning ("TestCase %s not found", test_name);
    return 1;
}

return test_case.run (args);
```

lenemter (changes requested):
I really like the use of `test_case.run`! Good job. I left a comment regarding the `get_test_case` function

tests/lib/Main.vala
leolost2605: Yeah I agree that's probably better, I updated it :)
(I also removed the name property because we do Object.new now so it wouldn't work as it's currently done)

lenemter (approved):

PR #2852, opened 2026-05-28
Daemon/WindowListener: Prevent auto-suspend/idle when going fullscreen

flodavid:
When an application is in fullscreen, I think it should prevent the system going to sleep automatically. This is annoying especially when looking at videos (especially movies) or when gaming with something else than mouse and keyboard.
I did not found an existing issue about this.

This adds two methods to the gala daemon that are triggered by gala itself via dbus. One is used when any window goes fullscreen, to mark it as "inhibiting" automatic sleep to the session manager. The other method to "unhibit" for the same window when it exits fullscreen. Multiple windows can inhibit sleeping, they all must exit fullscreen to restore auto-sleep.

NB: If any window is fullscreen, auto sleep is disabled, even if it is not on the current virtual desktop.

TODO: add gschema setting for disabling auto suspend when fullscreen.

### DIFF
--- a/daemon-gtk3/DBus.vala
+++ b/daemon-gtk3/DBus.vala
@@ -32,6 +32,9 @@ public class Gala.Daemon.DBus : GLib.Object {
 
     private List<MonitorLabel> monitor_labels = new List<MonitorLabel> ();
 
+    private Gee.HashMap<string, uint> suspend_cookies = new Gee.HashMap<string, uint> ();
+    private Gee.HashMap<string, uint> idle_cookies = new Gee.HashMap<string, uint> ();
+
     construct {
         Bus.watch_name (BusType.SESSION, DBUS_NAME, BusNameWatcherFlags.NONE, gala_appeared, lost_gala);
     }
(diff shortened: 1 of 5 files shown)


leolost2605:
Hmmm IMO the app should do that itself if it wants it rather than us uncoditionally for all fullscreen apps. E.g. it might make sense for videos but I for example use a handwritten note taking app in fullscreen and I definitely want my device to go to sleep if I went to the cafeteria and forgot to suspend it.

flodavid:
> Hmmm IMO the app should do that itself if it wants it rather than us uncoditionally for all fullscreen apps. E.g. it might make sense for videos but I for example use a handwritten note taking app in fullscreen and I definitely want my device to go to sleep if I went to the cafeteria and forgot to suspend it.

I mainly use fullscreen modes for videos or games and I do not know many apps that have one, but for those use cases auto-sleep is pretty annoying.
But it is also why I want to add a setting to enable/disable this behaviour. Maybe instead of disabling auto-sleep, I should set it to the maximum of two hours ? Or add a new setting to power options dedicated to fullscreen auto-sleep delay ? Or a setting in the "Applications" configurations ?

leolost2605:
> But it is also why I want to add a setting to enable/disable this behaviour.

IMO while we could of course add a setting, settings always bring more untested code paths, more complexity etc. so a settings shouldn't be added lightly: https://docs.elementary.io/hig/design-philosophy/accessible-configuration
And I think in this case adding a setting to work around apps that just don't work the way they should isn't a good reason. Especially since the same thing is easily accesible from the quick settings. (In case you didn't know there's a prevent sleep toggle in the quick settings that does exactly that)

flodavid:
Yes, I know about this toggle. But when I launch a game I often forget to toggle it (or to untoggle it after) and it its not accessible when in fullscreen. And changing the virtual desktop can sometimes mess with the game. Same thing when I launch a movie, I forget to activate it before going in my couch. When the screen starts to go black, I do not have the time to jump on the keyboard to stop suspending. Also the computer sometimes just hangs while going to sleep or out of it, so it is extra annoying.
Maybe I should add a notification 1 minute (or like 5% of the time to sleep) before going to sleep, like many TV do. Or maybe a notification suggesting to activate the "Prevent sleep" option when going full-screen. But this one could may be annoying.

danirabbit:
I agree with @leolost2605 here. There is an API for apps to request that the session does not suspend. It's up to apps to use it. And also there is the prevent sleep toggle in quick settings. So a setting already exists. 

We shouldn't add functionality that only exists to work around buggy app behavior. Usually trying to be clever in this way ends up being interpreted as buggy behavior in itself.

PR #2854, opened 2026-05-29
Implement the greeter

leolost2605:
This allows running gala as the greeter compositor. It does that while preparing for having an in session lock screen.

It introduces:
- an environment variable that controls the session type
- a way to specify shell clients for a certain session type only
- a lock screen layer that contains the lock screen and lock screen shell window groups as well as the blurred background which is always active when the session is a greeter session and will serve as the lock screen in the desktop session (the second part is more long term and not yet implemented)

This is already enough to run gala as the greeter compositor. The long term goal is that the lock screen layer with the lock screen and lock screen shell window groups will also be used as a lock screen without switching ttys. This allows for a bunch of things like e.g. smoother animations, lock screen notifications, media controls etc.

Any feedback is appreciated especially regarding architecture and how we should animate the activation/deactivation of the lock screen layer. This animation will in the future be used for the lock screen but right now it would also be used when logging in to animate out the lock screen (aka to transition from the blurred background in the greeter to the normal session).

Goes with elementary/greeter#878
Includes #2855

### DIFF
--- a/lib/SessionSettings.vala
+++ b/lib/SessionSettings.vala
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2026 elementary, Inc. (https://elementary.io)
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ *
+ * Authored by: Leonhard Kargl <leo.kargl@proton.me>
+ */
+
+namespace Gala.SessionSettings {
+    private enum SessionType {
+        DESKTOP,
+        GREETER,
+        INSTALLER;
+    }
+
+    private static SessionType? session_type = null;
+
+    private static SessionType get_session_type () {
+        if (session_type == null) {
+            var session_type_str = Environment.get_variable ("GALA_SESSION_TYPE") ?? "desktop";
+            switch (session_type_str) {
+                case "desktop":
+                    session_type = DESKTOP;
+                    break;
+                case "greeter":
+                    session_type = GREETER;
+                    break;
+                case "installer":
+                    session_type = INSTALLER;
+                    break;
+                default:
+                    warning ("Unknown session type: %s", session_type_str);
+                    session_type = DESKTOP;
+                    break;
+            }
+        }
+
+        return session_type;
+    }
+
+    public bool is_greeter () {
+        return get_session_type () != DESKTOP;
+    }
+}
--- a/lib/SessionSettings.vala
+++ b/lib/SessionSettings.vala
@@ -40,4 +40,17 @@ namespace Gala.SessionSettings {
     public bool is_greeter () {
         return get_session_type () != DESKTOP;
     }
+
+    public string get_shell_clients_type () {
+        switch (get_session_type ()) {
+            case DESKTOP:
+                return "desktop";
+            case GREETER:
+                return "greeter";
+            case INSTALLER:
+                return "installer";
+        }
+
+        return "desktop";
+    }
 }
--- a/data/io.elementary.desktop.wm.shell
+++ b/data/io.elementary.desktop.wm.shell
@@ -9,3 +9,19 @@ launch-on-x=true
 args=io.elementary.dock
 wait-for-n-panels=1
 session-type=desktop
+
+[Greeter wingpanel]
+args=io.elementary.wingpanel;-g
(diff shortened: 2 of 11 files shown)


leolost2605:
This is pretty much ready. It took so long because I wanted to already prepare for an in session lock screen and I wasn't sure what the best way would be. So any opinions and suggestions regarding architecture are much appreciated. The animation for the deactivation of the greeter can probably be left to a follow up because without it the behavior is exactly the same as with the greeter compositor.
Note that the in session lock screen probably requires a pretty big rewrite of the greeter because (I think? but I didn't investigate much further) we can't use lightdm from within the session.

I've been using this together with the osk branch for a while now on my starlite and it's working great with no need for any keyboard at all.

lib/SessionSettings.vala
```diff
@@ -0,0 +1,56 @@
...
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ *
+ * Authored by: Leonhard Kargl <leo.kargl@proton.me>
+ */
+
+namespace Gala.SessionSettings {
+    private enum SessionType {
+        DESKTOP,
+        GREETER,
+        INSTALLER;
```
zeebok, line 12: Is the installer type necessary? I don't see it used anywhere, and is also rolled into greeter type when is_greeter  is called.

zeebok:
In regards to architecture for the lock screen, if I understand Gala and the goal correctly I would think the lock screen would almost be a view much like the multitasking view but with restrictions on events, such as keyboard shortcuts, that are accepted and hiding/disabling the dock, and maybe swapping out wingpanel for a more basic one.

As for animations, I would imagine going into the lock screen would be performing the typical hide animation of windows, panel, and dock, followed by the background blurring, and then the lock screen widgets being shown. Going out of the lock screen would be the reverse.

leolost2605:
> In regards to architecture for the lock screen, if I understand Gala and the goal correctly I would think the lock screen would almost be a view much like the multitasking view but with restrictions on events, such as keyboard shortcuts, that are accepted and hiding/disabling the dock, and maybe swapping out wingpanel for a more basic one.

That's pretty much exactly how it's implemented here :)

> As for animations, I would imagine going into the lock screen would be performing the typical hide animation of windows, panel, and dock, followed by the background blurring, and then the lock screen widgets being shown. Going out of the lock screen would be the reverse.

That sounds really cool but I think it would be some more effort so ig I'd rather leave that to a follow up? Given that on main there's also no animation at all I think it would be better to keep this PR from getting too big.

zeebok:
> That's pretty much exactly how it's implemented here :)

Ok cool, having the manager, unlike the multitasking view, made me unsure it was working a bit on its own instead of behaving like another view. Sweet!

> That sounds really cool but I think it would be some more effort so ig I'd rather leave that to a follow up? Given that on main there's also no animation at all I think it would be better to keep this PR from getting too big.

Absolutely! I would expect a different PR as well, I figured I would toss in my thoughts since you had brought up animations in the description; don't know how it would be able to be done but maybe getting the gears turning would help somebody figure the how out.

danirabbit (approved):
This looks good to me. I'd like to get this merged so we can unblock OS 9 dailies and we can further hash it out before release 🚀

lib/SessionSettings.vala
leolost2605: The greeter compositor launches different clients depending on whether it's a normal greeter session or a installer session (it doesn't launch the greeter, session manager, etc. in the installer session) that's why I added it here but I'm not familiar with the installer session and I haven't tested it :shrug:

leolost2605:
A few notes:
- the wingpanel currently always shows the black bar but that probably has to be addressed in the wingpanel repo anyways
- it currently doesn't yet use the wayland protocol that the greeter uses in the greeter compositor because it already works without it and there is currently a mismatch in the order of the methods in the protocol with gala and greeter so that would require some more changes also on the greeter side.

But all in all it works and I think I'm happy with the architecture (though I'd always love to hear other ideas) so IMO it makes sense to merge and iterate :)

danirabbit:
I agree. Let's go!

PR #2855, opened 2026-05-31
ModalProxy: Explicitly allow window groups

leolost2605:
While a non grabbed modal proxy is active we currently just allow any shell window to receive input focus.
This has the problem that e.g. when a modal window like the polkit dialog is active and then a desktop shell window tries to get focus (e.g. wingpanel reasks for your wifi password) it will give focus to that window which might not be immediately apparent. This will be especially noticeable with the lock screen so this also serves as preparation for #2854 
With this PR we need to explicitly allow window groups to be able to receive focus.

### DIFF
--- a/lib/WindowManager.vala
+++ b/lib/WindowManager.vala
@@ -38,6 +38,12 @@ namespace Gala {
         MEDIA_KEYS
     }
 
+    public enum WindowGroup {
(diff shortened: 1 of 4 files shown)


zeebok (approved):

leolost2605:
Included in #2854

PR #2863, opened 2026-07-07
Put Greeter below all windows

lenemter:
Required for https://github.com/elementary/greeter/pull/878
~~Requires https://github.com/elementary/greeter/pull/886~~
~~Requires https://github.com/elementary/greeter/pull/893~~

Makes Greeter unmovable and puts it below apps such as Initial Setup

### DIFF
--- a/protocol/pantheon-desktop-shell-v1.xml
+++ b/protocol/pantheon-desktop-shell-v1.xml
@@ -181,9 +181,10 @@
   <interface name="io_elementary_pantheon_greeter_v1" version="1">
     <request name="destroy" type="destructor"/>
 
-    <request name="init">
-      <description summary="init greeter">
-        Init Greeter
+    <request name="make_greeter">
+      <description summary="make greeter">
+        Request to make the surface a Greeter window, which will be displayed bellow other windows
+        when running in greeter compositor mode.
       </description>
     </request>
   </interface>
--- a/protocol/pantheon-desktop-shell-v1.xml
+++ b/protocol/pantheon-desktop-shell-v1.xml
@@ -35,6 +35,14 @@
       <arg name="output" type="new_id" interface="io_elementary_pantheon_extended_behavior_v1"/>
       <arg name="surface" type="object" interface="wl_surface"/>
     </request>
+
+    <request name="get_greeter">
+      <description summary="create a greeter surface from a surface">
+        Create a greeter surface from an existing surface.
+      </description>
+      <arg name="output" type="new_id" interface="io_elementary_pantheon_greeter_v1"/>
+      <arg name="surface" type="object" interface="wl_surface"/>
+    </request>
   </interface>
 
   <interface name="io_elementary_pantheon_panel_v1" version="1">
@@ -167,6 +175,16 @@
       </description>
 
       <arg name="monitor_index" type="int" summary="the index of the monitor this surface should label"/>
+    </request>
+  </interface>
+
+  <interface name="io_elementary_pantheon_greeter_v1" version="1">
+    <request name="destroy" type="destructor"/>
+
+    <request name="init">
+      <description summary="init greeter">
+        Init Greeter
+      </description>
     </request>
   </interface>
 </protocol>
--- a/protocol/pantheon-desktop-shell.vapi
+++ b/protocol/pantheon-desktop-shell.vapi
@@ -1,5 +1,5 @@
 /*
- * Copyright 2023 elementary, Inc. <https://elementary.io>
+ * Copyright 2023-2026 elementary, Inc. <https://elementary.io>
  * Copyright 2023 Corentin Noël <tintou@noel.tf>
  * SPDX-License-Identifier: GPL-3.0-or-later
  */
@@ -12,7 +12,7 @@ namespace Pantheon.Desktop {
         public Pantheon.Desktop.GetPanel get_panel;
         public Pantheon.Desktop.GetWidget get_widget;
         public Pantheon.Desktop.GetExtendedBehavior get_extended_behavior;
-
+        public Pantheon.Desktop.GetGreeter get_greeter;
     }
 
     [CCode (cheader_filename = "pantheon-desktop-shell-server-protocol.h", cname = "enum io_elementary_pantheon_panel_v1_anchor", cprefix="IO_ELEMENTARY_PANTHEON_PANEL_V1_ANCHOR_", has_type_id = false)]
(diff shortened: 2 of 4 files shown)


leolost2605 (changes requested):
I fully agree with the separate group and making it desktop but (going from previous experience with our wayland protocols just organically growing) I would like to have some bigger picture where we want to take the greeter before deciding on the wayland protocol (which is why I haven't added it in the initial PR yet).
Most notably what we want to do with multiple monitors:
- Greeter on primary?
- Greeter on current?
- Greeter on every monitor?

I've been thinking we might need greeter on every monitor because most notably for touchscreen we don't have a current monitor. I'm thinking of the case "connect tablet to external monitor without any accessories". Then in theory the only way to be 100% sure to be able to log in would be having it on every monitor. But I would definitely like some opinions from @elementary/ux on this

lenemter:
@leolost2605 My idea here is to let Gala decide what to do with Greeter windows, since there's no point in letting Greeter control all of this, since it's the only user of that protocol

danirabbit:
@leolost2605 I agree I think it would make sense to have the greeter on all displays. I think @lenemter is right though that probably the greeter app should just declare itself as a greeter and Gala should control what that means

leolost2605:
> My idea here is to let Gala decide what to do with Greeter windows

Yeah I agree I just wanted to make sure that there's nothing that needs integration between greeter and gala. I had the monitor labels in mind because they need to specify what monitor they belong to but ig the greeter should be the same for all monitors? Anything else that might need integration?

lib/WindowManager.vala
```diff
@@ -40,6 +40,7 @@ namespace Gala {
 
     public enum WindowGroup {
         DESKTOP_SHELL,
+        LOCK_SCREEN_APPLICATION,
```
leolost2605: Hmm IMO `LOCK_SCREEN_APPLICATION` sounds like where the "normal" windows go. Maybe something like `LOCK_SCREEN_GREETER`?

protocol/pantheon-desktop-shell-v1.xml
```diff
@@ -128,6 +136,16 @@
     <request name="destroy" type="destructor"/>
   </interface>
 
+  <interface name="io_elementary_pantheon_greeter_v1" version="1">
```
leolost2605: Only a style nit: I think we should keep the order of the methods in the shell interface, so that this comes last i.e. after extended behavior.

protocol/pantheon-desktop-shell.vapi
```diff
@@ -53,6 +53,14 @@ namespace Pantheon.Desktop {
         public Destroy destroy;
     }
 
+    [CCode (cheader_filename = "pantheon-desktop-shell-server-protocol.h", cname = "struct io_elementary_pantheon_greeter_v1_interface")]
+    public struct GreeterInterface {
```
leolost2605: Same here regarding the order

protocol/pantheon-desktop-shell.vapi
```diff
@@ -70,6 +78,8 @@ namespace Pantheon.Desktop {
     [CCode (has_target = false, has_typedef = false)]
     public delegate void GetWidget (Wl.Client client, Wl.Resource resource, uint32 output, Wl.Resource surface);
     [CCode (has_target = false, has_typedef = false)]
+    public delegate void GetGreeter (Wl.Client client, Wl.Resource resource, uint32 output, Wl.Resource surface);
```
leolost2605: And here :)

protocol/pantheon-desktop-shell.vapi
```diff
@@ -86,6 +96,8 @@ namespace Pantheon.Desktop {
     [CCode (has_target = false, has_typedef = false)]
     public delegate void RemoveBlur (Wl.Client client, Wl.Resource resource);
     [CCode (has_target = false, has_typedef = false)]
+    public delegate void Init (Wl.Client client, Wl.Resource resource);
```
leolost2605: And here :)

src/ShellClients/ShellClientsManager.vala
```diff
@@ -182,6 +182,50 @@ public class Gala.ShellClientsManager : Object, GestureTarget {
...
 #endif
 
+    public void make_desktop (Meta.Window window) {
+#if HAS_MUTTER49
+        window.set_type (Meta.WindowType.DESKTOP);
+#else
+        if (Meta.Util.is_wayland_compositor ()) {
+            make_desktop_wayland (window);
+        } else {
+            make_desktop_x11 (window);
```
leolost2605: Since there's currently (and probably never will be?) no way this is called I would replace that with a critical or assert_not_reached

src/PantheonShell.vala
```diff
@@ -17,6 +17,7 @@ namespace Gala {
     private static Pantheon.Desktop.ShellInterface wayland_pantheon_shell_interface;
     private static Pantheon.Desktop.PanelInterface wayland_pantheon_panel_interface;
     private static Pantheon.Desktop.WidgetInterface wayland_pantheon_widget_interface;
+    private static Pantheon.Desktop.GreeterInterface wayland_pantheon_greeter_interface;
```
leolost2605: Order once again :)

src/PantheonShell.vala
```diff
@@ -48,6 +50,11 @@ namespace Gala {
             destroy_widget_surface,
         };
 
+        wayland_pantheon_greeter_interface = {
```
leolost2605: And here :)

src/PantheonShell.vala
```diff
@@ -59,6 +66,7 @@ namespace Gala {
 
         PanelSurface.quark = GLib.Quark.from_string ("-gala-wayland-panel-surface-data");
         WidgetSurface.quark = GLib.Quark.from_string ("-gala-wayland-widget-surface-data");
+        GreeterSurface.quark = GLib.Quark.from_string ("-gala-wayland-greeter-surface-data");
```
leolost2605: And here :)

src/PantheonShell.vala
```diff
@@ -105,6 +113,25 @@ namespace Gala {
         }
     }
 
+    public class GreeterSurface : GLib.Object {
```
leolost2605: And here :)

leolost2605 (changes requested):
A few nits (sorry about all of the order nits but since we've done it like that for all other methods and additions I think we should keep it here too)

Also I've been thinking since we make the greeter a desktop window anyway do we still need to put it into a separate group?

lenemter:
@leolost2605 Done

src/ShellClients/ShellClientsManager.vala
```diff
@@ -182,6 +182,31 @@ public class Gala.ShellClientsManager : Object, GestureTarget {
     }
 #endif
 
+    public void make_desktop (Meta.Window window) {
```
leolost2605:
```suggestion
    private void make_desktop (Meta.Window window) {
```

src/ShellClients/ShellClientsManager.vala
```diff
@@ -290,6 +315,12 @@ public class Gala.ShellClientsManager : Object, GestureTarget {
         window.unmanaged.connect_after (() => ibus_candidate_window = null);
     }
 
+    public void init_greeter (Meta.Window window) {
```
leolost2605: Maybe we should call this `make_greeter` to keep with the nomenclature used for other windows in the shell clients manager?

leolost2605 (changes requested):
Two more comments apart from that LGTM!

src/ShellClients/ShellClientsManager.vala
lenemter: Maybe we should rename the protocol method to make_greeter as well while we're here? What do you think?

leolost2605: Would make sense to me :+1:

lenemter: Done!

leolost2605 (approved):
LGTM! 🚀

PR #2865, opened 2026-07-08
 Implement tests for src

leolost2605:
Add a `TestRunner` where we register the test types instead of doing it manually in main.
Add a simple test for the transitionbuilder that makes sure it gets freed even it's not run.

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -1,4 +1,4 @@
-gala_bin_sources = files(
+gala_bin_base_sources = files(
     'BlurManager.vala',
     'DBus.vala',
     'DBusAccelerator.vala',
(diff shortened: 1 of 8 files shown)


lenemter:
TBH TransitionBuilder should be moved to the library, as it doesn't seem to use anything from Gala itself

I have a branch with tests for gala here: https://github.com/lenemter/gala/pull/3, but it still needs a lot of work to figure out how to test things like the dock

leolost2605:
> TBH TransitionBuilder should be moved to the library, as it doesn't seem to use anything from Gala itself

Yeah ig it was just an easy thing to test for an initial implementation. We could do that in a follow up?

> I have a branch with tests for gala here: lenemter#3, but it still needs a lot of work to figure out how to test things like the dock

I had a quick look and there seems to be a lot of unrelated things? This PR just tries to get the initial infrastructure in place. Does it look like a similar direction or would you do it differently?

PR #2866, opened 2026-07-12
Add support for Mutter 51

garaevdi:
Also includes some changes introduced in 50.3 but they _shouldn't_ break builds with 50.1 present in the ubuntu repository

Keeping as draft until proper release


vapi/mutter-mtk-14.vapi
```diff
@@ -172,6 +172,8 @@ namespace Mtk {
 #if HAS_MUTTER50
 	[CCode (cheader_filename = "mtk/mtk.h")]
 	public static int64 find_nearest_interval_boundary (int64 boundary_us, int64 base_us, int64 interval_us);
+	[CCode (cheader_filename = "mtk/mtk.h")]
+	public static bool is_fd_readable (int fd);
```
tintou, line 176:
```vala
#endif
#if HAS_MUTTER51
	[CCode (cheader_filename = "mtk/mtk.h")]
	public static bool is_fd_readable (int fd);
```

PR #2871, opened 2026-07-24
Update POTFILES

lenemter:
Add all missing files


leolost2605 (approved):
Whoops forgot about that, thank you!

PR #2872, opened 2026-07-25
Move window type methods to ManagedClient

lenemter:
I think it makes sense to move it here until we require Mutter 49. Removes checking for mutter version across the codebase. In the future this will allow us to easily replace `ManagedClient.make_dock (window)` with `window.set_type (DOCK)`


leolost2605 (approved):
Makes sense to me :)

PR #2873, opened 2026-07-25
lib: remove ActivatableComponent

lenemter:
This class is unused since https://github.com/elementary/gala/pull/2753


leolost2605 (approved):
LGTM, nice cleanup

PR #2875, opened 2026-07-26
Require Mutter 46

lenemter:
I think the only reason we kept compatibility with mutter 45 is openSUSE Leap 15.6, but newer OpenSUSE Leap 16.0 provides mutter 48


leolost2605 (approved):
LGTM. Afair it doesn't even build on 45 anymore and nobody complained so we can probably drop it.

PR #2879, opened 2026-07-29
Introduce LayoutManager

leolost2605:
Split the logic regarding UI layout out of WindowManagerGala and into a separate LayoutManager.
The new LayoutManager handles settings things up and acts a little bit like a facade to the UI and widgets.
The reason to do this is to reduce the size and complexity of WindowManagerGala and introduce some more explicit bounds between different parts of Gala.

### DIFF
--- a/lib/WindowManager.vala
+++ b/lib/WindowManager.vala
@@ -40,6 +40,7 @@ namespace Gala {
 
     public enum WindowGroup {
         DESKTOP_SHELL,
+        MENU,
         LOCK_SCREEN,
         LOCK_SCREEN_SHELL,
         MODAL,
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -42,46 +42,15 @@ namespace Gala {
         public Clutter.Actor top_window_group { get; protected set; }
 
         /**
-         * The group that contains all WindowActors that make shell elements, that is all windows reported as
-         * ShellClientsManager.is_shell_window.
-         * It will (eventually) never be hidden by other components and is always on top of everything. Therefore elements are
-         * responsible themselves for hiding depending on the state we are currently in (e.g. normal desktop, open multitasking view, fullscreen, etc.).
-         */
-        private Clutter.Actor shell_group { get; private set; }
-
-        private Clutter.Actor menu_group { get; set; }
-
-        private LockScreen lock_screen;
-
-        /**
-         * The group that contains all WindowActors that are system modal.
-         * See {@link ShellClientsManager.is_system_modal_window}.
-         */
-        public ModalGroup modal_group { get; private set; }
-
-        private Clutter.Actor overlay_group;
-
-        /**
          * {@inheritDoc}
          */
         public Meta.BackgroundGroup background_group { get; protected set; }
-
-        /**
-         * View that allows to see and manage all your windows and desktops.
-         */
-        public MultitaskingView multitasking_view { get; protected set; }
-
-        public PointerLocator pointer_locator { get; private set; }
-
-        private SystemBackground system_background;
 
 #if !HAS_MUTTER48
         private Meta.PluginInfo info;
 #endif
 
-        private WindowSwitcher? window_switcher = null;
-
-        public WindowOverview window_overview { get; private set; }
+        private LayoutManager layout_manager;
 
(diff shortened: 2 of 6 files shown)


lib/WindowManager.vala
```diff
@@ -149,7 +150,7 @@ namespace Gala {
          *
          * @param type The type of action to perform
          */
-        public abstract void perform_action (ActionType type);
+        public abstract void perform_action (ActionType type, HashTable<string, GLib.Variant>? hints = null);
```
lenemter: I am -1 on adding hints. I actually plan to remove `hints` from `WindowOverview.open ()` and separate the single-app overview into a separate method.

src/WindowManager.vala
```diff
@@ -358,9 +228,9 @@ namespace Gala {
             display.add_keybinding ("cycle-workspaces-previous", keybinding_settings, NONE, handle_cycle_workspaces);
             display.add_keybinding ("panel-main-menu", keybinding_settings, IGNORE_AUTOREPEAT, handle_applications_menu);
 
-            display.add_keybinding ("toggle-multitasking-view", keybinding_settings, IGNORE_AUTOREPEAT, multitasking_view.toggle);
+            display.add_keybinding ("toggle-multitasking-view", keybinding_settings, IGNORE_AUTOREPEAT, layout_manager.toggle_multitasking_view);
 
-            display.add_keybinding ("expose-all-windows", keybinding_settings, IGNORE_AUTOREPEAT, window_overview.toggle);
+            display.add_keybinding ("expose-all-windows", keybinding_settings, IGNORE_AUTOREPEAT, layout_manager.toggle_window_overview);
```
lenemter: I don't think it makes sense to user LayoutManager to toggle multitasking view / window overview. I find it... illogical?

src/WindowManager.vala
```diff
@@ -584,7 +432,7 @@ namespace Gala {
          * {@inheritDoc}
          */
         public void switch_to_next_workspace (Meta.MotionDirection direction, uint32 timestamp) {
-            multitasking_view.switch_to_next_workspace (direction);
+            layout_manager.switch_to_next_workspace (direction);
```
lenemter: Same here. LayoutManager shouldn't have methods for switching workspaces

lenemter (changes requested):
I don't like making multitasking view / window overview private and adding methods to use them. Why don't just make them public at this point?

lib/WindowManager.vala
leolost2605: Any opinions on how to do this? Add a new method to WindowManager or just pass the WindowOverview/LayoutManager? I don't want to keep the window overview exposed as a property of the WindowManager.

lenemter: Passing WindowOverview seems to be a good solution

leolost2605:
> I don't like making multitasking view / window overview private and adding methods to use them. Why don't just make them public at this point?

While I would prefer it to keep them private I'm not super attached so if you feel very strongly about it we can change it. 
But I want to explain my reasoning a bit:

I would like to expose less internals. If you need to interact with the UI you interact with the layout manager and nothing else (except for lock screen and session locker for now because I didn't want to add to much to this PR). The layout manager holds the UI so you use the layout manager to interact with it. Like I said in the description a bit like the facade pattern.
This way changes that are exclusive to UI will need no changes outside of layout manager. Say if we drop the general window overview and adjust the window overview API to only be used for the window overview of an app we would only have to do changes in layout manager.

Another example for exposing internals:

> I am -1 on adding hints. I actually plan to remove hints from WindowOverview.open () and separate the single-app overview into a separate method.

Having a public window overview would suggest you implement this by passing the WindowOverview to the desktopintegration. But now if you want to change something e.g. merging the WindowOverview into the MultitaskingView (idk just an example) you would have to make a lot of changes outside of the actual code that was supposed to change. Whereas with this you just pass the layoutmanager and no matter what changes you make to UI you will at most have to change a method call.

leolost2605:
@lenemter I made the multitasking view and window overview public now. We also pass the window overview to the desktop integration.
I might look into reducing the exposure of internals in the future but I thought a bit more and we should probably do that differently. Also ig it's pretty good as is now.

lenemter (approved):
LGTM! Nice cleanup

PR #2881, opened 2026-08-01
Add OSK icons

danirabbit:
Screenshot from 2026-08-02 11 27 56@2x


danirabbit:
@leolost2605 having trouble figuring out where this resource should go and what the path should be. Any help? :)

leolost2605:
@danirabbit yeah the naming is currently all over the place and the gala and daemon resources are just one resource though we should probably split it at some point. But for now maybe there where you currently have it under `/io/elementary/desktop/gala-daemon/icons`?

leolost2605 (approved):
Looks great, thank you!

manu20247:
Is there a way to make the virtual keyboard movable on the desktop, like it was before? It's just a suggestion.

leolost2605:
@manu20247 currently there isn't but I opened an issue at #2909

PR #2883, opened 2026-08-07
Match the visual weight  `Return` with `Shift` and `CapsLock`

wpkelso:
Also add a bit of dynamism 🐎


danirabbit:
that's quite nice, good job

PR #2886, opened 2026-08-10
WindowManager: Use TransitionBuilder for minimize/unminimize

### LINKED ISSUE #2640: "Reduce Motion" Toggle Causing Window Instability
### What Happened?

I turned on the "Reduce Motion" setting and found that applications minimized on my dock wouldn't come into focus. Also, non-minimized windows became sporadically less responsive to mouse clicks (clicking buttons, trying to close them, etc.). Logging out and back in fixed it. I was able to reproduce this by turning off "Reduce Motion" and then turning it on again. Seems to happen consistently.

### Steps to Reproduce

1) Have applications open, minimized or not
2) Turn on "Reduce Motion"
3) Try to interact with the applications you have open

### Expected Behavior

Either "Reduce Motion" doesn't cause windows to misbehave at all, or it prompts you to log out first if it needs that to happen before things work properly

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

leolost2605:
Less code and less fragile.

Fixes #2640 (on main with reduce motion on I couldn't restore minimized windows, this is fixed here)

### DIFF
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -1135,45 +1135,32 @@ namespace Gala {
         }
 
         public override void unminimize (Meta.WindowActor actor) {
-            if (!Meta.Prefs.get_gnome_animations ()) {
-                actor.show ();
+            animate_unminimize.begin (actor);
+        }
+
+        private async void animate_unminimize (Meta.WindowActor actor) {
+            actor.show ();
+
+            if (actor.meta_window.window_type != NORMAL) {
                 unminimize_completed (actor);
                 return;
             }
 
-            var duration = AnimationDuration.HIDE;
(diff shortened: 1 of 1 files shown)


flodavid:
Would this fix https://github.com/elementary/wingpanel/issues/661 by any chance ?

leolost2605:
Probably not, the issues seems to be about maximizing/unmaximizing whereas this PR is about minimizing/unminimizing. But I'm planning to do maximizing/unmaximizing next :)

lenemter (approved):
🚀

PR #2888, opened 2026-08-10
Add a workflow that checks for missing and invalid files in POTFILES

leolost2605:
I always forget to update them. 
This is taken from flatpak: https://github.com/flatpak/flatpak/blob/main/.github/workflows/check-potfiles.yml and the corresponding script but I simplified it a lot. I know very little bash so lmk if something looks off.

It looks for .vala, .desktop.in and .metainfo.xml.in files and if they contain _( it checks whether they are in the POTFILES

Then it looks through all POTFILES and checks if they exist

First commit adds the check and I confirmed that it fails. Second commit fixes the potfiles and now it passes.

### DIFF
--- a/.github/workflows/check-potfiles.sh
+++ b/.github/workflows/check-potfiles.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+
+missing=0
+invalid=0
+
+git ls-files | grep -E '\.(vala|desktop.in|metainfo.xml.in)$' | while read -r file; do
+    if grep -q -E '\b((_|C_|N_)\(|ngettext \()' "$file"; then
+        if ! grep -q "^$file$" po/POTFILES; then
+            echo "Missing from POTFILES: $file"
+            missing=1
+        fi
+    fi
+done
+
+while IFS= read -r file; do
+    if [[ -z "$file" ]]; then
+        continue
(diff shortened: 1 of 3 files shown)


danirabbit:
Should we not just integrate this into our gettext action?

leolost2605:
> Should we not just integrate this into our gettext action?

I'd like to get feedback about missing POTFILES/files that don't exist anymore already in the PR. That's why I added it here because I thought the gettext action isn't intended to run on pull requests. But maybe we could add a new action template that does this?

stsdc (approved):

.github/workflows/check-potfiles.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+missing=0
+invalid=0
+
+git ls-files | grep -E '\.(vala|desktop.in|metainfo.xml.in)$' | while read -r file; do
+    if grep -q -E '\b(_|C_|N_)\(' "$file"; then
```
lenemter: We should check for `ngettext` here too

leolost2605: Done!

lenemter (approved):
LGTM

PR #2890, opened 2026-08-11
Use TransitionBuilder for maximizing

### LINKED ISSUE #661: No maximize animation when panel is not transparent
### What Happened?

When I set a wallpaper which makes panel opaque, after maximizing or minimizing a window there's no animation. When panel is transparent, windows are animating well. It happens on all computers.

### Steps to Reproduce

1. Set a light wallpaper to make panel opaque.
2. Try to maximize window.
3. You'll see there's no animation

### Expected Behavior

Windows animations should be independent of panel's transparency.

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

### LINKED ISSUE #2674: Maximize/Restore animation stopped working
### What Happened?

Hi everyone!
In the daily build, the maximize/restore animation is not working. (wayland session HiDPI 2X)
Sometimes, when I change the wallpaper – but not with every wallpaper – the animation starts working again for a while after switching to certain backgrounds, but after a restart it stops working again.

### Steps to Reproduce

1. Work with windows on desktop
2. When click maximize button
3. Not smooth animations, it just snaps up and back.

### Expected Behavior

Smooth animation when maximize/restore window (all type and size of wallpaper)

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

Secure Session (Wayland)

### Software Version

Compiled from git

### Log Output

```shell

```

### Hardware Info

_No response_

leolost2605:
Use TransitionBuilder for maximizing. While here do a bit of cleanup and drop some hacks. I checked how  GNOME Shell handles the general flow and followed that a bit more:

- If we already know we won't animate, don't capture a snapshot and complete size change immediately
- Introduce a pending size change map. Now instead of having only one size change info we can have multiple pending. This is required because it sometimes happens that we have multiple size_change before size_changed and we would drop snapshots then and animate a window with a wrong snapshot
- The last one also fixes the last `error in size accouting` errors.
- In the actual maximize animation we can remove the hack with the FLASH_PREVENT_TIMEOUT
- By dropping the unnecessary workarounds the animation code is now the same when maximizing and unmaximizing (just like in gnome shell) so we can merge it

Fixes elementary/wingpanel#661
Fixes #2674

### DIFF
--- a/src/WindowManager.vala
+++ b/src/WindowManager.vala
@@ -18,6 +18,18 @@
 
 namespace Gala {
     public class WindowManagerGala : Meta.Plugin, WindowManager {
+        private class SizeChangeInfo {
+            public Meta.SizeChange change;
+            public Mtk.Rectangle old_rect;
+            public Clutter.Actor snapshot;
+
+            public SizeChangeInfo (Meta.SizeChange change, Mtk.Rectangle old_rect, Clutter.Actor snapshot) {
+                this.change = change;
+                this.old_rect = old_rect;
+                this.snapshot = snapshot;
+            }
+        }
+
         private const string OPEN_MULTITASKING_VIEW = "dbus-send --session --dest=org.pantheon.gala --print-reply /org/pantheon/gala org.pantheon.gala.PerformAction int32:1";
         private const string OPEN_APPLICATIONS_MENU = "io.elementary.wingpanel --toggle-indicator=app-launcher";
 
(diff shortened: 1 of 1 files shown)


lenemter:
With this branch I get this when unmaximizing:
Screenshot from 2026-08-12 09 15 20

The reason we calculate offset there is because maximized windows don't have shadows, but non-maximized do have them.

And about FLASH_PREVENT_TIMEOUT. It was needed on X11 because some gtk3 windows are so slow to redraw that it flashed in places the window texture was missing, but this doesn't happen on Wayland for some reason. Since we don't plan to support X11 anymore, :+1: from me

leolost2605:
@lenemter thanks for the info that's really useful!

> The reason we calculate offset there is because maximized windows don't have shadows, but non-maximized do have them.

I added back the offset calculation but since we have access to the window after its rects have been updated we can get the unmaximized geometry immediately from there and don't need window listener unmaximized geometry tracking. However I still kept the tracking since it will be needed anyways in #2789 

I double checked with a 10 seconds animation time that both when maximizing and unmaximizing the actor and the snapshot match exactly.

Regarding the screenshot just to be sure: It was meant to illustrate the offset or did the snapshot stay after the animation finished?

lenemter:
@leolost2605 I wanted to illustrate the incorrect offset. Sorry for the confusion

leolost2605:
> @leolost2605 I wanted to illustrate the incorrect offset. Sorry for the confusion

Ah great then this should be ready again :)

lenemter (approved):
Code looks good and works as expected. Great cleanup!

PR #2893, opened 2026-08-12
Use DBusMenuModel and DBusActionGroup for the window menu

leolost2605:
Currently we just pass the actions to the WindowManager but in a follow up we could remove perform_action (or at least most of it) from there and handle it directly in the WindowMenuManager.


leolost2605:
@lenemter this is what I meant in https://github.com/elementary/gala/pull/2508#issuecomment-3386311928 
Sorry that I didn't get to this earlier but I wanted to try it first because IMO it has some advantages:
- we don't have to invent our own DBus Interface but can rely on a tried and tested one
- we don't have to write our own UI which means the menu will look exactly like other menus in the system and it is less code to maintain

Also I think actions is a very good fit here since then, if we want to support plugin menu items, we don't have to deal with callbacks which often lead to memory leaks but can just provide an action and a label and we are done.

lenemter (changes requested):
This breaks move/resize menu items. Probably because when they're pressed, the gala-daemon window is focused. Maybe wrap `on_move` / `on_resize` into `Idle` for now, will that work?

leolost2605:
> This breaks move/resize menu items. Probably because when they're pressed, the gala-daemon window is focused. Maybe wrap on_move / on_resize into Idle for now, will that work?

They are unfortunately already broken on main (at least for me). I double checked with this PR and set_grab_trigger is being called with the correct window so ig there is something wrong in set_grab_trigger. I would leave that to a follow up if that's ok

lenemter (approved):
Code looks good. Thanks for taking on this :)

PR #2894, opened 2026-08-12
MultitaskingView: Only grab focus on ws switch if we are open

### LINKED ISSUE #2884: Moving to a different workspace loses app focus
### What Happened?

Hi.

I'm spending a lot of time in the terminal, and recently I've noticed that when I move it to a different workspace, I cannot continue typing because it loses focus. This never happened before. 

### Steps to Reproduce

1. Start working in the terminal
2. Move it to the right workspace
3. No typing is possible until a mouse click or a shift-tab switch selects the terminal window, even when there is no other app in that workspace

### Expected Behavior

Previously, I was able to continue typing in Terminal immediately after the move.

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

### Software Version

### Log Output

```shell

```

### Hardware Info

Lenovo IdeaCentre 3 AiO, 8 × AMD Ryzen 7 4800U with Radeon Graphics, AMD® Renoir [Radeon RX Vega 6 (Ryzen 4000/5000 Mobile Series)]

leolost2605:
A commit can happen even when we are not open anytime that a workspace was reordered, added or removed. In those cases we don't want to grab focus because that will cause the focused window to stop receiving keyboard input.

A particular case where this was a problem was when moving windows via a gesture or the keyboard shortcut. There when the animation is over we call thaw_remove on the workspace manager and because a window
was moved from a workspace chances are high we will get a new one or one will be removed. In that case the window would immediately stop receiving keyboard
input even though it still appears focused.

Fixes #2884

### DIFF
--- a/src/Widgets/MultitaskingView/MultitaskingView.vala
+++ b/src/Widgets/MultitaskingView/MultitaskingView.vala
@@ -307,10 +307,12 @@ public class Gala.MultitaskingView : Root, RootTarget {
                     target_workspace.activate (display.get_current_time ());
                 }
 
-                /* Activating a workspace will set the input focus to the stage so recapture it
+                /* Activating a workspace will set the input focus to the stage so recapture it if we are open
                    (Gala.Root will automatically recapture it if we try to move the focus but we want
                    it right away for escape to work) */
-                grab_key_focus ();
+                if (opened) {
+                    grab_key_focus ();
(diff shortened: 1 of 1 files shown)


lenemter (approved):
Fixes the issue, good job 🎉

PR #2898, opened 2026-08-13
WindowMenu: Use unique bus name

leolost2605:
We currently do Bus.own_name with the same name in two places causing one to loose the name (DBus.vala prints a warning on every start). Everything is still working and all exported objects are still there and functioning but I think it's better to not rely on that. 
So now we have two options:
- have DBus.vala export the menu and action group
- use another bus name

I went with the second option because it keeps everything neatly contained in the WindowMenuManager and also regarding naming conventions this is how e.g. mutter (and most others?) do it (mutter uses a matching bus name for every object it exports).
I don't have much of an opinion here so we can also go with the first option :shrug: 

While here fix an oversight in the gtk3 daemon where old object paths were used.

### DIFF
--- a/daemon-gtk3/DBus.vala
+++ b/daemon-gtk3/DBus.vala
@@ -32,8 +32,8 @@ public class Gala.Daemon.DBus : GLib.Object {
             return;
         }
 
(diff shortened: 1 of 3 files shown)


lenemter (approved):
Makes sense

PR #2899, opened 2026-08-13
NotificationStack: Use transition builder for destroy animation

leolost2605:
This saves a bunch of code in the WindowManager because we don't have to do some special stuff anymore when dealing with destroying notifications.


lenemter (approved):
👍

PR #2900, opened 2026-08-13
BackgroundBlurEffect: Dont update actor size in construct

leolost2605:
When we construct a background blur effect we don't have an actor yet, we only get it once we are attached to an actor in the set_actor virtual method. This means we try to access the width/height of a null actor leading to two warnings every time a background blur effect is constructed.

Since update_actor_size () is called in set_actor anyways I think we can just remove it.


lenemter (approved):
Makes sense

PR #2903, opened 2026-08-13
WindowSwitcher: Disable hover for a short time after key events

### LINKED ISSUE #2727: Task switcher is very sensitive to mouse movement
### What Happened?

It may sound silly, but I often find myself struggling with <kbd>Alt</kbd> + <kbd>Tab</kbd> when my cursor is in the middle of the screen (where the switcher appears). This is on a desktop computer (keyboard and mouse).

https://github.com/user-attachments/assets/8f81a245-3122-433f-aed8-fcd31fa0d941

Even the slightest movement will interfere with the window selection. I'm not even consciously moving it, I'm just holding the mouse still; but it still moves probably because I'm breathing (obviously 😅) and often shaking my legs.

### Steps to Reproduce

1. Position your cursor near the center of the screen.
2. With <kbd>Alt</kbd> pressed, keep pressing <kbd>Tab</kbd> to move the window selection forwards; and simultaneously hold your mouse there (with your hands).

### Expected Behavior

It could be made more forgiving? When the keyboard shortcut is being pressed, ignore mouse movement?

If anyone relies on using the keyboard shortcut to start the switcher and then the mouse to select a window, maybe this could become a setting in `Desktop > Multitasking`?

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

### Software Version

### Log Output

```shell

```

### Hardware Info

Razer Deathadder v2 Mini, configured at 800 dpi.

leolost2605:
Fixes #2727 

Disable hovering icons for selection for a short time when initiating and after every key press. See the linked issue for details on the reason. GNOME Shell uses 500ms but I personally found that a bit too much so I went with 300ms but I'm open to suggestions. 
Tbh in general I don't like things like this because it is sometimes noticeable but not noticeable enough that you know what's going on and that makes it feel unpredictable and buggy. But ig in this case the main usage is probably keyboard workflows anyways so optimizing for that a bit more might be worth it but I'll leave that to UX.

### DIFF
--- a/src/Widgets/WindowSwitcher/WindowSwitcher.vala
+++ b/src/Widgets/WindowSwitcher/WindowSwitcher.vala
@@ -16,6 +16,8 @@ public class Gala.WindowSwitcher : AbstractSwitcher, GestureTarget, RootTarget {
     private int modifier_mask;
     private Gala.ModalProxy? modal_proxy;
     private int previous_icon_index = 0;
+    private bool mouse_active = true;
+    private uint hover_timeout_id = 0;
 
     private WindowSwitcherIcon? _current_icon = null;
     private WindowSwitcherIcon? current_icon {
@@ -138,6 +140,8 @@ public class Gala.WindowSwitcher : AbstractSwitcher, GestureTarget, RootTarget {
             open_switcher ();
         }
 
+        disable_hover ();
+
         var binding_name = binding.get_name ();
         var backward = binding_name.has_suffix ("-backward");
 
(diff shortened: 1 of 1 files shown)


leolost2605:
@o-alquimista if you want to give this a try I'd be happy about feedback :)

danirabbit:
I hate magic numbers too but I've also been annoyed by this issue and can't think of a better way to solve it off the top of my head

danirabbit (approved):

PR #2904, opened 2026-08-14
Show windows that are on all workspaces in the multitasking view

leolost2605:
Currently they just disappear as soon as the multitasking view is opened or workspaces are being switched. This looks weird and also blocks you from interacting with the window in the multitasking view.

So instead show them like normal windows but on every workspace. I think this makes more sense than just hiding it. Maybe we could even change the label in the window menu from "Always on visible workspace"  to something like "Show on all workspaces" and then it would make even more sense to the user.

The only alternative I could think off would be to treat it like a window that is currently moved between workspaces e.g. via the gesture. I.e. when switching workspaces just keep it at its current position and when opening the multitasking view fade it out. That's a bit more complicated to implement (but not that bad, I had a working prototype a while back somewhere) but more importantly IMO is that you wouldn't have access to managing it in the multitasking view. And I personally use pretty much exclusively the multitasking view for switching between open windows on a workspace which I couldn't do then. 

If we want to go with this here an open question would be whether we want to allow drag and drop in the multitasking view. Currently it is handled IMO quite logically that when you drag it to another workspace `on_all_workspaces` is unset and it's once again a normal window on the workspace you moved it to.

### DIFF
--- a/src/Misc/WindowListener.vala
+++ b/src/Misc/WindowListener.vala
@@ -54,7 +54,6 @@ public class Gala.WindowListener : Object {
     }
 
     public signal void window_workspace_changed (Meta.Window window);
-    public signal void window_on_all_workspaces_changed (Meta.Window window);
     public signal void window_maximized_changed (Meta.Window window) {
         WindowGeometry window_geometry = {};
         window_geometry.inner = window.get_frame_rect ();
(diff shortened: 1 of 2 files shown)


flodavid:
Code is good and I like behaviour.
Although I think the dock should evolve too to follow this behaviour, to make them coherent. Currently an "Always on visible workspace" window is only shown in the current workspace preview in the dock, and moves when the workspace changes.

leolost2605:
>  Although I think the dock should evolve too to follow this behaviour, to make them coherent. Currently an "Always on visible workspace" window is only shown in the current workspace preview in the dock, and moves when the workspace changes.

Agreed. Once this is merged I will follow up there :)

flodavid (approved):
code looks good

PR #2905, opened 2026-08-14
Scale down maximized windows when moving them between workspaces

leolost2605:
When moving a maximized window between workspaces (e.g. via the gesture or via the keyboard shortcut) the user can't see the animation because it is completely covered by the window. Therefore the user has no indication that something is actually happening. Solve this by scaling the window down a bit so that the user can see the animation.

Fixes #2799

IMO this works pretty well with gestures and with keyboard shortcuts if we move only one workspace (even though it's already a bit fast in that case). But it gets quite jarring when moving over multiple workspace via keyboard shortcuts (super + alt + number of workspace you want to move to). This can be seen at the end of the video. The reason for this is, that since with the gesture we don't know on which workspace we will end up we scale down and back up between every workspace. However in the case where you switch over multiple workspace via shortcut the same time that is usually used for scaling down and back up once is now used to scale down and backup as many times as workspace you switch over. 
So ig this is better than no indication at all for now but if somebody has ideas to improve this or completely different ideas I'd be very happy about them :)
I will probably do some experiments in trying to differentiate the cases and when switching via shortcuts only scale down and up once no matter how many workspace you switch over which should fix the jarring in that case. But in general I'd be happy about some feedback regarding the general direction.

[Kooha-2026-08-14-16-03-46.webm](https://github.com/user-attachments/assets/d56cad5c-1f55-4021-b3df-5ca4878ef0c2)

### DIFF
--- a/src/Widgets/MultitaskingView/StaticWindowClone.vala
+++ b/src/Widgets/MultitaskingView/StaticWindowClone.vala
@@ -26,5 +26,45 @@ public class Gala.StaticWindowClone : Widget {
 
         window_actor.bind_property ("x", this, "x", SYNC_CREATE);
         window_actor.bind_property ("y", this, "y", SYNC_CREATE);
+
+        set_pivot_point (0.5f, 0.5f);
+
+        window.notify["maximized-horizontally"].connect (on_maximized_changed);
+        window.notify["maximized-vertically"].connect (on_maximized_changed);
+        on_maximized_changed ();
+    }
+
+    public override void update_progress (GestureAction action, double progress) {
+        if (action == SWITCH_WORKSPACE && window.maximized_horizontally && window.maximized_vertically) {
+            /* When moving a maximized window between workspaces the user can't see the animation
+               because it is completely covered by the window. Therefore the user has no indication
+               that something is actually happening. Solve this by scaling the window down a bit so
+               that the user can see the animation. */
+
+            /* We want the final value (`inverted`) to run between integer
+               values from 0 to 1 at the half way point back to 0. */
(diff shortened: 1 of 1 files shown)


flodavid:
I like the idea, but the amount of movement is definitely jarring, even more when switching multiple workspace. Could we just make the top panel translucent during the transition ? Or at least make the animation (way) more subtle ? I tried reducing `0.25` to `0.05` in scale calculation (this should be set in a `const` by the way) ; it make the feature still useful, but more acceptable in term of movement.

wpkelso:
Is it possible to disconnect the animation of the window from the animation of the workspace switching? I.e. the window shrinks, then the carousel moves to its target, then the window grows again?

leolost2605:
> Could we just make the top panel translucent during the transition

Hmm interesting idea that would solve it quite cleanly. Though I'm thinking people might think it's a bug?

> I tried reducing 0.25 to 0.05 in scale calculation (this should be set in a const by the way) ; it make the feature still useful, but more acceptable in term of movement.

Yeah we can definitely play around with the scale. That was just a placeholder for now

> Is it possible to disconnect the animation of the window from the animation of the workspace switching? I.e. the window shrinks, then the carousel moves to its target, then the window grows again?

That was my initial thought as well and I had a bit of a hacky prototype, I'll see if I can find it. Completely sequential would be a bit difficult with the gesture but we could scale down with a timed animation as soon as we start and when we end scale back up. I'll see if I can put something together

flodavid:
Maybe you should mark this as draft while you try to improve of it ?

PR #2906, opened 2026-08-14
Drop X11 support

leolost2605:
Drop support for running under X11. On mutter < 50 enforce this by exiting early if we aren't a wayland compositor. 
See individual commits for the individual parts. There might be some workarounds and x specific fixes I missed but for now these are the obvious parts.


leolost2605:
Galas current release looks pretty stable on OS8 and also OS9 is already very usable so IMO it's a good time to start breaking changes :)

Issue #2751, opened 2026-02-06
Disable touchegg in Secure Session


lenemter:

leolost2605:
Touchegg is also still used for touchscreen and pinch gesture on touchpad

ryonakano:
Leaving a note that touchegg is no longer available for OS 9

https://packages.ubuntu.com/search?lang=en&searchon=names&keywords=touchegg

PR #2907, opened 2026-08-15
Implement touchpad pinch gestures in gala

leolost2605:
Towards #2751 since touchegg is no longer packaged for OS9. With this only touchscreen is left to do.

First rename the TouchpadBackend to TouchpadSwipeBackend, then introduce TouchpadPinchBackend.

Note that unfortunately in mutter 46 which we use on OS8 a bug was introduced that pinch gestures are captured by wayland clients i.e. they aren't forwarded to us when over a window:
https://gitlab.gnome.org/GNOME/mutter/-/work_items/3505

This was fixed for GNOME 49 so it isn't a problem for our OS9 but unfortunately it wasn't back ported :/
https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/4417

Therefore I added build conditions that on mutter < 49 we still use touchegg and on mutter 49 or higher we use the new backend. So if you are on mutter < 49 to test this you have to comment out the build conditions to enable the new TouchpadPinchBackend (in GlobalTrigger) and to filter out pinch gesture from touchegg (in ToucheggBackend). Then you can test this over any non window elements, e.g. in the multitasking view or over the background.

### DIFF
--- a/lib/Gestures/Backends/ToucheggBackend.vala
+++ b/lib/Gestures/Backends/ToucheggBackend.vala
@@ -180,7 +180,7 @@ private class Gala.ToucheggBackend : Object, GestureBackend {
         signal_params.get ("(uudiut)", out type, out direction, out percentage, out fingers,
             out performed_on_device_type, out elapsed_time);
 
-        if (Meta.Util.is_wayland_compositor () && performed_on_device_type != DeviceType.TOUCHSCREEN && type != PINCH) {
+        if (Meta.Util.is_wayland_compositor () && performed_on_device_type != DeviceType.TOUCHSCREEN) {
             return;
         }
 
--- a/lib/Gestures/Backends/ToucheggBackend.vala
+++ b/lib/Gestures/Backends/ToucheggBackend.vala
@@ -180,7 +180,11 @@ private class Gala.ToucheggBackend : Object, GestureBackend {
         signal_params.get ("(uudiut)", out type, out direction, out percentage, out fingers,
             out performed_on_device_type, out elapsed_time);
 
+#if HAS_MUTTER49
         if (Meta.Util.is_wayland_compositor () && performed_on_device_type != DeviceType.TOUCHSCREEN) {
+#else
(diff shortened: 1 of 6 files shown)


lenemter:
This differs from the existing pinch behavior because of mutter's "gesture cancel" thing.

If you set pinch-to-zoom in System Settings and use it, when you run out of space on your touchpad, the zoom resets. The same thing happens when you add a 4th finger, for example. But in `main` this doesn't reset the gesture, just stops it.

That's what mutter source code says:
```
 * @CLUTTER_TOUCHPAD_GESTURE_PHASE_CANCEL: The gesture was cancelled, all
 *   changes should be undone.
 
 * Cancelled gestures may be so for a variety of reasons, due to hardware,
 * or due to the gesture recognition layers hinting the gesture did not
 * finish resolutely (eg. a 3rd finger being added during a pinch gesture).
 * In these cases, the last event with report the phase
 * %CLUTTER_TOUCHPAD_GESTURE_PHASE_CANCEL, this should be used as a hint
 * to undo any visible/permanent changes that were done throughout the
 * progress of the gesture.
```

TBH I think we need to preserve both behaviors (for example, cancellation may be useful for https://github.com/elementary/gala/pull/1804), but this just doesn't work with zoom

leolost2605:
> This differs from the existing pinch behavior because of mutter's "gesture cancel" thing.

Should be fixed now

> TBH I think we need to preserve both behaviors (for example, cancellation may be useful for https://github.com/elementary/gala/pull/1804), but this just doesn't work with zoom

For now it just treats cancel as end. If we ever want to allow cancelling a gesture here it will be trivial to add a boolean allow_cancel or something that just emits on end with 0 progress on cancel. 
But I also think we probably should integrate that better with the gesturecontroller since then we can do animated progress undoing and we do already cancelling here depending on speed and travel distance. I would leave that to follow ups though because it also requires some design decisions (e.g. when a finger leaves the touchpad do we want to always cancel or cancel only depending on speed and travel distance (how it is currently)).

lenemter (approved):

PR #2914, opened 2026-08-27
Add keyboard shortcut and hotcorners for notifications menu

### LINKED ISSUE #313: Open with a keyboard shortcut
### Problem

There's currently no quick way to access the notifications list via keyboard when a new notification comes in

### Proposal
- [x] Open with notifications indicator with a keyboard shortcut.
- [x] Propose a branch to shortcuts overlay to list this as a system shortcut (could also just file a feature request and someone else can pick it up)

 Super + N makes sense to me and doesn't seem to be assigned to anything. Though there's some prior art (GNOME) for that jumping to the active notification instead of notification center

### Prior Art (Optional)

Super + N seems to be what Windows uses: https://www.windowscentral.com/windows-11-four-new-keyboard-shortcuts

Super + V was suggested here as it's the default in GNOME: https://github.com/elementary/notifications/issues/238

Most sources I could find said there is no shortcut by default in macOS, but [this source](https://mefmobile.org/how-to-invoke-a-specific-notification-center-view-using-keyboard-shortcuts-on-mac/) said Alt + F1

danirabbit:
Fixes elementary/panel-notifications#313


lenemter:
You should create keyboard shortcuts via org.gnome.Shell dbus interface like we did it in dock (https://github.com/elementary/dock/pull/245)

danirabbit:
Ohhh interesting thanks for the tip

danirabbit:
@lenemter It looks like to do that we need to do a lot of refactoring because the sound indicator already registers a keygrabber for the panel process

But also, then we wouldn't get the hotcorner. I think it has to live here for that reason anyways right?

lenemter:
@danirabbit Replied in https://github.com/elementary/wingpanel/issues/698. Also why do we need hotcorner for this in the first place? This was not mentioned in the original issue nor was it discussed anywhere else

PR #2915, opened 2026-08-30
Release 8.6.0

leolost2605:
I think gala is in a pretty good state right now so let's do one last release for OS 8 with quite a few goodies and after a week or two allow requiring OS 9.

I think now is a good time to do this so that we can focus on OS 9. E.g. I have just upgraded my main system to OS 9 so I won't be able to test any gala changes on older mutter. Things like wingpanel and granite are OS 9 exclusive now as well so I think it makes sense :)

### DIFF
--- a/data/gala.metainfo.xml.in
+++ b/data/gala.metainfo.xml.in
@@ -27,6 +27,24 @@
   <update_contact>contact_at_elementary.io</update_contact>
 
   <releases>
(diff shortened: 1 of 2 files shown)


zeebok (approved):
This seems good to me to do this. Gala has been pretty solid for me on OS8 daily!

PR #2916, opened 2026-08-30
Introduce a BrightnessManager

leolost2605:
Make it build by making sure vala first includes meta/display which contains the necessary macros for the others. I will properly fix this by upstreaming the missing header includes.


leolost2605:
@garaevdi if you want to take a look :) I fixed the build issues for now by forcing vala to use a certain include order

This follows the gnome shell implementation a bit (mostly regarding to using logical monitors and the global scale calculation which will be used by keybinds to change all monitor brightnesses at the same time while keeping the ratio between the monitors)
Needs still some work and the keybinds are missing but everything else is working already

garaevdi:
> @garaevdi if you want to take a look :) I fixed the build issues for now by forcing vala to use a certain include order

Oh, that's smart, didn't know about this trick 😅️

At a glance your implementation looks much cleaner, I'll close my PR. Just a suggestion: how about implicitly placing primary monitor at array's start, so interface consumers could highlight it somehow?

leolost2605:
> At a glance your implementation looks much cleaner, I'll close my PR. Just a suggestion: how about implicitly placing primary monitor at array's start, so interface consumers could highlight it somehow?

Good idea, did that :)

I also added support for the interface the gnome settings daemon expects for auto brightness and dimming on idle and added some docs for the dbus interface. Should be ready for an initial review now

src/Misc/Brightness/BrightnessManager.vala
```diff
@@ -0,0 +1,159 @@
...
+            throw new IOError.INVALID_ARGUMENT ("Invalid scale value: %f".printf (scale));
+        }
+
+        /* Make sure we don't divide by zero */
+        var max = double.max (get_global_brightness (), 0.01);
+
+        foreach (var monitor in supported_monitors) {
+            /* Make sure we don't get trapped at zero */
+            var monitor_val = double.max (monitor.value, 0.01);
+            monitor.value = (monitor_val / max) * scale;
```
garaevdi, line 155: Somewhere here you can get `-nan` when passing 0.00000000000000000000000000000000000000000000 to this method

leolost2605: In an old version this was the case but I thought the `double.max` should fix that. Did you forget to pull or does this still happen with the current one?

garaevdi: Oh, yeah, there was some stray gala executable hiding in /usr/local/bin, sorry for the buzz😅️

PR #2923, opened 2026-09-04
Utils: Don't check for logical framebuffer on mutter-50+

garaevdi:
It was enable by default since https://gitlab.gnome.org/GNOME/mutter/-/commit/74965cdb047df0f2159d53cb12770949924c60e9 which makes this check redundant.

Also in mutter-51 gnome devs changed the underlying gschema which caused crashes on gala side.


leolost2605 (approved):
LGTM. 

One note though: this is in theory not correct since you can still set the monitor layout mode to physical via the monitor config dbus interface. However since there is no UI to do that (the easiest way is probably gdctl) and not really a reason afaik I think it's fine to say we just don't support it

lib/Utils.vala
```diff
@@ -333,6 +333,10 @@ namespace Gala {
 
         private static bool? framebuffer_is_logical = null;
```
leolost2605, line 335: Maybe put this also in HAS_MUTTER50 guards otherwise I think there'll be a warning about an unused variable

leolost2605 (commented):
One comment

lib/Utils.vala
garaevdi: Done

PR #2925, opened 2026-09-07
NotificationStack: remove animations

danirabbit:
We can do lots of our own fancy animations now in GTK


leolost2605 (commented):
I suppose this is OS 9 material? If so I would wait a bit to see if we need another fix for regressions on OS 8 and then merge in a week or so

leolost2605 (approved):
I think we are good to drop support for OS8 and the code changes LGTM 
Feel free to merge once the notifications side is ready :)

PR #2927, opened 2026-09-15
Meson: Simplify mutter library detection

leolost2605:
Instead of having `if` for every mutter version add a dictionary that maps mutter api version to gnome version (this can be a list in the future since mutter changed the api version to follow the gnome version from 51 on) and then find the installed mutter version in a loop. 

In the best case this means that adding support for detecting a new version only needs to add a new entry to the dictionary instead of copy pasting the if and manually changing the versions etc.

This is prep for some more improvements I'd like to make in this area


leolost2605:
@tintou @garaevdi does this make sense?

tintou:
I'm not really in favor because we've seen the dependency list changing from one version to another so I want to keep it simple even if it's a bit redundant

tintou:
Like here you have `mutter-cogl-pango-14` and `mutter-cogl-pango-15` that disappear from the list

leolost2605:
> dependency list changing

@tintou yeah that was my thought as well. Though with mtk in place and cogl-pango gone I thought it will probably be relatively stable now but I don't feel very strongly about this.

Another question while I have you here: Is there a reason we don't use the generated vapi directly but instead via a single one with `#if`? I did some experimenting and at least with mutter 18 and a few metadata adjustments it seems to work for me and would be less work, wouldn't it? That's the next thing I wanted to do :)

garaevdi:
I don't have much of an opinion here. Copying lines manually twice a year isn't a massive task and personally it isn't a huge deal for me. Most of the time there would be around 7 supported mutter version so the size of `meson.build` shouldn't grow that much over time.

But I also think that this looks much cleaner than what we have now, especially if you also shorten `vapi/meson.build` with a similar string substitution.

tintou:
We have to document the reasoning because this comes every few year :) but I don't mind, the reason we have the single .vapi with #ifs is because It makes very easy to spot differences in the .vapi and soo where some API has been moved while having a single one makes it harder to compare :slightly_smiling_face:

